PR #3, opened 2022-05-24
purego: use syscall.syscallX for functions with 3 or less arguments

TotallyGamerJet:


syscall.go
```diff
@@ -21,6 +21,11 @@ func SyscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 	if len(args) > maxArgs {
 		panic("too many arguments to SyscallN")
 	}
+	if len(args) <= 3 {
```
hajimehoshi, line 24: * Could you write a comment why we needed this?
* Do functions with 4-6 arguments work as expected?

hajimehoshi:
Also I'd be happy if you could take a look at these comments:

https://github.com/ebiten/purego/pull/1#discussion_r879473946
https://github.com/ebiten/purego/pull/1#discussion_r879474628

syscall.go
TotallyGamerJet: I believe so but just in case I used the 6 version instead of the 3 version. The tests still seem to pass.

hajimehoshi (approved):
LGTM, thanks!

PR #5, opened 2022-06-03
purego: add go:nosplit to SyscallN

hajimehoshi:
Without go:nosplit, a local variable's address on a stack in a SyscallN
caller might be invalidated unexpectedly.

### DIFF
--- a/syscall.go
+++ b/syscall.go
@@ -20,6 +20,8 @@ const maxArgs = 9
 //
 // On amd64, if there are more than 8 floats the 9th and so on will be placed incorrectly on the
 // stack.
+//
+//go:nosplit
 func SyscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 	if len(args) > maxArgs {
 		panic("too many arguments to SyscallN")
--- a/syscall.go
+++ b/syscall.go
@@ -48,6 +48,7 @@ var syscall9XABI0 uintptr
 
 func syscall9X() // implemented in assembly
 
+//go:nosplit
 func syscall_syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2, err uintptr) {
 	args := struct{ fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, r1, r2, err uintptr }{
 		fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, r1, r2, err}


hajimehoshi:
@TotallyGamerJet Please take a look, thanks

hajimehoshi:
I'm not 100% sure but this seems the cause of https://github.com/hajimehoshi/ebiten/issues/2102.

hajimehoshi:
https://github.com/hajimehoshi/ebiten/issues/2120

Hmm, `go:nosplit` has a potential risk of stack overflow...

hajimehoshi:
Oops

```
$ go build -tags=example -gcflags=all=-N ./examples/blocks/
# github.com/hajimehoshi/ebiten/v2/examples/blocks
gosave_systemstack_switch: nosplit stack overflow
        792     assumed on entry to github.com/ebiten/purego.SyscallN<1> (nosplit)
        320     after github.com/ebiten/purego.SyscallN<1> (nosplit) uses 472
        312     on entry to syscall.syscall6X<1> (nosplit)
        208     after syscall.syscall6X<1> (nosplit) uses 104
        200     on entry to runtime.entersyscall<1> (nosplit)
        176     after runtime.entersyscall<1> (nosplit) uses 24
        168     on entry to runtime.reentersyscall<1> (nosplit)
        120     after runtime.reentersyscall<1> (nosplit) uses 48
        112     on entry to runtime.save<1> (nosplit)
        88      after runtime.save<1> (nosplit) uses 24
        80      on entry to runtime.badctxt<1> (nosplit)
        56      after runtime.badctxt<1> (nosplit) uses 24
        48      on entry to runtime.throw<1> (nosplit)
        8       after runtime.throw<1> (nosplit) uses 40
        0       on entry to runtime.systemstack<0> (nosplit)
        -8      on entry to gosave_systemstack_switch<28> (nosplit)
```

hajimehoshi:
* Without `//go:nosplit`, the current `SyscallN` seems dangerous since a local variable's address can be invalidated anytime after `SyscallN` is called, as far as I understand.
* With `//go:nosplit`, the compilation w/o optimization fails with warnings of stack overflow.

So, we have to reduce the stack usages in `SyscallN`, or we have to give up `SyscallN`. @TotallyGamerJet What do you think?

TotallyGamerJet:
@hajimehoshi sorry, I just saw this. I will take a look at this tomorrow!

TotallyGamerJet:
My mind is boggled. I can not reproduce the error that caused https://github.com/hajimehoshi/ebiten/issues/2120. I tried optimizations and no optimizations and then tried using nosplit all while just using syscall_Syscall9X and it worked for all tests. Thinking of a replacement, I came across [noncgo/x](https://github.com/noncgo/x/blob/7e35ac977becc6f51486f575c7eba616ba5db42d/darwin/internal/cabi/cabi.go) which has C calling by using a different method. It is also in the public domain so it's safe to use as inspiration. I think this method can avoid `go:nosplit` because it doesn't erase types like SyscallN does. I'll try to use a similar method.

hajimehoshi:
Let me close % https://github.com/ebitengine/purego/pull/6

PR #7, opened 2022-06-19
purego: add //go:uintptrescapes to SyscallN

TotallyGamerJet:
this forces the uintptr arguments to escape in SyscallN. When optimizations were enabled Go assumed that the []uintptr argument in SyscallN could be placed on the stack. This is why functions like gl.GenFramebuffersEXT would fail. Although this change forces all of the arguments to be placed on the heap it ensures that no optimizations will remove our precious arguments. This also allows us to only have one implementation inside SyscallN (i.e. syscall_syscall9X). A potential replacement would be //go:cgo_unsafe_args but that can only be used in CGO generated files. There is also //go:uintptrkeepalive although that was not enough to stop the function from failing. That is why I went with //go:uintptrescapes. More information about this pragma can be found on the [compile](https://pkg.go.dev/cmd/compile#section-documentation) documentation.

I tested this change on darwin (arm64 & amd64) as well as with -gcflags=all=-N and without. All tests pass on these platforms.

I have not tested this on windows. If you could that would be much appreciated.

### DIFF
--- a/runtime.go
+++ b/runtime.go
@@ -10,9 +10,6 @@ import (
 	"unsafe"
 )
 
-//go:linkname syscall_syscall6X syscall.syscall6X
-func syscall_syscall6X(fn, a1, a2, a3, a4, a5, a6 uintptr) (r1, r2, err uintptr) // from runtime/sys_darwin_64.s
-
 //go:linkname runtime_libcCall runtime.libcCall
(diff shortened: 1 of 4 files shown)


hajimehoshi (approved):
LGTM, thanks!

PR #8, opened 2022-06-26
No red zone

TotallyGamerJet:
sys_darwin_amd64.s was pushing the struct pointer to -8(SP) which is a part of the red-zone. We have enough space on the stack to push it above the arguments. Doing so is safer too


hajimehoshi (approved):
LGTM, thanks!

PR #9, opened 2022-06-27
remove strings package

TotallyGamerJet:
there is no reason to import strings just for one function

### DIFF
--- a/dl_darwin.go
+++ b/dl_darwin.go
@@ -5,7 +5,6 @@ package purego
 
 import (
 	"runtime"
-	"strings"
 	"unsafe"
 )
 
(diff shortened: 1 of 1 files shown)


dl_darwin.go
```diff
@@ -5,16 +5,20 @@ package purego
...
-	"strings"
 	"unsafe"
 )
 
 const RTLD_GLOBAL = 0x8
 
 const RTLD_DEFAULT = ^uintptr(1)
 
+// HasSuffix tests whether the string s ends with suffix.
+func _HasSuffix(s, suffix string) bool {
```
hajimehoshi, line 16: `hasSuffix`?

hajimehoshi (approved):
LGTM, thanks!

PR #11, opened 2022-07-20
add Objective-C runtime bindings

TotallyGamerJet:
This pull request adds Cgo-free bindings to the Objective-C runtime. Some frameworks may still require CGO_ENABLED=1 if it uses TLS (such as AudioToolbox). However, the bindings are fully functional and do not require a C compiler to build.

I plan on rewriting the Cgo runtime in Go similar to [nocgo/internal/fakecgo](https://github.com/notti/nocgo/tree/master/internal/fakecgo). I think it would be best to port Ebitengine and Oto to use purego/objc before starting this because it will give a larger test base.


hajimehoshi:
Thanks! Please let me know if this is ready for review!

TotallyGamerJet:
It is!

internal/strings/strings.go
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebiten Authors
```
hajimehoshi, line 2: The Ebitengine Authors

objc/objc_runtime.go
```diff
@@ -0,0 +1,124 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebiten Authors
```
hajimehoshi, line 2: The Ebitengine Authors

internal/abi/abi_amd64.h
```diff
@@ -0,0 +1,99 @@
+// Copyright 2021 The Go Authors. All rights reserved.
```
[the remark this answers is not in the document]
TotallyGamerJet: Yes it is copied from runtime/cgo. What should be included in this README?

zcallback_darwin_amd64.s
```diff
@@ -0,0 +1,2014 @@
+// Code generated by wincallback.go using 'go generate'. DO NOT EDIT.
```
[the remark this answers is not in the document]
TotallyGamerJet: The zcallback files are copied from runtime but they are generated by a file. I also changed them to call purego’s version of callbackasm1 instead of the runtime version. There is no license header to these files since they are generated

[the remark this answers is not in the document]
TotallyGamerJet: The generator is also in the runtime package

objc/objc_runtime.go
```diff
@@ -0,0 +1,124 @@
...
+
+//go:build darwin
+// +build darwin
+
+// Package objc is a low-level pure Go objective-c runtime. This package is easy to use incorrectly so it is best
+// to use a wrapper that provides the functionality you need in a safer way.
+//
+// All functions that take a string as an argument are NULL terminated ('\x00'). This is so that there is no
```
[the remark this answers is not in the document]
TotallyGamerJet: They are regular Go strings just need to add “\x00” to the end of it. And the package does verify that all the strings have been terminated properly. You can look at my Oto PR for what it looks like to actually use. I’m willing to allow non-null-terminated strings if you want but since calling into the runtime happens a LOT in Objc it would be best not to put extra pressure on the GC.  if you write the string as a constant with a null termination then the compiler is likely to put it in the data section and then there is no copying of data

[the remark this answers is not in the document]
hajimehoshi: >  I’m willing to allow non-null-terminated strings if you want but since calling into the runtime happens a LOT in Objc it would be best not to put extra pressure on the GC. 

I cannot say anything without measuring, but I feel like this is a too early optimization. What about accepting regular strings and revisting this when we find this is really a problem?

internal/abi/abi_amd64.h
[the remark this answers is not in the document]
hajimehoshi: What files are copied from the Go runtime.

On second thought, with `zcallback_darwin_*.s`, we can add the note to the exisitng README instead of adding a new README in this directory

objc/objc_runtime.go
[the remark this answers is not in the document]
TotallyGamerJet: Well we can accept both! That way if someone finds that the GC pressure is too much then they can use null-terminated strings

internal/abi/abi_amd64.h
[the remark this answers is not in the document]
TotallyGamerJet: So add a note about both zcallback and internal/abi? What should it include?

objc/objc_runtime.go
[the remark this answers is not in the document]
hajimehoshi: Ah yes, always checking the string be null-terminated or not and adding \x00 if needed would work :-) (Is my understanding correct?)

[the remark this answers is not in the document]
TotallyGamerJet: Yep that is correct!

internal/abi/abi_amd64.h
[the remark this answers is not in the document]
hajimehoshi: Please list all the files from the Go runtime and add a note that the license is BSD2 for these files.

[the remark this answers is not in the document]
TotallyGamerJet: Does the README.md look good?

internal/strings/strings.go
```diff
@@ -0,0 +1,26 @@
...
+
+// hasSuffix tests whether the string s ends with suffix.
+func hasSuffix(s, suffix string) bool {
+	return len(s) >= len(suffix) && s[len(s)-len(suffix):] == suffix
+}
+
+// CString converts a go string to *byte that can be passed to C code.
+// if requireNil is true it will panic if the passed string doesn't have
+// a null byte at the end.
+func CString(name string, requireNil bool) *byte {
```
hajimehoshi, line 16: We no longer need `requrieNil` as this is always false.

internal/abi/abi_amd64.h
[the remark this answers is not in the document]
hajimehoshi: Yes!

objc/objc_runtime.go
```diff
@@ -0,0 +1,174 @@
...
+		superClass: id.Class(),
+	}
+	var tmp = make([]uintptr, len(args)+2)
+	createArgs(tmp, Id(unsafe.Pointer(_super)), sel, args...)
+	ret, _, _ := purego.SyscallN(objc_msgSendSuper2, tmp...)
+	return Id(ret)
+}
+
+func createArgs(out []uintptr, cls Id, sel SEL, args ...interface{}) {
+	if len(out) < len(args)+2 {
```
hajimehoshi, line 70: What about making a slice inside `createArgs`?

```
out := make([]uintptr, len(args) + 2)
```

By the way, why is `len(out) < len(args)+2` required?

objc/objc_runtime.go
```diff
@@ -0,0 +1,174 @@
...
+			out = append(out, uintptr(v))
+		default:
+			panic(fmt.Sprintf("objc: unknown type %T", v))
+		}
+	}
+}
+
+type SEL uintptr
+
+func RegisterName(name string) SEL {
```
hajimehoshi, line 106: Can we add a comment?

sys_darwin_arm64.s
```diff
@@ -60,3 +62,75 @@ TEXT syscall9X(SB), NOSPLIT, $0
...
+	MOVD $callbackWrapInternal<>(SB), R0 // fn unsafe.Pointer
+	MOVD R13, R1                         // frame (&callbackArgs{...})
+	MOVD $0, R3                          // ctxt uintptr
+
+/*
+ * We still need to save all callee save register as before, and then
+ *  push 3 args for fn (R0, R1, R3), skipping R2.
+ * Also note that at procedure entry in gc world, 8(RSP) will be the
+ *  first arg.
+ */
```
hajimehoshi, line 111: Let's use `//`

syscall_darwin.go
```diff
@@ -18,3 +26,116 @@ func syscall_syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2,
 	callc(syscall9XABI0, unsafe.Pointer(&args))
 	return args.r1, args.r2, args.err
 }
+
+func NewCallback(fn interface{}) uintptr {
```
hajimehoshi, line 30: Is it possible to dispose a created callback? We need a comment about the lifetime.

objc/objc_runtime.go
TotallyGamerJet: The +2 is there because self and SEL are the first two arguments yet aren’t in the args slice

TotallyGamerJet: Oh your asking about why check the length when it’s an internal method. It’s never too crazy to check input but I guess it’s not needed bc it’s internal

hajimehoshi: I mean why we don't create a new slice inside createArgs.

README.md
```diff
@@ -2,3 +2,12 @@
 [![Go Reference](https://pkg.go.dev/badge/github.com/ebitengine/purego?GOOS=darwin.svg)](https://pkg.go.dev/github.com/ebitengine/purego?GOOS=darwin)
 
 A library for calling C functions from Go without Cgo.
+
+### External Code
+Purego uses code that originates from the Go runtime. These files are under the BSD-3
```
hajimehoshi, line 7: Add an empty line after `### External Code`

objc/objc_runtime.go
```diff
@@ -0,0 +1,192 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build darwin
+// +build darwin
```
hajimehoshi, line 5: Use the filename `objc_runtime_darwin.go` instead of a build tag.

syscall_darwin.go
```diff
@@ -18,3 +26,118 @@ func syscall_syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2,
...
+// NewCallback converts a Go function to a function pointer conforming to the C calling convention
+// This is useful when interoperating with C code requiring callbacks. The argument is expected to be a
+// function with one uintptr-sized result. The function must not have arguments with size larger than the size
+// of uintptr. Only a limited number of callbacks may be created in a single Go process, and any memory allocated
+// for these callbacks is never released. At least 2000 callbacks can always be created.
+func NewCallback(fn interface{}) uintptr {
+	return compileCallback(fn)
+}
+
+// only increase this if you have added more to the callbackasm function
```
hajimehoshi, line 39: Let's follow the Go's commen rule i.e., let's start the comment with `maxCB`.

syscall_darwin.go
```diff
@@ -18,3 +26,118 @@ func syscall_syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2,
...
+	cbs.numFn++
+	return callbackasmAddr(cbs.numFn - 1)
+}
+
+const ptrSize = unsafe.Sizeof((*int)(nil))
+
+const callbackMaxFrame = 64 * ptrSize
+
+// Implemented in zcallback_GOOS_GOARCH.s
+var callbackasmABI0 uintptr
```
hajimehoshi, line 100: ditto

syscall_darwin.go
```diff
@@ -18,3 +26,118 @@ func syscall_syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2,
...
+
+// this exists to remove the unused function error message
+var _ = callbackWrap
```
hajimehoshi, line 103: What is "the unused function error message"? I have never seen...

objc/objc_runtime.go
```diff
@@ -0,0 +1,192 @@
...
+	tmp[0] = uintptr(unsafe.Pointer(super)) // if createArgs splits the stack the pointer would be wrong
+	ret, _, _ := purego.SyscallN(objc_msgSendSuper2, tmp...)
+	return Id(ret)
+}
+
+func createArgs(cls Id, sel SEL, args ...interface{}) (out []uintptr) {
+	out = make([]uintptr, len(args)+2)
+	out[0] = uintptr(cls)
+	out[1] = uintptr(sel)
+	out = out[:2]
```
hajimehoshi, line 76: Optional: what about

```go
	out = make([]uintptr, 2, len(args)+2)
	out[0] = uintptr(cls)
	out[1] = uintptr(sel)
```

?

objc/objc_runtime.go
```diff
@@ -0,0 +1,192 @@
...
+	objc_allocateClassPair = purego.Dlsym(objc, "objc_allocateClassPair")
+	objc_registerClassPair = purego.Dlsym(objc, "objc_registerClassPair")
+	sel_registerName       = purego.Dlsym(objc, "sel_registerName")
+	class_getSuperclass    = purego.Dlsym(objc, "class_getSuperclass")
+	class_addMethod        = purego.Dlsym(objc, "class_addMethod")
+	object_getClass        = purego.Dlsym(objc, "object_getClass")
+)
+
+// Id is an opaque pointer to some Objective-C object
+type Id uintptr
```
hajimehoshi, line 39: Let's use `ID` rather than `Id` for the Go's convention (or the Go's standard libraries).

objc/objc_runtime.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// Send is a convenience method for sending messages to objects.
+func (id Id) Send(sel SEL, args ...interface{}) Id {
+	tmp := createArgs(id, sel, args...)
+	ret, _, _ := purego.SyscallN(objc_msgSend, tmp...)
+	return Id(ret)
+}
+
+// sending a message to the super requires this struct instead of the object itself
```
hajimehoshi, line 54: objc_super is ...

hajimehoshi (commented):
Thank you!

syscall_darwin.go
TotallyGamerJet: It was an IDE warning. There is no issue compiling with or without this line

TotallyGamerJet:
Question: On Windows, NewCallback requires a pointer-sized return value. However, on macOS there are a lot of callbacks that return void and it is strange to `return 0` in these cases. Do you think we should support returning nothing? It's a 3 line change in `callbackWrap`. This can always be done later since it would be a superset and not effect current uses

hajimehoshi:
> Question: On Windows, NewCallback requires a pointer-sized return value. However, on macOS there are a lot of callbacks that return void and it is strange to return 0 in these cases. Do you think we should support returning nothing? It's a 3 line change in callbackWrap. This can always be done later since it would be a superset and not effect current uses

I think it's OK to allow a non-returning function for Darwin's NewCallback. We should not be confused with Windows' one and Darwin's one.

README.md
```diff
@@ -2,3 +2,13 @@
 [![Go Reference](https://pkg.go.dev/badge/github.com/ebitengine/purego?GOOS=darwin.svg)](https://pkg.go.dev/github.com/ebitengine/purego?GOOS=darwin)
 
 A library for calling C functions from Go without Cgo.
+
+### External Code
+
+Purego uses code that originates from the Go runtime. These files are under the BSD-3
+License that can be found [here](https://github.com/golang/go/blob/master/LICENSE).
```
hajimehoshi, line 9: Hmm, let's avoid using 'here' as a link. What about 'in the Go source'?

README.md
```diff
@@ -2,3 +2,13 @@
...
+This is a list of the copied files:
+
+`zcallback_darwin_*.s` from package `runtime`
+
+`internal/abi/abi_*.h` from package `runtime/cgo`
```
hajimehoshi, line 14: Use a bullet list

```
 * `zcallback_darwin_*.s` from package `runtime`
 * `internal/abi/abi_*.h` from package `runtime/cgo`
```

internal/strings/strings.go
```diff
@@ -0,0 +1,23 @@
...
+
+import "unsafe"
+
+// hasSuffix tests whether the string s ends with suffix.
+func hasSuffix(s, suffix string) bool {
+	return len(s) >= len(suffix) && s[len(s)-len(suffix):] == suffix
+}
+
+// CString converts a go string to *byte that can be passed to C code.
+// if requireNil is true it will panic if the passed string doesn't have
```
hajimehoshi, line 14: Update this comment.

syscall_darwin.go
```diff
@@ -18,3 +26,120 @@ func syscall_syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2,
...
+			reflect.Interface, reflect.Func, reflect.Slice,
+			reflect.Chan, reflect.Complex64, reflect.Complex128,
+			reflect.String, reflect.Map, reflect.Invalid:
+			panic("purego: unsupported argument type: " + in.Kind().String())
+		}
+	}
+	if ty.NumOut() > 1 || ty.NumOut() == 1 && ty.Out(0).Size() != ptrSize {
+		panic("purego: callbacks can only have one pointer-sized return")
+	}
+	(&cbs.lock).Lock()
```
hajimehoshi, line 87: Can we write just `cbs.lock.Lock()`?

TotallyGamerJet: I copied this strange way of doing it from how the runtime on windows. I believe the reason was so the mutex never pointed to a null variable

hajimehoshi: OK so please leave a comment to explain why, or replace this with `cbs.lock.Lock`. I'm still not sure the reason.

internal/abi/abi_arm64.h
```diff
@@ -0,0 +1,37 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+// Macros for transitioning from the host ABI to Go ABI0.
+//
```
hajimehoshi, line 5: https://cs.opensource.google/go/go/+/master:src/runtime/cgo/abi_arm64.h?q=abi_arm64.h&ss=go%2Fgo Isn't this different from the original source?

hajimehoshi (approved):
LGTM % one comment

internal/abi/abi_arm64.h
TotallyGamerJet: I didn’t intentionally make any changes and they look the same to me. Are you seeing something I don’t?

hajimehoshi: There should be an empty line after the license comment

hajimehoshi: We need an empty line after

```
// Copyright 2021 The Go Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.
```

PR #13, opened 2022-07-28
don't split the stack on amd64

TotallyGamerJet:
It appears the Go runtime would occasionally decide to split the stack while the callbackArgs struct was on the unmovable system stack and so the pointer would then point to garbage memory. I did this right on arm64


hajimehoshi (approved):
LGTM, thanks!

PR #14, opened 2022-08-07
implement dlerror & dlclose

TotallyGamerJet:


dlfcn.go
```diff
@@ -17,20 +18,66 @@ const RTLD_GLOBAL = 0x8
...
+		// use unsafe.Add once we reach 1.17
+		//_ = (*byte)(unsafe.Add(unsafe.Pointer(msg), length))
+		if *(*byte)(unsafe.Pointer(msg + uintptr(length))) == '\x00' {
+			break
+		}
+		length++
+	}
+	// use unsafe.Slice once we reach 1.17
+	// we make a slice then cast to a string to force Go to copy the string from C memory into Go memory
+	return string(*(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{Data: msg, Len: length, Cap: length})))
```
hajimehoshi, line 67: What is the lifetime of msg?

hajimehoshi (commented):
I think it is better to split the PR into the documentation part and the Dlerror part.

dlfcn.go
TotallyGamerJet: It is valid until the next call to dlerror. Which is why we copy it into Go so that Go can decide when it no longer needs the string

hajimehoshi: Does this line really copy the string? Probably at casting from []byte to string?

TotallyGamerJet: Yeah a cast []byte to string copies bc it can’t prove that the []byte won’t be modified and break the immutability of string

TotallyGamerJet:
I will split these once I have a chance

dlfcn.go
TotallyGamerJet: I went ahead and made the copying more obvious

dlfcn.go
```diff
@@ -31,6 +32,32 @@ func Dlsym(handle uintptr, name string) uintptr {
 	return ret
 }
 
+func Dlerror() string {
+	msg, _, _ := SyscallN(dlerrorABI0)
```
hajimehoshi, line 36: Add a comment about the lifetime of `msg`.

hajimehoshi (approved):
LGTM % one comment

hajimehoshi:
So, are you sending a PR to add comments later? Thanks,

PR #15, opened 2022-08-11
add Ivar support and Objc Tests

TotallyGamerJet:
Ivar or InstanceVariables are like fields in a struct except for Objective-C objects. This PR add support for adding Ivars to classes. It also provides an example of how to use them properly.

This will be needed to fully port glfw to Go

### DIFF
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -7,6 +7,7 @@ package objc
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 	"runtime"
 	"unsafe"
@@ -159,12 +160,13 @@ func (c Class) AddMethod(name SEL, imp _IMP, types string) bool {
 // It may only be called after AllocateClassPair and before Register.
 // Adding an instance variable to an existing class is not supported.
 // The class must not be a metaclass. Adding an instance variable to a metaclass is not supported.
-// The instance variable's minimum alignment in bytes is 1<<align. The minimum alignment of an
-// instance variable depends on the ivar's type and the machine architecture.
-// For variables of any pointer type, pass math.Log2(unsafe.Alignof(type)).
-func (c Class) AddIvar(name string, size uintptr, alignment uint8, types string) bool {
+// It takes the instance of the type of the Ivar and a string representing the type.
+func (c Class) AddIvar(name string, ty interface{}, types string) bool {
 	n := strings.CString(name)
 	t := strings.CString(types)
+	typeOf := reflect.TypeOf(ty)
+	size := typeOf.Size()
+	alignment := uint8(math.Log2(float64(typeOf.Align())))
 	ret, _, _ := purego.SyscallN(class_addIvar, uintptr(c), uintptr(unsafe.Pointer(n)), size, uintptr(alignment), uintptr(unsafe.Pointer(t)))
 	runtime.KeepAlive(n)
 	runtime.KeepAlive(t)
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -21,15 +21,18 @@ import (
 var (
 	objc = purego.Dlopen("/usr/lib/libobjc.A.dylib", purego.RTLD_GLOBAL)
 
-	objc_msgSend           = purego.Dlsym(objc, "objc_msgSend")
-	objc_msgSendSuper2     = purego.Dlsym(objc, "objc_msgSendSuper2")
-	objc_getClass          = purego.Dlsym(objc, "objc_getClass")
-	objc_allocateClassPair = purego.Dlsym(objc, "objc_allocateClassPair")
-	objc_registerClassPair = purego.Dlsym(objc, "objc_registerClassPair")
-	sel_registerName       = purego.Dlsym(objc, "sel_registerName")
-	class_getSuperclass    = purego.Dlsym(objc, "class_getSuperclass")
-	class_addMethod        = purego.Dlsym(objc, "class_addMethod")
-	object_getClass        = purego.Dlsym(objc, "object_getClass")
+	objc_msgSend              = purego.Dlsym(objc, "objc_msgSend")
+	objc_msgSendSuper2        = purego.Dlsym(objc, "objc_msgSendSuper2")
+	objc_getClass             = purego.Dlsym(objc, "objc_getClass")
+	objc_allocateClassPair    = purego.Dlsym(objc, "objc_allocateClassPair")
+	objc_registerClassPair    = purego.Dlsym(objc, "objc_registerClassPair")
+	sel_registerName          = purego.Dlsym(objc, "sel_registerName")
+	class_getSuperclass       = purego.Dlsym(objc, "class_getSuperclass")
+	class_getInstanceVariable = purego.Dlsym(objc, "class_getInstanceVariable")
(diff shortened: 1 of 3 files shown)


objc/objc_runtime_darwin.go
```diff
@@ -152,17 +155,53 @@ func (c Class) AddMethod(name SEL, imp _IMP, types string) bool {
...
+// The class must not be a metaclass. Adding an instance variable to a metaclass is not supported.
+// The instance variable's minimum alignment in bytes is 1<<align. The minimum alignment of an
+// instance variable depends on the ivar's type and the machine architecture.
+// For variables of any pointer type, pass math.Log2(unsafe.Alignof(type)).
+func (c Class) AddIvar(name string, size uintptr, alignment uint8, types string) bool {
+	n := strings.CString(name)
+	t := strings.CString(types)
+	ret, _, _ := purego.SyscallN(class_addIvar, uintptr(c), uintptr(unsafe.Pointer(n)), size, uintptr(alignment), uintptr(unsafe.Pointer(t)))
+	runtime.KeepAlive(n)
+	runtime.KeepAlive(t)
```
hajimehoshi, line 170: Do we need `runtime.KeepAlive` for C pointers? I don't think so. Rather, we have to free them explicitly like

```
n := strings.CString(name)
defer C.free(unsafe.Pointer(n))
```

objc/objc_runtime_darwin_test.go
```diff
@@ -0,0 +1,59 @@
+package objc
```
hajimehoshi, line 1: Add a license comment.

objc/objc_runtime_darwin.go
hajimehoshi: I realized there are similar things in other functions. Please update them. Thanks,

hajimehoshi: Or I'll send a PR for other existing part as Ebitengine already depends on it and this might cause memory leak so the fix is necessary asap...

TotallyGamerJet: This isn't a C pointer. It's a pointer to Go memory. The function definition is in internal/strings

hajimehoshi: Oh sorry I was misunderstanding.. Thanks!

syscall_darwin.go
```diff
@@ -116,7 +116,20 @@ func callbackWrap(a *callbackArgs) {
...
+		case reflect.Int, reflect.Int64, reflect.Int32, reflect.Int16, reflect.Int8:
+			a.result = uintptr(ret[0].Int())
+		case reflect.Bool:
+			if ret[0].Bool() {
+				a.result = 1
+			} else {
+				a.result = 0
+			}
+		default:
+			panic("purego: unsupported kind: %s" + k.String())
```
hajimehoshi, line 131: What is %s?

TotallyGamerJet: I had originally written using fmt but realized all the rest of the panics didn't so I removed it and forgot to remove the %s

objc/objc_runtime_darwin_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func ExampleAllocateClassPair() {
+	var class = AllocateClassPair(GetClass("NSObject\x00"), "FooObject\x00", 0)
+	class.AddMethod(RegisterName("run\x00"), IMP(func(self ID, _cmd SEL) {
+		fmt.Println("Hello World!")
+	}), "v@:\x00")
+	class.Register()
+
+	var FooObject = ID(class).Send(RegisterName("new\x00"))
+	FooObject.Send(RegisterName("run\x00"))
```
hajimehoshi, line 31: Just a qeustion: is the function `run` a instance method or a static method? If this is an instance method, `fooObject` as a Go variable is better than `FooObject`.

objc/objc_runtime_darwin_test.go
```diff
@@ -0,0 +1,62 @@
...
+		return *(*int)(unsafe.Pointer(uintptr(unsafe.Pointer(self)) + barOffset))
+	}), "q@:\x00")
+	class.AddMethod(RegisterName("setBar:\x00"), IMP(func(self ID, _cmd SEL, bar int) {
+		*(*int)(unsafe.Pointer(uintptr(unsafe.Pointer(self)) + barOffset)) = bar
+	}), "v@:q\x00")
+	class.Register()
+
+	var BarObject = ID(class).Send(RegisterName("new\x00"))
+	BarObject.Send(RegisterName("setBar:\x00"), 123)
+	var bar = int(BarObject.Send(RegisterName("bar\x00")))
```
hajimehoshi, line 49: ditto

objc/objc_runtime_darwin_test.go
TotallyGamerJet: These are instance methods

objc/objc_runtime_darwin.go
```diff
@@ -152,17 +155,53 @@ func (c Class) AddMethod(name SEL, imp _IMP, types string) bool {
...
 }
 
-// Register registers a class that was allocated using objc_allocateClassPair.
+// AddIvar adds a new instance variable to a class.
+// It may only be called after AllocateClassPair and before Register.
+// Adding an instance variable to an existing class is not supported.
+// The class must not be a metaclass. Adding an instance variable to a metaclass is not supported.
+// The instance variable's minimum alignment in bytes is 1<<align. The minimum alignment of an
+// instance variable depends on the ivar's type and the machine architecture.
+// For variables of any pointer type, pass math.Log2(unsafe.Alignof(type)).
```
hajimehoshi, line 164: My understanding is that we cannot have a utility function to calculate the size due to `unsafe.Alignof` like this

```go
func IvarSize(value interface{}) uint8 {
    return uint8(math.Log2(unsafe.Alignof(value)))
}
```

Instead, can we have this utility function?

```go
func IvarSizeFromAlign(align uintptr) uint8 {
    return uint8(math.Log2(align))
}
```

hajimehoshi: Another option is to calcuate `math.Log2` in `AddIvar` so that the user doesn't have to care.

TotallyGamerJet: Well according to the docs `reflect.TypeOf(x).Align()` is equal to `unsafe.Alignof`. So it would be possible to do what you first suggested. Although I think it should one called `IvarAlignment` instead of `IvarSize` since it is calculating the alignment.

In your second comment are you referring to taking just unsafe.Alignof for the alignment parameter and then do math.Log2 inside AddIvar?

Perhaps, we make the signature `AddIvar(name string, ty interface{}, types string) bool` and then inside AddIvar calculate the size and alignment using reflect.

hajimehoshi: > In your second comment are you referring to taking just unsafe.Alignof for the alignment parameter and then do math.Log2 inside AddIvar?

Yes.

> Perhaps, we make the signature AddIvar(name string, ty interface{}, types string) bool and then inside AddIvar calculate the size and alignment using reflect.

This seems the best idea so far. What do you think?

TotallyGamerJet: I do like it. It makes using it a lot easier and makes it harder to use improperly.

hajimehoshi (approved):
LGTM, thanks!

PR #16, opened 2022-08-11
objc: change the test package

hajimehoshi:
For testing, a package with the suffix `_test` is preferrable since it prevents accidental white-box testings.

### DIFF
--- a/objc/objc_runtime_darwin_test.go
+++ b/objc/objc_runtime_darwin_test.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-package objc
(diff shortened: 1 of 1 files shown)


hajimehoshi:
@TotallyGamerJet Please take a look. Thanks,

TotallyGamerJet:
I've never seen this pattern before. Thanks for showing me something new. It looks good to me. One question, will these examples still show up on go.dev even though they are in a different package? If so, I think it's good to merge

hajimehoshi:
> One question, will these examples still show up on go.dev even though they are in a different package?

Yes. For example,

https://pkg.go.dev/strconv#example-AppendBool
https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/strconv/example_test.go;l=13-20

PR #18, opened 2022-08-13
objc: support string arguments to Send/SendSuper

TotallyGamerJet:


objc/objc_runtime_darwin_test.go
```diff
@@ -20,42 +21,62 @@ func init() {
...
+		fmt.Println("In Child")
+		self.SendSuper(_cmd)
+	}), "v@:\x00")
+	child.Register()
+
+	objc.ID(child).Send(objc.RegisterName("new\x00")).Send(objc.RegisterName("doSomething\x00"))
+
+	// Output: In Child
+	// In Super!
+}
```
hajimehoshi, line 82: Option: what about not using \x00 in the examples?

hajimehoshi (approved):
LGTM, thanks!

objc/objc_runtime_darwin_test.go
TotallyGamerJet: Sure!

hajimehoshi (approved):
LGTM, thanks!

PR #19, opened 2022-08-16
purego: use a safe way to convert a pointer to a byte slice

hajimehoshi:
Creating a reflect.SliceHeader is invalid. See https://pkg.go.dev/unsafe#Pointer


hajimehoshi:
@TotallyGamerJet Please take a look, thanks!

TotallyGamerJet:
LGTM!

Issue #12, opened 2022-07-24
Implement fakecgo


TotallyGamerJet:
This issue tracks the implementation of fakecgo which is needed for some frameworks to function properly when CGO_ENABLED=0. It will be similar to [notti/nocgo](https://github.com/notti/nocgo/tree/master/internal/fakecgo)

TotallyGamerJet:
The next step for this issue is to either implement it for iOS and/or linux

hajimehoshi:
It's up to you which way you go with :-)

I think porting GLFW is another big issue.

TotallyGamerJet:
I'll probably port the rest of Ebitengine which includes internal/ui/ui_glfw_darwin.go and internal/gamepad/gamepad_darwin.go before attempting GLFW.

hajimehoshi:
Sure!

TotallyGamerJet:
So I think this can be closed as fakecgo is implemented on Linux and macOS. @hajimehoshi agree?

hajimehoshi:
Yes!

PR #20, opened 2022-08-18
fakecgo support on macOS arm64 and amd64

TotallyGamerJet:
This adds fakecgo support for darwin on amd64 and arm64 and updates #12. This means that Oto can now be built using `CGO_ENABLED=0`.

How does it work?
* If `CGO_ENABLED=1` then `runtime/cgo` is built into the binary which requires a C compiler.
* if `CGO_ENABLED=0` then `internal/fakecgo` is built into the binary which implements everything `runtime/cgo` does but entirely in Go requiring only the Go compiler.

The only limitation right now is that it does not work on iOS so it is important to still use `import _ "runtime/cgo"` to force Go's Cgo runtime on iOS.


hajimehoshi:
> The only limitation right now is that it does not work on iOS so it is important to still use import _ "runtime/cgo" to force Go's Cgo runtime on iOS.

Do you have a plan to remove Cgo usage for iOS in the future?

hajimehoshi:
Another question: do we have to update this whenever a new Go runtime is released?

TotallyGamerJet:
> > The only limitation right now is that it does not work on iOS so it is important to still use import _ "runtime/cgo" to force Go's Cgo runtime on iOS.
> 
> Do you have a plan to remove Cgo usage for iOS in the future?

Perhaps I’m the future yes. I think it’s not too big of an issue since most people tend to use Cgo when using Go on iOS.

> do we have to update this whenever a new Go runtime is released?

I don’t except any major changes since what this PR does is rewrite the C code in Go. This C code is quite standard and hasn’t changed much. This would only have to change if Go changed how the entire Cgo runtime worked which is unlikely since it’d require all GOOS to be updated too.

TotallyGamerJet:
Plus I’ve tried to take as much as possible from `runtime/cgo` to make it easier to maintain and to port it to other GOOSs and GOARCHs

TotallyGamerJet:
Oh there is one more limitation. You can’t use `CGO_ENABLED=0` and `buildmode=c-archive` because that requires special initialization that this PR doesn’t do. Although I don’t see this as a major issue since you can just enable Cgo and everything will work. Plus you’ll probably be using `//export` which purego can’t do

hajimehoshi:
> I don’t except any major changes since what this PR does is rewrite the C code in Go. This C code is quite standard and hasn’t changed much. This would only have to change if Go changed how the entire Cgo runtime worked which is unlikely since it’d require all GOOS to be updated too.

Thanks!

> Oh there is one more limitation. You can’t use CGO_ENABLED=0 and buildmode=c-archive because that requires special initialization that this PR doesn’t do. Although I don’t see this as a major issue since you can just enable Cgo and everything will work. Plus you’ll probably be using //export which purego can’t do

I see, so it would not make sense to do much effort to remove Cgo for iOS.

go_runtime.go
```diff
@@ -10,9 +10,5 @@ import (
...
 )
 
-//go:linkname runtime_libcCall runtime.libcCall
-//go:linkname runtime_entersyscall runtime.entersyscall
-//go:linkname runtime_exitsyscall runtime.exitsyscall
-func runtime_libcCall(fn uintptr, arg unsafe.Pointer) int32 // from runtime/sys_libc.go
-func runtime_entersyscall()                                 // from runtime/proc.go
-func runtime_exitsyscall()                                  // from runtime/proc.go
+//go:linkname runtime_cgocall runtime.cgocall
+func runtime_cgocall(fn uintptr, arg unsafe.Pointer) int32 // from runtime/sys_libc.go
```
hajimehoshi, line 14: Is this function available even though `runtime/cgo` is not imported?

hajimehoshi: It looks like so.

nocgo.go
```diff
@@ -0,0 +1,12 @@
...
+
+//go:build !cgo && !ios
+// +build !cgo,!ios
+
+package purego
+
+// if CGO_ENABLED=0 import fakecgo to setup the Cgo runtime correctly.
+// This is required since some frameworks need TLS setup the C way which Go doesn't do.
+// We currently don't support ios in fakecgo mode so force Cgo or fail
+import _ "github.com/ebitengine/purego/internal/fakecgo"
```
hajimehoshi, line 12: I'm not sure why just importing this package works (this is magical!) Could you elaborate this?

hajimehoshi: My understanding is that fakecgo does what `runtime/cgo` does instead of it. Is this correct?

hajimehoshi: So, does fakecgo creates these symbols instead of runtime/cgo?

https://cs.opensource.google/go/go/+/master:src/runtime/cgo.go;l=13-19;drc=2580d0e08d5e9f979b943758d3c49877fb2324cb

nocgo.go
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build !cgo && !ios
+// +build !cgo,!ios
```
hajimehoshi, line 5: Isn't `!cgo` enough? It's because `!cgo` & `ios` is an impossible combination so far.

TotallyGamerJet: This way, running `CGO_ENABLED=0 GOOS=ios go build ./cmd/main.go`
```
# command-line-arguments
/Users/jarrettkuklis/go/go1.19/pkg/tool/darwin_arm64/link: running clang failed: exit status 1
Undefined symbols for architecture arm64:
  "__cgo_yield", referenced from:
      _runtime.cgo_yield in go.o
  "_x_cgo_init", referenced from:
      __cgo_init in go.o
  "_x_cgo_notify_runtime_init_done", referenced from:
      __cgo_notify_runtime_init_done in go.o
  "_x_cgo_setenv", referenced from:
      _runtime._cgo_setenv in go.o
  "_x_cgo_thread_start", referenced from:
      __cgo_thread_start in go.o
  "_x_cgo_unsetenv", referenced from:
      _runtime._cgo_unsetenv in go.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
If `ios` is removed from `nocgo.go` then you get this:
`2022/08/19 09:02:27 duplicated definition of symbol _cgo_init, from github.com/ebitengine/purego/internal/fakecgo and runtime/cgo`.

Which error is more helpful?

nocgo.go
TotallyGamerJet: Yeah that's the basic idea. All you need to do is `go:linkname` each of those symbols with a Go version (well, asm since you need to map from GCC ABI to Go ABI). We don't replace all of them just the one's listed below:
```go
iscgo             bool
_cgo_init         unsafe.Pointer
_cgo_thread_start unsafe.Pointer
_cgo_notify_runtime_init_done unsafe.Pointer
_cgo_setenv unsafe.Pointer  // these two are in runtime/env_posix.go
_cgo_unsetenv unsafe.Pointer
```
[notti/nocgo](https://github.com/notti/nocgo#fakecgo) contains a pretty good explanation of how his fakecgo works which this PR is inspired by.

TotallyGamerJet: It's just enough to pass this `iscgo` test in [runtime/proc.go](https://github.com/golang/go/blob/c82bbc0e8edbbebe47e92729e8f3f1b60d380b5b/src/runtime/proc.go#L212) and setup thread-local storage properly.

nocgo.go
hajimehoshi: I think both are not so helpful and don't have significant change. Then, I slighly prefer not having ios. What do you think?

TotallyGamerJet: I think removing ios is better bc it at least mentions fakecgo so it’d be easier to spot and suggest a fix (ie Enable Cgo)

hajimehoshi: It is possible to show a pseudo compile error message like this

```go
//go:build !cgo && ios
// +build !cgo,ios

var _ = _IOS_REQUIRES_CGO
```

This is optional. I'm fine with both having and not having this.

TotallyGamerJet: I like it! I also added a comment that describes what the error means.

cgo.go
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build cgo || ios
+// +build cgo ios
```
hajimehoshi, line 5: Just cgo

hajimehoshi: Or `cgo && darwin`? If so, please add a suffix for this file

nocgo.go
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build !cgo && darwin
+// +build !cgo,darwin
```
hajimehoshi, line 5: Please add a suffix '_darwin.go' to this file instead of specifying darwin here.

cgo.go
TotallyGamerJet: It should just be cgo

cgo.go
```diff
@@ -0,0 +1,12 @@
...
+
+//go:build cgo
+// +build cgo
+
+package purego
+
+// if CGO_ENABLED=1 import the Cgo runtime to ensure that it is set up properly.
+// This is required since some frameworks need TLS setup the C way which Go doesn't do.
+// We currently don't support ios in fakecgo mode so force Cgo or fail
+import _ "runtime/cgo"
```
hajimehoshi, line 12: Do we really need this file? What if we don't have this file but a user specifies cgo? In this case, `runtime/cgo` is automatically imported, right?

isios_darwin.go
```diff
@@ -0,0 +1,11 @@
+//go:build !cgo && ios
```
hajimehoshi, line 1: Add a license comment.

cgo.go
TotallyGamerJet: We do need it. Even if `CGO_ENABLED=1` if there are no usages of `import "C"` or `import "runtime/cgo` then the Cgo runtime is NOT compiled in. Normally, it would not be possible to call into C without using one of those imports but with purego it is.

hajimehoshi: I see, that's a good point.

hajimehoshi:
I'll take a look further tomorrow. Thank you for your hard work!

hajimehoshi:
Will fakecgo work for non-Darwin environments like Linux in the future?

TotallyGamerJet:
Yeah most of the code I’d the same. There are only two platform specific files: the symbols_GOOS file and go_GOOS_GOARCH. The second of which is quite similar between different GOOSs. So yeah, it should be easy

internal/fakecgo/go_util.go
```diff
@@ -0,0 +1,30 @@
...
+// _cgo_thread_start(ThreadStart *arg) (runtime/cgo/gcc_util.c)
+// This get's called instead of the go code for creating new threads
+// -> pthread_* stuff is used, so threads are setup correctly for C
+// If this is missing, TLS is only setup correctly on thread 1!
+// This function should be go:systemstack instead of go:nosplit (but that requires runtime)
+//
+//go:nosplit
+func x_cgo_thread_start(arg *ThreadStart) {
+	var ts *ThreadStart
+	/* Make our own copy that can persist after we return. */
```
hajimehoshi, line 19: Let's use `//` comments rather than `/* */`

internal/fakecgo/trampolines_amd64.s
```diff
@@ -0,0 +1,150 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+/*
+    trampoline for emulating required C functions for cgo in go (see cgo.go)
```
hajimehoshi, line 5: We don't need indentations in this comment.

internal/fakecgo/go_util.go
```diff
@@ -0,0 +1,30 @@
...
+	//	_cgo_tsan_acquire();
+	ts = (*ThreadStart)(malloc(unsafe.Sizeof(*ts)))
+	//	_cgo_tsan_release();
+	if ts == nil {
+		println("fakecgo: out of memory in thread_start")
+		abort()
+	}
+	// *ts = *arg would cause a writebarrier so use memmove instead
+	memmove(unsafe.Pointer(ts), unsafe.Pointer(arg), unsafe.Sizeof(*ts))
+	_cgo_sys_thread_start(ts) /* OS-dependent half */
```
hajimehoshi, line 29: Use //

hajimehoshi:
I'll try this PR with my actual devices later. Thanks!

TotallyGamerJet:
Awesome! If you test with Oto makes sure to remove the `import "runtime/cgo"` in the api file and set CGO_ENABLED=0

hajimehoshi:
I'll take a look tomorrow. Sorry to keep you waiting.

hajimehoshi:
Could you rebase this PR? It's beacuse there was a breaking change around `IMP` and then the current PR is not available for the latest Oto and Ebitengine.

hajimehoshi (approved):
LGTM, thanks!

hajimehoshi:
I'm fine to merge this by me, but you are the best person to edit the comment and merge it for your PR.

I'd be happy if you could follow these:

* Add a prefix `internal/fakecgo: ` to the title
* Include `Closes #12` or `Updates #12` in the comment

hajimehoshi:
I have confirmed this PR worked with a modified Oto without `runtime/cgo` :D

```
~/oto 
$ CGO_ENABLED=0 go run -tags=example ./example
```

TotallyGamerJet:
This is very exciting! It's the first step towards removing Cgo entirely from ebitengine! :D

PR #24, opened 2022-08-22
remove go:nosplit at SyscallN

TotallyGamerJet:
this is allowed because SyscallN also uses go:uintptrescapes which forces any uintptr's that are passed to this function to be placed on the heap where a stack split won't affect them. See the [compile docs](https://pkg.go.dev/cmd/compile) for an explanation.

I've tested this change on Oto (amd64 & arm64) without optimizations. I also ran all the test for ebitengine (amd64 & arm64) with OpenGL and Metal.

All this to say, this won't break anything.

This change fixes [Oto#182](https://github.com/hajimehoshi/oto/issues/182).

### DIFF
--- a/syscall.go
+++ b/syscall.go
@@ -20,7 +20,8 @@ const maxArgs = 9
 // stack.
 //
 // the pragma go:nosplit is not needed at this function declaration because it uses go:uintptrescapes
-// to push all the uintptr onto the heap where a stack split won't affect their memory location.
+// which forces all the objects that the uintptrs point to onto the heap where a stack split won't affect
+// their memory location.
 //
 //go:uintptrescapes
 func SyscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
--- a/syscall.go
+++ b/syscall.go
@@ -18,7 +18,10 @@ const maxArgs = 9
 //
 // On amd64, if there are more than 8 floats the 9th and so on will be placed incorrectly on the
 // stack.
-//go:nosplit
+//
+// the pragma go:nosplit is not needed at this function declaration because it uses go:uintptrescapes
+// to push all the uintptr onto the heap where a stack split won't affect their memory location.
+//
 //go:uintptrescapes
 func SyscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 	if fn == 0 {
--- a/syscall.go
+++ b/syscall.go
@@ -19,7 +19,7 @@ const maxArgs = 9
 // On amd64, if there are more than 8 floats the 9th and so on will be placed incorrectly on the
 // stack.
 //
-// the pragma go:nosplit is not needed at this function declaration because it uses go:uintptrescapes
+// The pragma go:nosplit is not needed at this function declaration because it uses go:uintptrescapes
 // which forces all the objects that the uintptrs point to onto the heap where a stack split won't affect
 // their memory location.
 //


hajimehoshi (approved):
LGTM, thanks! (Please merge this with your comments. The title should have `purego: ` prefix)

syscall.go
```diff
@@ -18,7 +18,10 @@ const maxArgs = 9
 //
 // On amd64, if there are more than 8 floats the 9th and so on will be placed incorrectly on the
 // stack.
-//go:nosplit
+//
+// the pragma go:nosplit is not needed at this function declaration because it uses go:uintptrescapes
+// to push all the uintptr onto the heap where a stack split won't affect their memory location.
```
hajimehoshi, line 23: Doesn't `go:uintptrescapese` assure that objects indicated by the uintptr pointers are in the heap rather than the uintptrs are in the heap?

hajimehoshi (changes requested):
Sorry but I left one question here... Please take a look at my comment

syscall.go
TotallyGamerJet: yeah this is true. what I wrote wasn't clear

syscall.go
```diff
@@ -18,7 +18,11 @@ const maxArgs = 9
 //
 // On amd64, if there are more than 8 floats the 9th and so on will be placed incorrectly on the
 // stack.
-//go:nosplit
+//
+// the pragma go:nosplit is not needed at this function declaration because it uses go:uintptrescapes
```
hajimehoshi, line 22: `The pragma`...

syscall.go
```diff
@@ -18,7 +18,11 @@ const maxArgs = 9
...
+// which forces all the objects that the uintptrs point to onto the heap where a stack split won't affect
+// their memory location.
```
hajimehoshi, line 24: So, my understanding is that the objects pointed by uintptrs are on the heap with fixed, not on the stack, and the objects' addresses never change during the call, and that's why we're safe without nosplit. Is that correct?

hajimehoshi: If a pointer addresses to an object on the stack, what would happen? (A compiile error, or a runtime error?)

hajimehoshi: I have confirmed that `uintptrespcapese` changes the escape analysis, so just having this annotation seems fine:

With `//go:uintptrescapes`

```go
package main

import (
        "unsafe"
)

//go:uintptrescapes
func Foo(x uintptr) {
}

func main() {
        var x int
        Foo(uintptr(unsafe.Pointer(&x)))
}
```

```
$ go run -gcflags="-m -N -l" escape.go 
# command-line-arguments
./escape.go:8:10: marking x as escaping uintptr
./escape.go:12:6: moved to heap: x
```

Without `//go:uintptrescapes`

```go
package main

import (
        "unsafe"
)

func Foo(x uintptr) {
}

func main() {
        var x int
        Foo(uintptr(unsafe.Pointer(&x)))
}
```

```
$ go run -gcflags="-m -N -l" escape.go
```

TotallyGamerJet: You are correct as to how uintptrescapes works. If the pointer points to a stack object and there is no stack split then it is completely safe. If there is a stack split then the program can crash at runtime bc the C function might do just about anything with that memory which we don't know what its pointing to

hajimehoshi: Thanks!

hajimehoshi (approved):
LGTM

Issue #31, opened 2022-09-05
internal/abi/abi_arm64.h Not included when vendoring dependencies


mechanical-lich:
I noticed this issue when upgrading Ebiten from 2.3.8 to 2.4.0, however this was the dependency throwing the error.  So I hope this is the right place to report this. 
Error: `# github.com/ebitengine/purego
vendor/github.com/ebitengine/purego/sys_darwin_arm64.s:8: #include: open /Users/<redacted>/sdk/go1.19/pkg/include/internal/abi/abi_arm64.h: no such file or directory`

Digging further I found that the folder `/internal/abi/` didn't exist in my vendored copy but `strings` and `fakecgo` did.  Which led me to [this comment](https://github.com/goccy/go-graphviz/issues/28#issuecomment-674405467) that mentions that there needs to be at least one .go file present in a directory in order for all the files to be included during vendoring.    I made sure it wasn't a fluke by deleting the vendor folder and starting over but the results were consistent.

There is a work around, I cloned the repo and manually copied it to the vendor folder and the issue resolved (until I vendor again)

hajimehoshi:
I'm not familiar with vendoring, but would putting a dummy Go file in `internal/abi` solve this vendoring issue?

/CC @TotallyGamerJet

hajimehoshi:
https://github.com/goccy/go-graphviz/pull/37/files

Hm, I see. Do we need a `required` build tag for this?

hajimehoshi:
I've upated purego.

I'll update Oto and Ebitengine and then close this issue.

hajimehoshi:
Until v2.4.2 is released, please try:

```
go get github.com/hajimehoshi/ebiten/v2@557bef998d148e4e19da0a70c1106c206aff0cef
```

Thanks,

mechanical-lich:
Sorry for the late response!   I really appreciate you looking into this and thanks for your help!

PR #32, opened 2022-09-05
internal/abi: add a dummy Go file for go mod vendor

hajimehoshi:
Before this change, the package `internal/abi` was ignored by `go mod vendor`
as there was no Go file there. Then building purego failed as `internal/abi`
didn't exist under the `vendor` directory.

This change fixes this issue by adding a dummy Go file to `internal/abi`
and adding an explicit import to it, but with a `dummy` build tag.

I referred a change https://github.com/goccy/go-graphviz/pull/37 for this
hack.

Updates #31

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,7 @@ jobs:
           echo 'package main' > main.go
           echo 'import (' >> main.go
           echo '  _ "github.com/ebitengine/purego"' >> main.go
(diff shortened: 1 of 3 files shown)


hajimehoshi:
@TotallyGamerJet Please take a look, thanks!

TotallyGamerJet (approved):
This is the correct solution to this problem. LGTM! Thanks

PR #34, opened 2022-09-06
fix callbacks on go 1.15

TotallyGamerJet:
This passes all the objc tests excepts ExampleClass_AddIvar which fails because barOffset is wrong in the callback. Trying to find the cause of the issue may take some time. First, I wanted to see if you think it's worth supporting Go 1.15 with the amount of complexity it adds?


hajimehoshi:
Hm, so we would need completely different implementation for Go 1.15, right?

By the way, what do you think about my previous idea ('fake purego' for Go 1.15)? I feel like we can drop Go 1.15 support, if this would be hard.

Thank you for tackling this!

TotallyGamerJet:
So it's likely possible but I'm not sure how much it will help reduce code duplication. I can sketch up a simple PR so you can make the best judgement

TotallyGamerJet:
I looked into it and using Cgo would make the code more complicated. It now works fully and I simplified as much as could be. Because of the way Go 1.15 calls functions it needs a `**callbackArgs` while Go 1.16+ doesn't hence the conversion in `callbackWrapPicker()`.

It's up to you if you think this is a good idea or not. It still does make the code a little more complicated but definitely not too much.

callback_go1_15_darwin_amd64.go
```diff
@@ -0,0 +1,13 @@
+//go:build !go1.16
```
hajimehoshi, line 1: Add a license comment.

callback_go1_16_darwin_amd64.go
```diff
@@ -0,0 +1,21 @@
+//go:build go1.16
```
hajimehoshi, line 1: Add a license comment.

callback_go1_15_darwin_amd64.go
hajimehoshi, line 1: This is just my preference, but what about the file name like

```
callback_go115_darwin_amd64.go
```

?

callback_go1_16_darwin_amd64.go
hajimehoshi, line 1:
```
callback_go116_darwin_amd64.go
```

?

hajimehoshi (commented):
The change seems much more concise than I expected. I would like to accept this change and continue supporting Go 1.15 with this change.

hajimehoshi:
> I looked into it and using Cgo would make the code more complicated.

Acknowledged.

callback_go1_15_darwin_amd64.go
TotallyGamerJet: Or maybe just `callback_go15_darwin_amd64.go` since go2 won't happen anytime soon?

hajimehoshi: Hmm, `callback_go1_15_darwin_amd64.go` would be clearer than go15 (or even go115).

OK, let's keep with the current name `callback_go1_15_darwin_amd64.go`. Sorry for going back and forth.

hajimehoshi (approved):
LGTM, thanks!

hajimehoshi:
I'm amazed at your work. Thank you very much!

PR #35, opened 2022-09-07
arm64 broke

TotallyGamerJet:
I forgot callbackWrapPicker needs to be implemented on arm64

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,10 @@ jobs:
         run: |
           # Check cross-compiling Windows binaries.
           env GOOS=windows GOARCH=arm64 go build -v ./...
+          
+          # Check cross-compiling macOS binaries.
+          env GOOS=darwin GOARCH=amd64 go build -tags=example -v ./...
+          env GOOS=darwin GOARCH=arm64 go build -tags=example -v ./...
 
       - name: go mod vendor
         run: |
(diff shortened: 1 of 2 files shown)


TotallyGamerJet:
It also might be a good idea to test cross compilation to arm64

hajimehoshi (commented):
Can we integrate this and `callback_go1_16_darwin_amd64.go` to `callback_go1_16_darwin.go` ?

hajimehoshi:
> It also might be a good idea to test cross compilation to arm64

I agree. I'm fine you add this in this PR, or I'll do in a later PR.

TotallyGamerJet:
> Can we integrate this and `callback_go1_16_darwin_amd64.go` to `callback_go1_16_darwin.go` ?

We could but it would mean changing the assembly to purposefully make an indirection. Do you want that?

hajimehoshi:
> We could but it would mean changing the assembly to purposefully make an indirection. Do you want that?

I was wondering why we would need assembly changes by integrating... Sorry if I am missing something

TotallyGamerJet:
Because amd64 is using the Go 1.15 mechanism of double indirection (`**callbackArgs`) while arm64 is using 1.16 mechanism of single indirection (`*callbackArgs`). They are not the exact same file.

hajimehoshi:
Aha, I missed such a slight difference. Thank you!

hajimehoshi (approved):
LGTM, thanks!

PR #36, opened 2022-09-07
.githug/workflows: simplify test.yml

hajimehoshi:
* A build tag `example` is not used anywhre.
* Building with GOOO=darwin and GOARCH=arm64 are duplicated.
* Tests for iOS were missing

In the ideal world, we should run the binary with ARM64 for macOS,
in addition to the buildings.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,13 +57,13 @@ jobs:
           env GOOS=windows GOARCH=arm64 go build -v ./...
 
       - name: go build (iOS, Go 1.15)
-        if : ${{ startsWith(matrix.go, '1.15.') && startsWith(matrix.os, 'macos-') && }}
+        if : ${{ startsWith(matrix.go, '1.15.') && startsWith(matrix.os, 'macos-') }}
         run: |
           env CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 go build -v -tags=ios ./...
           env CGO_ENABLED=1 GOOS=darwin GOARCH=arm64 go build -v -tags=ios ./...
 
-      - name: go build (iOS, Go 1.16)
-        if : ${{ !startsWith(matrix.go, '1.15.') && startsWith(matrix.os, 'macos-') && }}
+      - name: go build (iOS, Go 1.16 and newer)
+        if : ${{ !startsWith(matrix.go, '1.15.') && startsWith(matrix.os, 'macos-') }}
         run: |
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Please take a look, thanks!

hajimehoshi:
Oh, (building succeeded but) does purego work with Go 1.15 for darwin/arm64?

TotallyGamerJet:
Would that be iOS? If so I haven't tested that

hajimehoshi:
> Would that be iOS?

Yes, for iOS. Probably not for M1 but I am not sure.

TotallyGamerJet:
If it is iOS I can guarantee that it won't work since the arm64 is doing what amd64 did. 

It should be a relatively easy fix since I've already done this once.

Will we still need to revert Oto again? This only effects people building for iOS on Go 1.15 which is only on macOS amd64. Unless you can build iOS apps from Linux?

hajimehoshi:
Building an iOS app actually requires Cgo for c-archive, so this would be a problem only when someone is trying to use Go 1.15 for ARM macOS. I don't think this is urgent and the change doesn't have to be reverted.

In the first place, I was wondering why the current tests passed (building with Go 1.15 for ARM) Do you have any insights?

hajimehoshi:
https://go.dev/doc/go1.16

Apparently ARM macOS is supported as of Go 1.16.

hajimehoshi:
https://github.com/ebitengine/purego/runs/8226036114

So buliding for iOS w/ Go 1.15 doesn't work now...

hajimehoshi:
My understanding is:

* Go 1.15 + amd64 for macOS works fine.
* Go 1.15 + amd64 for iOS works fine.
* Go 1.15 + arm64 succeeds to build but would not work as `callbackWrapPicker` is not defined.
* Go 1.15 + arm64 for macOS doesn't work, but it is OK since Go 1.15 doesn't support Arm macOS.
* Go 1.15 + arm64 for iOS doesn't work. This is actually an issue but low priority to fix since we don't aim to cross-compile for iOS.

This PR just simplifies the current Yaml and should not make the situation worse. I'd like to merge if you are fine.

Thanks,

TotallyGamerJet:
I am confused as to how `GOOS=darwin GOARCH=arm64` builds successfully without callbackWrapPicker for Go 1.15?

hajimehoshi:
> I am confused as to how GOOS=darwin GOARCH=arm64 builds successfully without callbackWrapPicker for Go 1.15?

I think it's because this function is referred only from .s files.

PR #38, opened 2022-09-07
add linux support (amd64 & arm64)

TotallyGamerJet:
I was curious how difficult it would be to add basic linux support because of our discussion in [ebitengine#2284](https://github.com/hajimehoshi/ebiten/issues/2284). I whipped this up but I don't have a linux computer rn to test. I did build an executable though for amd64. Arm64 gives the following error:

```
# command-line-arguments
dlopen: unsupported dynamic relocation for symbol _dlopen (type=10 (R_CALLARM64) stype=45 (SDYNIMPORT))
dlerror: unsupported dynamic relocation for symbol _dlerror (type=10 (R_CALLARM64) stype=45 (SDYNIMPORT))
dlopen: unhandled relocation for _dlopen (type 45 (SDYNIMPORT) rtype 10 (R_CALLARM64))
dlerror: unhandled relocation for _dlerror (type 45 (SDYNIMPORT) rtype 10 (R_CALLARM64))

```


TotallyGamerJet:
I think this code should work on linux (amd64 & arm64) although it can't be cross compiled in Go 1.15 because the internal linker doesn't support the dynamic relocations but can in 1.16+. Probably the best thing to do is to then have a Cgo backup for 1.15 so that it will still compile if there is a C cross-compiler

hajimehoshi:
I see. What happens if this is used with other architectures like mips?

hajimehoshi:
Ideally, a user should not have to care about purego supports the Go versions and the architecture and can always use purego, and purego has a C fallback for such environments.

TotallyGamerJet:
> I see. What happens if this is used with other architectures like mips?

Well rn it doesn't compile since the C fallback isn't added yet

hajimehoshi:
> Well rn it doesn't compile since the C fallback isn't added yet

I see. I hope adding it would not be difficult....

Thank you for your prototyping!

hajimehoshi:
> I think this code should work on linux (amd64 & arm64) although it can't be cross compiled in Go 1.15 because the internal linker doesn't support the dynamic relocations but can in 1.16+.

I'm not familiar with linkers. What are dynamic relocations? Is this the key to enable `dlopen` without Cgo?

TotallyGamerJet:
The linker just creates the final executable in the proper format. A dynamic relocation is a special function pointer that when called will have the dynamic linker find where that function is and jump to it.

hajimehoshi:
I see... I'll learn it more. Thanks!

TotallyGamerJet:
This code works for Linux arm64 and amd64 with `CGO_ENABLED=0` and `CGO_ENABLED=1`!

However, when cgo in enabled you must use internal linking by the command: 

`CGO_ENABLED=1 go run -ldflags="-linkmode internal" main.go` 

or you'll receive this scare message:

```
# command-line-arguments
dlopen: unhandled relocation for purego_dlopen (type 44 (SDYNIMPORT) rtype 9 (R_CALLARM64))
dlsym: unhandled relocation for purego_dlsym (type 44 (SDYNIMPORT) rtype 9 (R_CALLARM64))
dlerror: unhandled relocation for purego_dlerror (type 44 (SDYNIMPORT) rtype 9 (R_CALLARM64))
dlclose: unhandled relocation for purego_dlclose (type 44 (SDYNIMPORT) rtype 9 (R_CALLARM64))
```

Which fails in the linker [here](https://github.com/golang/go/blob/d2901205c15d5d839c95988497c190b652c5cf3a/src/cmd/link/internal/ld/data.go#L248). I'm trying to figure out if there is a way to not require this flag.

TotallyGamerJet:
Sadly, it does not work on [go.dev/playground](https://go.dev/play/p/7Illx4KeAB5). :(

`/tmpfs/play: error while loading shared libraries: libdl.so.2: cannot open shared object file: No such file or directory`

Although, it did try to load the shared library so perhaps it is possible if we can figure out the location of it.

TotallyGamerJet:
I deleted this branch accidentally. Closing to open a new one.

PR #40, opened 2022-09-09
rename cgo.go to cgo_darwin.go

TotallyGamerJet:


TotallyGamerJet:
I'm gonna use Github Actions to test this behavior

TotallyGamerJet:
Well that test did absolutely nothing :(

TotallyGamerJet:
I guess it won't let me test anymore. I'll get a VM to see if I can reproduce this error in the operating systems

TotallyGamerJet:
Cause I'm a little confused as to what is actually failing. I don't think it's the OS itself. Perhaps, cross-compilation?

TotallyGamerJet:
Ok now I'm thoroughly confused. It can't build an empty main on linux?

TotallyGamerJet:
Also, I am unable to reproduce this on a vm. Any luck for you?

TotallyGamerJet:
It's only failing on 1.15 Ubuntu and not the other versions. Why is this?

hajimehoshi:
> It's only failing on 1.15 Ubuntu and not the other versions. Why is this?

I'm not 100% confident but this might be a bug in Go 1.15, which has already been fixed. I think it is OK so far to forbid importing purego for Linux e.g. purego can be imported only from Windows and Darwin.

TotallyGamerJet:
That works. We need to not test 1.15 then on Linux

hajimehoshi:
Hm, the test failure happens cross-building without Cgo?

hajimehoshi:
I think the fix will be like this: https://github.com/ebitengine/purego/compare/main...test

TotallyGamerJet:
@hajimehoshi do you just want to submit that as a PR?

hajimehoshi:
Sure!

TotallyGamerJet:
Closed because solved in #41

Issue #37, opened 2022-09-07
support Arm64 iOS with Go 1.15


hajimehoshi:
Now Go 1.15 for Arm64 iOS doesn't work as `callbackWrapPicker` is missing.

We don't have to care about macOS since Go 1.15 doesn't support Arm macOS. This support started as of [Go1.16](https://go.dev/doc/go1.16#darwin).

This is nice to have and low priority.

hajimehoshi:
https://cs.opensource.google/go/x/mobile/+/master:go.mod

Go mobile itself doesn't Go 1.15, so the target would be a user who still uses a pretty old Go mobile...

hajimehoshi:
We don't have to do this per #43.

PR #41, opened 2022-09-10
purego: skip importing `runtime/cgo` for Windows

### LINKED ISSUE #39: `go build` a main package with purego with Cgo enabled might fail on Windows and Linux
https://github.com/ebitengine/purego/blob/test/example/main.go

The test bot builds the above main package like below:

```
go build -v ./...
```

And this failed at least on Linux like this:

https://github.com/ebitengine/purego/runs/8276795548

```
# github.com/ebitengine/purego/example
2022/09/09 18:57:24 duplicated definition of symbol _cgo_init
```

I think this is a minimal reproducible case of hajimehoshi/ebiten#2313: just building a main package with purego with Cgo enabled.

hajimehoshi:
This change resolves these issues:

* Buiding purego for `GOOS=darwin` and `GOARCH=arm64` with Go 1.15
  is not supported so far (#37). Skip this.
* Buiding purego for `GOOS=windows` with Cgo failed due to duplicated
  symbols (#39). Let's not import `runtime/cgo` for Windows. `fakecgo` is
  implemented only for Darwin so this should be fine. As a side effect, purego
  can no longer be imported from non-Darwin or non-Windows so far.

Updates #37
Closes #39


hajimehoshi:
Please take a look, thanks

.github/workflows/test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           go-version: ${{ matrix.go }}
 
       - name: go vet
-        if: ${{ !startsWith(matrix.os, 'windows-') && !startsWith(matrix.os, 'macos-') }}
+        if: ${{ !startsWith(matrix.os, 'macos-') && !startsWith(matrix.os, 'windows-') }}
```
hajimehoshi, line 32: This change is not related to the issue. I just wanted to reorder them in an alphabetical order. I can make it in a later PR.

hajimehoshi: Reverted this change.

.github/workflows/test.yml
```diff
@@ -48,15 +48,21 @@ jobs:
...
-          env GOOS=darwin GOARCH=arm64 go build -v ./...
 
       - name: go build (Windows, ARM64)
         if: ${{ !startsWith(matrix.go, '1.15.') && !startsWith(matrix.go, '1.16.') }}
         run: |
           # Check cross-compiling Windows binaries.
           env GOOS=windows GOARCH=arm64 go build -v ./...
 
+      - name: go build (macOS, ARM64)
+        if: ${{ !startsWith(matrix.go, '1.15.') }}
```
TotallyGamerJet, line 59: Should we add a comment describing why we don't build for 1.15?

.github/workflows/test.yml
```diff
@@ -48,15 +48,21 @@ jobs:
 
           # Check cross-compiling macOS binaries.
           env GOOS=darwin GOARCH=amd64 go build -v ./...
-          env GOOS=darwin GOARCH=arm64 go build -v ./...
 
       - name: go build (Windows, ARM64)
         if: ${{ !startsWith(matrix.go, '1.15.') && !startsWith(matrix.go, '1.16.') }}
```
TotallyGamerJet, line 53: Ditto for the comment?

TotallyGamerJet (commented):
Thoughts?

TotallyGamerJet:
Why the swap?

.github/workflows/test.yml
hajimehoshi: Done

.github/workflows/test.yml
hajimehoshi: Done

hajimehoshi:
> Why the swap?

I think it's reverted.

TotallyGamerJet (approved):
LGTM!

PR #46, opened 2022-09-14
objc: add protocol support

TotallyGamerJet:
This is needed for [ebitengien#2326](https://github.com/hajimehoshi/ebiten/pull/2326).

Should the `GetProtocol` function return a pointer to Protocol or just Protocol? The Objective-C [runtime](https://developer.apple.com/documentation/objectivec/1418870-objc_getprotocol?language=objc) returns a pointer however there is never a reason for us to need to dereference this pointer so I thought it best to just remove it.

### DIFF
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -232,6 +232,7 @@ type Protocol uintptr
 func GetProtocol(name string) *Protocol {
 	n := strings.CString(name)
 	p, _, _ := purego.SyscallN(objc_getProtocol, uintptr(unsafe.Pointer(n)))
+	runtime.KeepAlive(n)
 	return *(**Protocol)(unsafe.Pointer(&p))
 }
 
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -233,6 +233,7 @@ func GetProtocol(name string) *Protocol {
 	n := strings.CString(name)
 	p, _, _ := purego.SyscallN(objc_getProtocol, uintptr(unsafe.Pointer(n)))
 	runtime.KeepAlive(n)
+	// We take the address and then dereference it to trick go vet from creating a possible miss-use of unsafe.Pointer
 	return *(**Protocol)(unsafe.Pointer(&p))
 }
 
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -25,6 +25,7 @@ var (
 	objc_msgSend              = purego.Dlsym(objc, "objc_msgSend")
 	objc_msgSendSuper2        = purego.Dlsym(objc, "objc_msgSendSuper2")
 	objc_getClass             = purego.Dlsym(objc, "objc_getClass")
+	objc_getProtocol          = purego.Dlsym(objc, "objc_getProtocol")
 	objc_allocateClassPair    = purego.Dlsym(objc, "objc_allocateClassPair")
 	objc_registerClassPair    = purego.Dlsym(objc, "objc_registerClassPair")
 	sel_registerName          = purego.Dlsym(objc, "sel_registerName")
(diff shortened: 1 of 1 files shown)


objc/objc_runtime_darwin.go
```diff
@@ -189,6 +191,14 @@ func (c Class) AddIvar(name string, ty interface{}, types string) bool {
 	return byte(ret) != 0
 }
 
+// AddProtocol adds a protocol to a class.
+// Returns true if the protocol was added successfully, otherwise NO (for example,
```
hajimehoshi, line 195: otherwise false?

hajimehoshi (approved):
LGTM!

TotallyGamerJet:
Did you want to discuss my question first or do you agree with my conclusion?

hajimehoshi:
Sorry I missed that.

> Should the GetProtocol function return a pointer to Protocol or just Protocol? The Objective-C [runtime](https://developer.apple.com/documentation/objectivec/1418870-objc_getprotocol?language=objc) returns a pointer however there is never a reason for us to need to dereference this pointer so I thought it best to just remove it.

Hm, then

```go
	p, _, _ := purego.SyscallN(objc_getProtocol, uintptr(unsafe.Pointer(n)))
	return Protocol(p)
```

Doesn't this SyscallN return a pointer to Protocol instead of a Protocol itself?

objc/objc_runtime_darwin.go
```diff
@@ -215,6 +225,16 @@ func (i Ivar) Offset() uintptr {
...
 }
 
+// Protocol is a type that declares methods that can be implemented by any class.
+type Protocol uintptr
+
+// GetProtocol returns the protocol for the given name or nil if there is no protocol by that name.
+func GetProtocol(name string) Protocol {
+	n := strings.CString(name)
+	p, _, _ := purego.SyscallN(objc_getProtocol, uintptr(unsafe.Pointer(n)))
+	return Protocol(p)
```
hajimehoshi, line 235: `runtime.KeepAlive(n)` after `SyscallN`?

hajimehoshi (changes requested):
Let's resolve the signature issue

objc/objc_runtime_darwin.go
```diff
@@ -215,6 +225,17 @@ func (i Ivar) Offset() uintptr {
 	return ret
 }
 
+// Protocol is a type that declares methods that can be implemented by any class.
+type Protocol uintptr
+
+// GetProtocol returns the protocol for the given name or nil if there is no protocol by that name.
+func GetProtocol(name string) *Protocol {
```
hajimehoshi, line 232: Let's comment that this is nil when the protocol is not found.

objc/objc_runtime_darwin.go
```diff
@@ -215,6 +225,17 @@ func (i Ivar) Offset() uintptr {
...
+	n := strings.CString(name)
+	p, _, _ := purego.SyscallN(objc_getProtocol, uintptr(unsafe.Pointer(n)))
+	runtime.KeepAlive(n)
+	return *(**Protocol)(unsafe.Pointer(&p))
```
hajimehoshi, line 236: Isn't this:

```go
return (*Protocol)(unsafe.Pointer(p))
```

?

TotallyGamerJet: go vet will complain that it is potentially unsafe conversion which is why I did the indirection

objc/objc_runtime_darwin.go
TotallyGamerJet: It is, isn't it?

> GetProtocol returns the protocol for the given name or **nil if there is no protocol** by that name.

objc/objc_runtime_darwin.go
hajimehoshi: I see. Syscall doesn't work well `go vet` actually (https://github.com/golang/go/issues/41205)

It is fine to not doing go-vet, but what do you think?

If we keep go-vet and use the indirection way, let's leave a comment for the intention.

objc/objc_runtime_darwin.go
hajimehoshi: Oh sorry I missed that...

hajimehoshi (approved):
LGTM!

PR #47, opened 2022-09-28
Simplify assembly

TotallyGamerJet:
Using go:linkname and a call5 trampoline function it is possible to remove assembly and push it into Go code. Arguably, it is pretty ugly Go code but functional. Having less assembly means less chance to make a copy paste error and make it easier to port to other architectures

### DIFF
--- a/internal/fakecgo/trampolines_stubs.s
+++ b/internal/fakecgo/trampolines_stubs.s
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build darwin
+// +build darwin
+
+#include "textflag.h"
+
+// these stubs are here because it is not possible to go:linkname directly the C functions on darwin arm64
+
+TEXT _malloc(SB), NOSPLIT, $0-0
+	JMP libc_malloc(SB)
+	RET
+
+TEXT _free(SB), NOSPLIT, $0-0
+	JMP libc_free(SB)
+	RET
+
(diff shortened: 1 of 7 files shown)


internal/fakecgo/trampolines_stubs.s
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build darwin
+// +build darwin
+
+#include "textflag.h"
+
+// these stubs are here because it is not possible to go:linkname directly the C functions on darwin arm64
```
hajimehoshi, line 9: Is this file needed only for darwin/arm64?

TotallyGamerJet: No they are needed on both amd64 and arm64. The reason it has to be done this way is bc arm64 doesn't allow directly linking to dynamic C functions

internal/fakecgo/trampolines_stubs.s
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build darwin
+// +build darwin
```
hajimehoshi, line 5: Let's rename the file to `trampolines_stubs_darwin.s` or `trampolines_darwin.s` instead of build tags here

hajimehoshi (approved):
LGTM

hajimehoshi:
Oh sorry I should have let you merge this as usual...

TotallyGamerJet:
All good! It happens

PR #48, opened 2022-10-03
purego: set R2 to be the float return value

TotallyGamerJet:
SyscallN returns 2 uintptr. Currently, the 2nd return value gets what is in the second integer register which isn't very helpful. This PR puts the first float return value into R2. This is used while porting GLFW to Go

### DIFF
--- a/sys_darwin_amd64.s
+++ b/sys_darwin_amd64.s
@@ -64,7 +64,7 @@ TEXT syscall9X(SB), NOSPLIT, $0
 
 	MOVQ 24(BP), DI              // get the pointer back
 	MOVQ AX, syscall9Args_r1(DI) // r1
-	MOVQ DX, syscall9Args_r2(DI) // r2
+	MOVQ X0, syscall9Args_r2(DI) // r2
 
 	XORL AX, AX  // no error (it's ignored anyway)
 	ADDQ $32, SP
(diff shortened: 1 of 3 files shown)


hajimehoshi (commented):
IIUC the second returning value is used to return 64bit values on 32 bit machines. Now the current fix is applied to 64bit macOS, but would it be consistent?

TotallyGamerJet:
There is no 32bit macOS so it's not an issue. As for eventually supporting 32bit linux I'm not sure what is best. I still do plan on writing the replacement for SyscallN that will support any return or argument arrangement. However, the replacement will be significantly slower so I was planning on keeping SyscallN as a faster calling mechanism.

hajimehoshi:
So the SyscallN behaviors would differ among OSes (Windows, macOS, Linux). Right?

TotallyGamerJet:
Well that was the plan. However, the more I think about it the less I like that idea. I don't like that SyscallN behaves differently on different platforms. Perhaps I write the replacement and just use SyscallN for returning 64bit numbers like on Windows. What do you think?

hajimehoshi:
Yeah, I also prefer consistent behavior. If we need to change the behavior from SyscallN, we should rename it or add a new function with a new name.

TotallyGamerJet:
Ok. I will close this and work on the general calling mechanism when I have time.

PR #49, opened 2022-10-18
purego: new mechanism for calling C functions

TotallyGamerJet:
This PR adds a new function `purego.Func` which takes a shared object handle, the name of a symbol in the object and a pointer to a function matching the calling convention. This is the start of a better way of calling C functions. First, it is type safe from the callers perspective. Secondly, it removes repeated code such as `strings.CString` and `runtime.KeepAlive` into one place (`purego.Func`). I ported the objective-c runtime to use this new mechanism to test it out and see how it looks in code. I think it's a big improvement.

The downside is that this is a slower method than calling `purego.SyscallN` directly but will eventually allow for more complex arguments and returns such as floats, complex, and structs.

We need to decide on if `purego.Func` should return an error?

I'm kind of leaning towards no since the only error you could *potentially* recover from is one where the symbol isn't found in the shared object but then what would you do? Stub the function? The other errors (fptr not being a function pointer, too many return arguments) just aren't recoverable in any meaningful way which is why I think it should panic.

Should the objc package use this method or call `purego.SyscallN` directly?

I think the later since this package gets called a LOT in code ported from Objective-C and we want it to be as fast as possible. The overhead of determining the types seems like a waste for these core functions.

TODO: 
[X] Document the conversion from Go types to C types as part of `purego.Func`.
[X] Keep pointer types alive
[ ] Implement those other types.
[ ] Port oto/ebitengine to use this new method?


func_darwin.go
```diff
@@ -0,0 +1,72 @@
+package purego
```
hajimehoshi, line 1: Add a license comment.

func_darwin.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"errors"
+	"github.com/ebitengine/purego/internal/strings"
+	"reflect"
+	"unsafe"
+)
```
hajimehoshi, line 8:
```
import (
	"errors"
	"reflect"
	"unsafe"

	"github.com/ebitengine/purego/internal/strings"
)
```

objc/objc_runtime_darwin.go
```diff
@@ -24,27 +24,41 @@ var (
 
 	objc_msgSend              = purego.Dlsym(objc, "objc_msgSend")
 	objc_msgSendSuper2        = purego.Dlsym(objc, "objc_msgSendSuper2")
```
hajimehoshi, line 26: Don't we replace them with funcs?

func_darwin.go
```diff
@@ -0,0 +1,72 @@
...
+			case reflect.Uintptr, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+				sysargs[i] = uintptr(v.Uint())
+			case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+				sysargs[i] = uintptr(v.Int())
+			case reflect.Ptr, reflect.UnsafePointer, reflect.Slice:
+				sysargs[i] = v.Pointer() // TODO: keep alive
+			case reflect.Func:
+				sysargs[i] = NewCallback(v.Interface())
+			default:
+				panic("purego: unsupported kind: " + v.Kind().String())
```
hajimehoshi, line 44: return error?

func_darwin.go
```diff
@@ -0,0 +1,72 @@
...
+			v.SetInt(int64(r1))
+		case reflect.Bool:
+			v.SetBool(r1 != 0)
+		case reflect.UnsafePointer:
+			// We take the address and then dereference it to trick go vet from creating a possible miss-use of unsafe.Pointer
+			v.SetPointer(*(*unsafe.Pointer)(unsafe.Pointer(&r1)))
+		case reflect.Ptr:
+			v = reflect.NewAt(outType, unsafe.Pointer(&r1)).Elem()
+		default:
+			panic("purego: unsupported return kind: " + outType.Kind().String())
```
hajimehoshi, line 66: ditto

objc/objc_runtime_darwin.go
```diff
@@ -24,27 +24,41 @@ var (
...
+	purego.Func(objc, "objc_getProtocol", &objc_getProtocol)
+	purego.Func(objc, "objc_allocateClassPair", &objc_allocateClassPair)
+	purego.Func(objc, "objc_registerClassPair", &objc_registerClassPair)
+	purego.Func(objc, "sel_registerName", &sel_registerName)
+	purego.Func(objc, "class_getSuperclass", &class_getSuperclass)
+	purego.Func(objc, "class_getInstanceVariable", &class_getInstanceVariable)
+	purego.Func(objc, "class_addMethod", &class_addMethod)
+	purego.Func(objc, "class_addIvar", &class_addIvar)
+	purego.Func(objc, "class_addProtocol", &class_addProtocol)
+	purego.Func(objc, "ivar_getOffset", &ivar_getOffset)
```
hajimehoshi, line 53: We should handle errors from `purego.Func`

hajimehoshi (commented):
Seems good.

objc/objc_runtime_darwin.go
TotallyGamerJet: Did you see my question about if `purego.Func` should return an error in the first place?

objc/objc_runtime_darwin.go
TotallyGamerJet: Their signatures are not known. They are determined by whichever ObjC method is being called. I didn't want to make any breaking changes to the exported api so I left them alone. How would we go about changing `ID.Send` and `ID.SendSuper` to utilize `purego.Func`?

func_darwin.go
TotallyGamerJet: Inside here cannot return errors since these are calling C functions and the type is set by the function pointer

func_darwin.go
TotallyGamerJet: See other comment about the panic

objc/objc_runtime_darwin.go
hajimehoshi: Sure. So can't we treat a variable number of function (`func(args ...interface{})`) with Func?

TotallyGamerJet: Potentially, it would mean that `purego.Func` would need a special case for when the last argument is a slice of interfaces. I can draft something up tomorrow

objc/objc_runtime_darwin.go
hajimehoshi: Sorry I read it now. So you don't want not to return an error from Func, right? I think it's OK to panic instead of returning error. This is not Go-conventional in general, but purego is a low-layer API and panicking should be allowed like what `reflect` package does.

hajimehoshi (commented):
Pretty nice. Are you working on the TODOs now?

TotallyGamerJet:
Yes, I am!

func_darwin.go
```diff
@@ -0,0 +1,159 @@
...
+//
+// These conversions describe how a Go type in the fptr will be used to call
+// the C function. It is important to note that there is no way to verify that fptr
+// matches the C function. This also holds true for struct types where the padding
+// needs to be ensured to match that of C; FuncHandle does not verify this.
+//
+// Conversion Type (Go => C)
+//
+//	string => char*
+//
```
hajimehoshi, line 30: I think we don't need an empty line for each item.

TotallyGamerJet:
I've ported Oto to use `purego.FuncHandle`. Do you want to wait until floats and struct support is added before merging this? I'm not sure when I'll be able to do that since I'm still working on writing the generator code for the glfw.dylib. This is fully functional as it for all the types currently supported by `SyscallN`

dlfcn.go
```diff
@@ -52,26 +51,7 @@ func Dlerror() string {
...
-	// use unsafe.Slice once we reach 1.17
-	s := make([]byte, length)
-	var src []byte
-	h := (*reflect.SliceHeader)(unsafe.Pointer(&src))
-	h.Data = msg
-	h.Len = length
-	h.Cap = length
-	copy(s, src)
-	return string(s)
+	return strings.GoString(msg)
```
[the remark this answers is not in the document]
TotallyGamerJet: I did this change so that `FuncHandle` could return string types

hajimehoshi:
> Do you want to wait until floats and struct support is added before merging this?
> [...]
> This is fully functional as it for all the types currently supported by SyscallN

I'm confused Do you mean the current `FuncHandle` is strong enough to replace SyscallN even without floats and structs?

TotallyGamerJet:
Yes it can replace all occurrences of SyscallN in Oto and Ebitengine in its current state

func_darwin.go
```diff
@@ -0,0 +1,142 @@
...
+	sym := Dlsym(handle, name)
+	if sym == 0 {
+		panic("purego: couldn't find symbol" + Dlerror())
+	}
+	_Func(sym, fptr)
+}
+
+// _Func takes a C function ptr and a pointer to a Go function which
+// will be set to a function calling the C function with those arguments.
+func _Func(cfn uintptr, fptr interface{}) {
```
hajimehoshi, line 60: `setFunc` or `registerFunc`?

func_darwin.go
```diff
@@ -0,0 +1,142 @@
...
+//	int64 => int64_t
+//	float32 => float
+//	float64 => double
+//	struct => struct
+//	func => C function
+//	[]T, unsafe.Pointer, *T => void*
+//
+// There is a special case when the last argument of fptr is a variadic interface
+// it will be expanded into a call to the C function as if it had those arguments.
+func FuncHandle(handle uintptr, name string, fptr interface{}) {
```
hajimehoshi, line 50: I feel like this function name should start with a verb like `Register`, but what do you think?

TotallyGamerJet: `RegisterFuncHandle` seems good. Kind of long though

TotallyGamerJet: Hmmm Should it actually be `RegisterHandleFunc` since in English the adjective usually goes on the right of the noun it describes. In this case Handle describes the type of function it is. Does that make sense?

hajimehoshi: Or just `RegisterFunc`?

TotallyGamerJet: That could work. However, `RegisterFuncHandle` specifically takes a handle to a shared object while the unexported function `registerFunc`  takes the C function. There may be a time that exporting that function might make sense. It could be useful while porting C code. Something like this:

```
// int foo(char *c) { /* ... */ }
import "C"
var fooFn func(string) int
purego.RegisterFunc(C.foo, &fooFn)
```
This way you don't have to port the entire C function before having something running.

If you don't think that's a convincing argument than let's just go with `RegisterFunc`. Thoughts?

hajimehoshi: I see. What about

* `FuncHandle` → `RegisterFuncFromLib`
* `_Func` → `RegisterFunc`

? The former might be too verbose...

```go
func init() {
    purego.RegisterFuncFromLib(objc, "objc_msgSend", &objc_msgSend)
}
```

Even without the former, we can write like this:

```go
func init() {
    purego.RegisterFunc(Dlsym(objc, "objc_msgSend"), &objc_msgSend)
}
```

I slightly prefer only having RegisterFunc.

TotallyGamerJet: That is correct. The function is the destination and handle and name are source arguments

hajimehoshi: OK so I'd be happy if you could swap the arguments.

TotallyGamerJet: RegisterHandleFunc is just a wrapper around registerFunc which can be added later if it is useful. I'm kind of thinking tho that most people will do what you wrote and not check that the Dlopen succeeded or not. RegisterHandleFunc will at least force it

hajimehoshi: I see. Needing to check errors makes sense. Another suggestion is `RegisterLibFunc`:

```go
func init() {
    purego.RegisterLibFunc(&objc_msgSend, objc, "objc_msgSend")
}
```

TotallyGamerJet: I like it! If that way if we decide to export `RegisterFunc` we can.

func_darwin.go
```diff
@@ -0,0 +1,142 @@
...
+	sym := Dlsym(handle, name)
+	if sym == 0 {
+		panic("purego: couldn't find symbol: " + Dlerror())
+	}
+	registerFunc(sym, fptr)
+}
+
+// registerFunc takes a C function ptr and a pointer to a Go function which
+// will be set to a function calling the C function with those arguments.
+func registerFunc(cfn uintptr, fptr interface{}) {
```
hajimehoshi, line 60: Let's swap the arguments here too

func_darwin.go
```diff
@@ -0,0 +1,142 @@
...
+// the C function. It is important to note that there is no way to verify that fptr
+// matches the C function. This also holds true for struct types where the padding
+// needs to be ensured to match that of C; RegisterLibFunc does not verify this.
+//
+// Type Conversions (Go => C)
+//
+//	string => char*
+//	bool => _Bool
+//	uintptr => uintptr_t
+//	uint => System Dependent
```
hajimehoshi, line 32: `uint32_t` or `uint64_t`? Or just `uint`?

hajimehoshi: As `unsigned int` is 32bit in most environments, doesn't this lose information...?

func_darwin.go
hajimehoshi: I'm fine to expose RegisterFunc later, but we don't have to do it now :-)

func_darwin.go
TotallyGamerJet: True. I'll go with what you first said

hajimehoshi: Thanks!

func_darwin.go
```diff
@@ -0,0 +1,142 @@
...
+//	int16 => int16_t
+//	int32 => int32_t
+//	int64 => int64_t
+//	float32 => float
+//	float64 => double
+//	struct => struct
+//	func => C function
+//	[]T, unsafe.Pointer, *T => void*
+//
+// There is a special case when the last argument of fptr is a variadic interface (or []interface}
```
hajimehoshi, line 48: Just out of curiosity, would it be hard to support other types than `interface{}`?

func_darwin.go
```diff
@@ -0,0 +1,142 @@
...
+//	uint32 => uint32_t
+//	uint64 => uint64_t
+//	int => int32_t or int64_t
+//	int8 => int8_t
+//	int16 => int16_t
+//	int32 => int32_t
+//	int64 => int64_t
+//	float32 => float
+//	float64 => double
+//	struct => struct
```
hajimehoshi, line 44: Let's note that floats and structs are still WIP here

hajimehoshi (approved):
LGTM, thanks!

func_darwin.go
TotallyGamerJet: Do you have an example?

Also note that this does not support calling variadic functions like `int printf(const char* format, ...);` It's basically a cast to the desired C function. `(*(char*, int)())(printf)("%d", 3)` is not the same as `printf("%d, 3)` since the second uses va_args under the hood. The first is what is currently handled

func_darwin.go
hajimehoshi: Confirmed RegisterFunc was exported. That's also fine :-)

func_darwin.go
hajimehoshi: For example: https://pkg.go.dev/reflect#Append (I don't think this can be a C function but this is an example with a non-interface{} type)

hajimehoshi: I now understand that the purpose of `RegisterFunc` is to have a Go function that is a counterpart of a C function. So, having another type than `interface{}` would not make sense.

By the way, in the comment, the order of types is `(Go type) => (C type)`, but is there an order here?

hajimehoshi:
I'm gonna sleep so I might reply tomorrow. Thank you for your hard work!

func_darwin.go
TotallyGamerJet: Yes the Go function should have the same type as the C function being called.

There is no order except for how I thought of the type

TotallyGamerJet: As for the Append example that would be a C function like `Append(Value s, Value *v)`. `...` in Go is just a slice and slices of any type are supported. Perhaps this example will clear things up about what `...interface{}` means in the context of `RegisterFunc`. This is only really useful for objc_msgSend variants since it is completely different from C's variadic which use a different calling convention then normally passing those arguments.

```go
var foo func(string, ...any)
purego.RegisterLibFunc(&foo, handle, "foo")
foo("hello", 123) // calls foo(string, int)
foo("hello", 1.342) // calls foo(string, double)
```
Neither of these calls are `foo(string, ...)`.

hajimehoshi: I understood.

> Yes the Go function should have the same signature as the C function being called.

Maybe `<=>` would be better than `=>`?

TotallyGamerJet: I like that since some types like slice cannot be returned from C to Go since there is no way to get its length

hajimehoshi: OK so I'm fine with `=>`. Still LGTM

PR #50, opened 2022-10-21
rename libc_* to purego_*

TotallyGamerJet:
While porting purego to Linux I discovered that I had missed the cgo_import_dynamic comment for pthread_sigmask. The reason purego was working was because the symbol is normally linked to in the Go runtime. I decided to make a unique name using `purego_*` for the imported symbols. This way we aren't relying on runtime behavior that can easily change.


hajimehoshi (approved):
LGTM, thanks!

PR #51, opened 2022-10-23
use RegisterFunc for dlfcn.go

TotallyGamerJet:
limit usage of SyscallN in favor of RegisterFunc and RegisterLibFunc

### DIFF
--- a/dlfcn.go
+++ b/dlfcn.go
@@ -7,15 +7,24 @@
 package purego
 
 import (
-	"runtime"
 	"unsafe"
-
-	"github.com/ebitengine/purego/internal/strings"
 )
 
 const RTLD_GLOBAL = 0x8
 
 const RTLD_DEFAULT = ^uintptr(1)
+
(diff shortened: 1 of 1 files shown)


dlfcn.go
```diff
@@ -7,16 +7,25 @@
...
 )
 
 const RTLD_GLOBAL = 0x8
 
 const RTLD_DEFAULT = ^uintptr(1)
 
+var fnDlopen func(path string, mode int) uintptr
+var fnDlsym func(handle uintptr, name string) uintptr
+var fnDlerror func() string
+var fnDlclose func(handle uintptr) bool
```
hajimehoshi, line 20: I slightly prefer this style:

```go
var (
    fnDlopen func(path string, mode int) uintptr
    fnDlsym func(handle uintptr, name string) uintptr
    fnDlerror func() string
    fnDlclose func(handle uintptr) bool
)
```

hajimehoshi (approved):
LGTM, thanks!

PR #53, opened 2022-10-23
Move constants under platform specific file

### LINKED ISSUE #52: RTLD_DEFAULT has wrong value
According to [opensource.apple.com](https://opensource.apple.com/source/dyld/dyld-360.14/include/dlfcn.h.auto.html) `RTLD_DEFAULT` is defined as `#define	RTLD_DEFAULT	((void *) -2)`. In purego I wrote `const RTLD_DEFAULT = ^uintptr(1)` which is not the same binary representation. 

This is a bug in version 0.1.0 and should be corrected to something like `const RTLD_DEFAULT   = uintptr(0x8000000000000002)` .

Secondly, I think we should remove this constant from purego 0.2.0 since it is not used in purego, Oto or Ebitengine and is darwin specific. See the [Unix specification](https://pubs.opengroup.org/onlinepubs/7908799/xsh/dlfcn.h.html).

TotallyGamerJet:
This PR does three things.

1. moves the constants under dlfcn_darwin.go since the value is platform specific
2. remove RTLD_DEFAULT which is specific to darwin.
3. adds RTLD_LAZY, RTLD_NOW, RTLD_LOCAL to complete the unix specification

Closes #52

### DIFF
--- a/dlfcn.go
+++ b/dlfcn.go
@@ -10,10 +10,6 @@ import (
 	"unsafe"
 )
 
-const RTLD_GLOBAL = 0x8
-
-const RTLD_DEFAULT = ^uintptr(1)
-
 var (
 	fnDlopen  func(path string, mode int) uintptr
 	fnDlsym   func(handle uintptr, name string) uintptr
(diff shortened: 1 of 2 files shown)


hajimehoshi (approved):
LGTM

Issue #54, opened 2022-10-29
Give up supporting Go 1.16 and Go 1.17


hajimehoshi:
* gccgo12 already supports Go 1.18 (https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libgo/VERSION;h=39560f0fcdd63e4bb8eead6a3f849e91636d31eb;hb=2ee5e4300186a92ad73f1a1a64cb918dc76c8d67)
* ~Debian~Ubuntu stable already supports Go 1.18 (note that ~Ubuntu~Debian stable supports Go 1.15 by default...)
* The Go team no longer supports Go 1.17 and older, as we know

@TotallyGamerJet what do you think? Are there extra tasks for this?

TotallyGamerJet:
There isn't anything that needs to be done since purego already works for 1.18. Updating to 1.18 would allow us to remove the stack and register hack files though. This is because on Darwin amd64 and arm64 would use register based calling convention by 1.18

hajimehoshi:
Another reason is that semi-standard libraries like `golang.org/x/sys/unix` uses `unsafe.Slice`, which Go 1.16 doesn't support.

> There isn't anything that needs to be done since purego already works for 1.18. Updating to 1.18 would allow us to remove the stack and register hack files though. This is because on Darwin amd64 and arm64 would use register based calling convention by 1.18

Sure! As there seems no issue, I'll drop Go 1.16 and 1.17 supports Ebitengine and Oto first. After that, we can remove the hacks for old Go. It's not hurry anyway.

TotallyGamerJet:
Oh yeah! `unsafe.Slice` will cleanup some of the code

hajimehoshi:
Ebitengine: https://github.com/hajimehoshi/ebiten/commit/606ff77d89eadf53a045ea94553875da2ce0b2ba
Oto: https://github.com/hajimehoshi/oto/commit/ad5f0332d3b1850735fb7639dd6dfb100a0d4c86

OK, it's ready! As I said, it's not hurry.

hajimehoshi:
By the way, do you think it is possible to remove go-vet warnings?

```
$ go vet ./...
# github.com/ebitengine/purego/objc_test
objc/objc_runtime_darwin_test.go:31:41: possible misuse of unsafe.Pointer
objc/objc_runtime_darwin_test.go:34:34: possible misuse of unsafe.Pointer
```

I feel like these are false positives. See also https://github.com/golang/go/issues/56487

[...]

PR #57, opened 2022-10-31
objc: remove unsafe usage from test

TotallyGamerJet:
This PR is based on my comment in #54. It removes all uses of package unsafe from the tests by creating a struct that matches the structure of the Objective-C object.


hajimehoshi (approved):
Nice! LGTM

PR #58, opened 2022-11-01
purego: enumerate types that can be returned from callbacks

TotallyGamerJet:
This PR explicitly enumerates which types can be returned from a callback. It also adds support for returning `reflect.Pointer` and `reflect.UnsafePointer`


hajimehoshi (approved):
LGTM

PR #59, opened 2022-11-03
objc: add RegisterClass API

TotallyGamerJet:
This PR adds `RegisterClass` function which takes a `Selector` interface that it uses to add methods based on the argument sent to `RegisterClass`. It returns an Objective-C class that matches the pointer to struct or an error if something failed. I also included an example under `example/objc/main_darwin.go`.


example/objc/main_darwin.go
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build example
```
hajimehoshi, line 4: I think we no longer need `example` tag as we dropped the support of Go 1.16 and 1.17: https://github.com/hajimehoshi/ebiten/issues/1462#issuecomment-1229493555

example/objc/main_darwin.go
```diff
@@ -0,0 +1,56 @@
...
+
+package main
+
+import (
+	"fmt"
+	"unsafe"
+
+	"github.com/ebitengine/purego/objc"
```
hajimehoshi, line 12: As this is imported here, do we need `example/main.go`?

objc/objc_runtime_darwin.go
```diff
@@ -189,7 +506,7 @@ func NewIMP(fn interface{}) IMP {
 			ty.In(0).Elem().Field(0).Type != reflect.TypeOf(Class(0))):
 		fallthrough
 	case ty.In(1).Kind() != reflect.Uintptr:
-		panic("objc: NewIMP must take a (id, SEL) as its first two arguments")
+		panic("objc: NewIMP must take a (id, SEL) as its first two arguments got " + ty.String())
```
hajimehoshi, line 509: `argumenta; got`?

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
 }
 
 // AllocateClassPair creates a new class and metaclass. Then returns the new class, or Nil if the class could not be created
+// Deprecated: use RegisterClass instead
```
hajimehoshi, line 106:
```go
//
// Deprecated: use RegisterClass instead
```

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+// then the returned Class has been registered successfully.
+//
+// The struct's first field must be of type Class and have a tag that matches the format
+// `objc:"ClassName : SuperClassName <Protocol, ...>`. This tag is equal to how the class
+// would be defined in Objective-C.
+func RegisterClass(object Selector) (Class, error) {
+	ptr := reflect.TypeOf(object)
+	strct := ptr.Elem()
+	if strct.NumField() == 0 || strct.Field(0).Type != reflect.TypeOf(Class(0)) {
+		return 0, fmt.Errorf("%w: need objc.Class as first field", MismatchError)
```
hajimehoshi, line 139:
```go
fmt.Errorf("objc: need objc.Class as first field: %w", MismatchError)
```

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+	}
+	isa := strct.Field(0)
+	tag := isa.Tag.Get("objc")
+	if tag == "" {
+		return 0, fmt.Errorf("missing objc tag: %w", TagFormatError)
```
hajimehoshi, line 144:
```go
fmt.Errorf("objc: missing objc tag: %w", TagFormatError)
```

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+			}
+		}
+		split[0] = tag[0:i] // store ClassName
+		tag = tag[i:]
+
+		skipSpace()
+
+		// check for ':'
+		if len(tag) > 0 && tag[0] != ':' {
+			return 0, fmt.Errorf("missing ':': %w", TagFormatError)
```
hajimehoshi, line 179: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+		}
+		if len(tag) < i {
+			return 0, fmt.Errorf("missing SuperClassName: %w", TagFormatError)
+		}
+		split[1] = tag[:i] // store SuperClassName
+		tag = tag[i:]      // drop SuperClassName
+		skipSpace()
+		if len(tag) > 0 {
+			if tag[0] != '<' {
+				return 0, fmt.Errorf("expected '<': %w", TagFormatError)
```
hajimehoshi, line 204: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+					case '>':
+						// If there is actually an identifier - add it.
+						if i > 0 {
+							split = append(split, tag[:i])
+							tag = tag[i:]
+						}
+						break outer
+					}
+				}
+				return 0, fmt.Errorf("expected '>': %w", TagFormatError)
```
hajimehoshi, line 236: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+	class := objc_allocateClassPair(GetClass(split[1]), split[0], 0)
+	if class == 0 {
+		return 0, fmt.Errorf("failed to create class with name '%s'", split[0])
+	}
+	if len(split) > 2 {
+		// Add Protocols
+		for _, n := range split[2:] {
+			succeed := class.AddProtocol(GetProtocol(n))
+			if !succeed {
+				return 0, fmt.Errorf("couldn't add Protocol %s", n)
```
hajimehoshi, line 249: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+		}
+		sel := object.Selector(met.Name)
+		if sel == 0 {
+			continue
+		}
+		fn := met.Func.Interface()
+		imp, err := func() (imp IMP, err error) {
+			defer func() {
+				if r := recover(); r != nil {
+					err = fmt.Errorf("failed to create IMP: %s", r)
```
hajimehoshi, line 269: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+				}
+			}()
+			return NewIMP(fn), nil
+		}()
+		if err != nil {
+			return 0, fmt.Errorf("couldn't add Method %s: %w", met.Name, err)
```
hajimehoshi, line 275: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+		}
+		succeed := class.AddMethod(sel, imp, encodeFunc(fn))
+		if !succeed {
+			return 0, fmt.Errorf("couldn't add Method %s", met.Name)
```
hajimehoshi, line 279: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+	}
+	// Add Ivars
+	// Start at 1 because we skip the class object which is first
+	for i := 1; i < strct.NumField(); i++ {
+		f := strct.Field(i)
+		size := f.Type.Size()
+		alignment := uint8(math.Log2(float64(f.Type.Align())))
+		succeed := class_addIvar(class, f.Name, size, alignment, encodeType(f.Type))
+		if !succeed {
+			return 0, fmt.Errorf("couldn't add Ivar %s", f.Name)
```
hajimehoshi, line 290: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+		}
+		if offset := class.InstanceVariable(f.Name).Offset(); offset != f.Offset {
+			return 0, fmt.Errorf("couldn't add Ivar %s", f.Name)
```
hajimehoshi, line 293: ditto

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+		}
+	}
+	objc_registerClassPair(class)
+	if size1, size2 := class.InstanceSize(), strct.Size(); size1 != size2 {
+		return 0, fmt.Errorf("%w: sizes don't match %d != %d", MismatchError, size1, size2)
```
hajimehoshi, line 298:
```go
fmt.Errorf("objc: sizes don't match %d != %d: %w", size1, size2, MismatchError)
```

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,315 @@ func GetClass(name string) Class {
...
+
+// RegisterClass takes a pointer to a struct that implements the Selector interface.
+// It will register the structs fields and pointer receiver methods in the Objective-C
+// runtime using the SEL returned from Selector. Any errors that occur trying to add
+// a Method or Ivar is returned as an error. Such errors may occur in parsing or because
+// the size of the struct does not match the size in Objective-C. If no errors occur
+// then the returned Class has been registered successfully.
+//
+// The struct's first field must be of type Class and have a tag that matches the format
+// `objc:"ClassName : SuperClassName <Protocol, ...>`. This tag is equal to how the class
```
hajimehoshi, line 133: Is a space allowed between `ClassName` and `SuperClassName`? (I think that's fine)

example/objc/main_darwin.go
```diff
@@ -0,0 +1,56 @@
...
+	class, err := objc.RegisterClass(&barObject{})
+	if err != nil {
+		panic(err)
+	}
+
+	var object = objc.ID(class).Send(objc.RegisterName("new"))
+	object.Send(objc.RegisterName("setBar:"), 123)
+	var bar = int(object.Send(objc.RegisterName("bar")))
+	fmt.Println(bar)
+	// Output: 123
```
hajimehoshi, line 55: If output matters, we can have `Example_*` in a test.

example/objc/main_darwin.go
TotallyGamerJet: Wasn't that to test that purego compiled on Darwin and Windows? This only works on Darwin bc of ObjC

objc/objc_runtime_darwin.go
TotallyGamerJet: Yes whitespace is allowed in between. That is how it is for Objective-C class definitions

example/objc/main_darwin.go
TotallyGamerJet: The reason I didn't put it as an example was because you won't see the type and methods in Go Doc only the code inside the Example func. Unless I'm wrong about that

example/objc/main_darwin.go
hajimehoshi: I'm suggesting to remove `example/main.go`

hajimehoshi: Or, did you mean `example/main.go` is still needed for Windows?

example/objc/main_darwin.go
hajimehoshi: Acknowledged.

example/objc/main_darwin.go
TotallyGamerJet: Yeah that's what I was saying. I believe it was added for Windows since purego didn't compile correctly on Windows and so the example was created to ensure that it was.

example/objc/main_darwin.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+	"unsafe"
+
+	"github.com/ebitengine/purego/objc"
+)
+
+type barObject struct {
+	isa objc.Class `objc:"BarObject:NSObject <NSDelegateWindow>"`
```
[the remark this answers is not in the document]
hajimehoshi: Or

```go
ClassName() (className string, superClassName string, protocols []string)
```

example/objc/main_darwin.go
hajimehoshi: Sure!

objc/objc_runtime_darwin.go
```diff
@@ -98,10 +103,316 @@ func GetClass(name string) Class {
...
+// encodeType returns a string representing a type as if it was given to @encode(typ)
+// Source: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html#//apple_ref/doc/uid/TP40008048-CH100
+func encodeType(typ reflect.Type) string {
+	if typ == reflect.TypeOf(Class(0)) {
+		return encClass
+	} else if typ == reflect.TypeOf(ID(0)) {
+		return encId
+	} else if typ == reflect.TypeOf(SEL(0)) {
+		return encSelector
+	}
```
hajimehoshi, line 335: Let's use `switch`

example/objc/main_darwin.go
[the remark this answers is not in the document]
TotallyGamerJet: You'd still need `Selector(string) SEL` as well so that's 4 methods that need to be implemented. Two of which are just returning constant strings presumably. Seems like a lot of boilerplate when one single tag and a Selector method is sufficient. This tag can be copied directly from the ported Objective-C code which makes it even less code to write.

The original idea comes from [macdriver](https://github.com/progrium/macdriver) which uses tags. I feel it's an elegant design that mimics the style of original Objective-C code without being a hack.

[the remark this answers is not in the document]
hajimehoshi: What do you think about returning three values from one method?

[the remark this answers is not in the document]
TotallyGamerJet: That's not a bad idea. What would we call the interface since `Selector` doesn't apply now that it has another method? `ClassDefiner`? `Specifier`? `Implementor`? I don't really like any of those. Also, macdriver uses tags to define IBOutlets but I'm not sure if we ever plan on supporting those using RegisterClass?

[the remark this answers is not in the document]
hajimehoshi: > What would we call the interface since Selector doesn't apply now that it has another method? ClassDefiner? Specifier? Implementor?

Just `Class`? Hmm...

OK, let's go with a tag and revisit when we come up with a better idea. It's still v0 so we can change this.
>

example/objc/main_darwin.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (b *barObject) Bar(_cmd objc.SEL) int {
+	return b.bar
+}
+
+func (b *barObject) SetBar(_cmd objc.SEL, bar int) {
+	b.bar = bar
+}
+
+func (_ *barObject) Selector(metName string) objc.SEL {
```
hajimehoshi, line 30: Is `Selector` called once for one name? i.e. isn't `RegisterName` called twice or more for the same method?

hajimehoshi: What if `Selector` returns a string instead of `objc.SEL`?

hajimehoshi: If it is enough to return a string, would it be possible to calculate a method name automatically? I feel like

1. Lower the first character
2. add `:` and argument names if necessary

sounds enough. If a method takes two ore more arguments, things would not be so easy though.

TotallyGamerJet: Yeah my original idea was to add : after if it took arguments. I sadly couldn't think of a good way to handle determining the name for multiple returns tho. Returning a SEL allows the user to optimize bc they may have a global variable with the selector and so RegisterName would only be called once. It also allows for a potential where the SEL is determined in a way we haven't even thought of yet. But if we return just a string then we guarantee to have to call RegisterName

hajimehoshi: >  I sadly couldn't think of a good way to handle determining the name for multiple returns tho.

I was talking about mulitiple arguments, not returning values.

TotallyGamerJet: Yep! I was thinking arguments but accidentally wrote return. ObjC doesn't have multiple returns. 

I see no easy way to determine the ObjC name based on the argument list and name of the Go method

hajimehoshi: OK.

> It also allows for a potential where the SEL is determined in a way we haven't even thought of yet. But if we return just a string then we guarantee to have to call RegisterName

I slightly prefer returning a string to objc.SEL. The risk is, as you said, we would not be able to have a flexible way to get SEL. What do you think?

TotallyGamerJet: Why do you think returning string is more appropriate? Is it because `Selector(string) string` is a possible Interface you'd see in the wild?

hajimehoshi: It's because most of the implementation would just return SELs generated by `objc.RegisterName` so returning SELs seemed redundant.

TotallyGamerJet: I foresee most return a variable that was created using RegisterName so this way each method's RegisterName is only called once instead of twice

hajimehoshi: My point is

```go
	switch metName {
	case "Init":
		return "init"
	case "SetBar":
		return "setBar:"
	case "Bar":
		return "bar"
	default:
		return ""
	}
```

would be simpler and easier for those who have to implement this than

```go
	switch metName {
	case "Init":
		return objc.RegisterName("init")
	case "SetBar":
		return objc.RegisterName("setBar:")
	case "Bar":
		return objc.RegisterName("bar")
	default:
		return 0
	}
```

hajimehoshi: But, I am not confident this is a good idea. For example, returning SEL might be more correct in terms of semantics.

TotallyGamerJet: True if that is the only case of using the Selector. However if you want to call the method that matches that selector you'll likely need a global variable. Global variables are currently used in [ebitengine](https://github.com/hajimehoshi/ebiten/blob/50021ef3e4e3028e81ee77e3763a2b248cc8666f/internal/cocoa/api_cocoa_darwin.go#L36).

```go
var sel_init  = objc.RegisterName("init")
// others...

        switch metName {
	case "Init":
		return sel_init
	case "SetBar":
		return sel_setBar
	case "Bar":
		return sel_bar
	default:
		return 0
	}
```

hajimehoshi: Yeah, it'd be nice if the example reflects this.

But, we still need `RegisterName` for global variables, right? Or do we use the selectors for other usecaes?

TotallyGamerJet: RegisterName needs to be called somewhere bc it's not possible to call an ObjC method without one. 

I don't think there really are any other uses for selectors other than calling ObjC methods

hajimehoshi (approved):
LGTM, thanks!

PR #61, opened 2022-11-06
use R10 instead of R12

### LINKED ISSUE #60: purego: SIGSEGV on amd64 because of NewCallback
See [oto#195](https://github.com/hajimehoshi/oto/issues/195) for more information.

TotallyGamerJet:
On x86-64 R12 is callee-saved which means if we use it in callbackasm1 we MUST preserve its value. We were not doing that so it caused the SIGSEGV to happen in Oto. I replaced R12 with R10 because R10 is caller-saved so we don't need to worry about throwing away the value inside of it.

Source: https://web.stanford.edu/class/archive/cs/cs107/cs107.1174/guide_x86-64.html

Closes #60

### DIFF
--- a/sys_darwin_amd64.s
+++ b/sys_darwin_amd64.s
@@ -84,7 +84,7 @@ TEXT callbackasm1(SB), NOSPLIT, $0
 	MOVQ 0(SP), AX
 	ADDQ $8, SP
 
-	MOVQ 0(SP), R12 // get the return SP so that we can align register args with stack args
+	MOVQ 0(SP), R10 // get the return SP so that we can align register args with stack args
 
 	// make space for first six arguments below the frame
 	ADJSP $6*8, SP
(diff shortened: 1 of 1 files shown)


hajimehoshi (approved):
LGTM, thanks!

Can we have a test for this (calling a calback with many arguments multiple times, for example)?

TotallyGamerJet:
> Can we have a test for this (calling a calback with many arguments multiple times, for example)?

Hmm. `ExampleIMP` already tests our max supported number of arguments. The issue with testing this is that we were using a register improperly. The register in question is not an argument register so we can pass anything to a function. It might be possible to write a test in Go but that could break in any update including minor releases bc it's just testing that we aren't using a register improperly which may change.

TLDR: not really

hajimehoshi:
Sure. Thank you for elaborating! Please merge this.

PR #62, opened 2022-11-07
internal/fakecgo: refactor

TotallyGamerJet:
Carrots make the symbol only available in that source file. We need these functions available outside in order for us to call them.

Use the syscall version of Timespec. That way we can clean up our code and not have duplicated types.

### DIFF
--- a/internal/fakecgo/go_libinit.go
+++ b/internal/fakecgo/go_libinit.go
@@ -20,9 +20,12 @@ func x_cgo_notify_runtime_init_done() {
 //
 //go:nosplit
 func _cgo_try_pthread_create(thread *pthread_t, attr *pthread_attr_t, pfn unsafe.Pointer, arg *ThreadStart) int {
-	var tries int
+	var ts syscall.Timespec
+	// tries needs to be the same type as syscall.Timespec.Nsec
+	// but the fields are int32 on 32bit and int64 on 64bit
+	// we assign tries to whatever type syscall.Timespec.Nsec is
+	var tries = ts.Nsec
 	var err int
-	var ts timespec
 
 	for tries = 0; tries < 20; tries++ {
 		err = int(pthread_create(thread, attr, pfn, unsafe.Pointer(arg)))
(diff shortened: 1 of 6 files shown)


internal/fakecgo/go_libinit.go
```diff
@@ -20,9 +20,12 @@ func x_cgo_notify_runtime_init_done() {
 //
 //go:nosplit
 func _cgo_try_pthread_create(thread *pthread_t, attr *pthread_attr_t, pfn unsafe.Pointer, arg *ThreadStart) int {
-	var tries int
+	var ts syscall.Timespec
+	// tries needs to be the same type as syscall.Timespec.Nsec
+	// but the fields are int32 on 32bit and int64 on 64bit
+	// we assign tries to whatever type syscall.Timespec.Nsec is
```
hajimehoshi, line 26: I don't understand this sentence...

```go
	// tries needs to be the same type as syscall.Timespec.Nsec
	// but the fields are int32 on 32bit and int64 on 64bit.
	// We assign to tries whatever type syscall.Timespec.Nsec is
```

?

TotallyGamerJet: It's just describing why it's tries = ts.Nsec. It's because the type is different on each architecture

hajimehoshi: Yeah, I know the meaning but the sentence seemed tricky (disclaimer: I'm not an English native speaker)

hajimehoshi: So my question is:

1. Is the setence ended at `64bit`?
2. Isn't `assign to tries` correct rather than `assign tries to`?

TotallyGamerJet: The sentence does end after 64Bit. I believe the second sentence is right. It might sound weird bc `tries` is being used as a noun instead of a verb in this case.

hajimehoshi: At least, `We` is better than `we`

hajimehoshi: So, the second sentence means

```
We assign `tries` to something whose type is type syscall.Timespec.Nsec
```

?

TotallyGamerJet: Basically.

Perhaps it's better phrased as `We assign `tries` to syscall.Timespec.Nsec in order to match its type.`

?

hajimehoshi: Yeah I think it's much better

In addition, avoiding `we` is nicer if possible, since it is not clear who is `we`, but this is optional. (We would be able to avoid `we` by using a passive sentence like "`tries` is assigned to ...")

hajimehoshi (approved):
LGTM, thanks!

PR #66, opened 2022-11-09
purego: add Linux support

TotallyGamerJet:
Continuation of #38 

This PR adds Go support for Linux (amd64 and arm64) and CGO support for all other Linux architectures.


.github/workflows/test.yml
```diff
@@ -50,6 +50,10 @@ jobs:
           # Check cross-compiling macOS binaries.
           env GOOS=darwin GOARCH=amd64 go build -v ./...
           env GOOS=darwin GOARCH=arm64 go build -v ./...      
+          
+          # Check cross-compiling Linux binaries.
+          env GOOS=linux GOARCH=amd64 go build -v ./...
+          env GOOS=linux GOARCH=arm64 go build -v ./...
```
hajimehoshi, line 56: What about other architectures? Can we have Cgo fallbacks?

syscall_unknown_linux.go
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build cgo
+// +build cgo
```
hajimehoshi, line 5: We no longer need `// +build`

.github/workflows/test.yml
hajimehoshi: Found syscall_unknown_linux. Thanks

TotallyGamerJet:
@hajimehoshi this is ready for review! :D 

It still doesn't work on go.dev/play but I don't think that's a requirement for addition.

hajimehoshi:
> Sadly, it does not work on [go.dev/playground](https://go.dev/play/p/7Illx4KeAB5). :(

If we do, we can detect the playground by the fixed date (https://go.dev/play/p/RIOCpg_6Rkm), but we don't have to work on this in this PR.

I'll take a look, thanks!

You are still working on macOS version to remove Cgo from Ebitengine/macOS, right?

syscall_linux.go
```diff
@@ -0,0 +1,34 @@
...
+		math.Float64frombits(uint64(a4)), math.Float64frombits(uint64(a5)), math.Float64frombits(uint64(a6)),
+		math.Float64frombits(uint64(a7)), math.Float64frombits(uint64(a8)),
+		r1, r2, err}
+	runtime_cgocall(syscall9XABI0, unsafe.Pointer(&args))
+	return args.r1, args.r2, args.err
+}
+
+func NewCallback(_ interface{}) uintptr {
+	panic("purego: NewCallback not supported")
+}
```
hajimehoshi, line 34: Do we need the definition of `NewCallback` for Linux, if this doesn't work?

syscall_unknown_linux.go
```diff
@@ -0,0 +1,27 @@
...
+	f1, f2, f3, f4, f5, f6, f7, f8         float64
+	r1, r2, err                            uintptr
+}
+
+//go:nosplit
+func syscall_syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2, err uintptr) {
+	return unknown.Syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9)
+}
+
+func NewCallback(_ interface{}) uintptr {
```
hajimehoshi, line 25: ditto

.github/workflows/test.yml
hajimehoshi, line 56: (I don't think we don't do this in this PR but) do we have a plan to support 386 and arm?

internal/fakecgo/symbols_linux.go
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package fakecgo
+
+// we have to use the 3 argument format here :( - 2 argument format is only allowed from inside cgo
```
hajimehoshi, line 6: I don't understand this comment.

internal/fakecgo/go_linux_amd64.go
```diff
@@ -0,0 +1,92 @@
...
+	   here, before we ever call malloc.
+
+	   This is only required for the memory sanitizer, so it's
+	   unfortunate that we always run it.  It should be possible
+	   to remove this when we no longer care about versions of
+	   clang before 3.8.  The test for this is
+	   misc/cgo/testsanitizers.
+
+	   GCC works hard to eliminate a seemingly unnecessary call to
+	   malloc, so we actually use the memory we allocate.  */
```
hajimehoshi, line 77: I think you copied C files in the runtime for `internal/fakecgo/go_*_*.go`. Let's update README, and license headers in these files.

hajimehoshi (commented):
> However, when cgo is enabled you must use internal linking with the command:
> CGO_ENABLED=1 go run -ldflags="-linkmode internal" main.go

Hmm, this sounds a critical as this would break compatibility unfortunately...

syscall_unknown_linux.go
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build cgo
+
+package purego
+
+import "github.com/ebitengine/purego/internal/unknown"
```
hajimehoshi, line 8: What about using the term `cgo` rather than `unknown`?

TotallyGamerJet:
> > However, when cgo is enabled you must use internal linking with the command:
> 
> > CGO_ENABLED=1 go run -ldflags="-linkmode internal" main.go
> 
> 
> 
> Hmm, this sounds a critical as this would break compatibility unfortunately...

This is no longer an issue! I fixed it in make a go work with external

TotallyGamerJet:
> You are still working on macOS version to remove Cgo from Ebitengine/macOS, right?

Yes! I just took a break to work on this.

syscall_linux.go
TotallyGamerJet: I don't think we need it for Oto, and OpenGL doesn't need it nor does Linux gamepad. It'll probably be needed once we try removing glfw but that will be a while before that happens.

.github/workflows/test.yml
TotallyGamerJet: I would love to! It should be really easy. Just write `sys_linux_{386+arm}.s` and `go_linux_{386+arm}.go`. There are also examples for those GOARCHs for windows which could be used as inspiration. However, I don't have systems to test these with so I can't do it :/

internal/fakecgo/symbols_linux.go
TotallyGamerJet: It wasn't needed

internal/fakecgo/go_linux_amd64.go
TotallyGamerJet: The files themselves aren't copied but this comment is. Should I just attribute the whole file?

TotallyGamerJet: I had to revert the commit with this comment bc using malloc fails. 

```
main: malloc.c:2379: sysmalloc: Assertion `(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' failed.
SIGABRT: abort
```

syscall_linux.go
hajimehoshi: OK. I prefer removing this than leaving this with panic.

.github/workflows/test.yml
hajimehoshi: We should be able to use QEMU to confirm, right?

internal/fakecgo/go_linux_amd64.go
hajimehoshi: > Should I just attribute the whole file?

I think yes, we should use Go's license comment instead of ours. Disclaimer: I am not a lawyer so I'm not 100% sure about the license issue.

syscall_cgo_linux.go
```diff
@@ -0,0 +1,29 @@
...
+
+//go:build cgo
+
+package purego
+
+import (
+	"github.com/ebitengine/purego/internal/cgo"
+)
+
+// this is linked to a C function from internal/unknown
```
hajimehoshi, line 12: internal/cgo?

internal/cgo/syscall_cgo_linux.go
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package cgo
+
+// this file moved to package unknown since Cgo and assembly files can't be in the same package
```
hajimehoshi, line 6: package cgo?

internal/fakecgo/go_linux_amd64.go
TotallyGamerJet: Could I do theirs and then follow it with ours?

```
// Copyright 2011 The Go Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

```
Whatever you are comfortable with I'll do. If there becomes some issue we'll just fix the comment.

syscall_linux.go
TotallyGamerJet: It can't be removed since it will fail to compile in RegisterFunc which tries to convert func arguments to C functions.

.github/workflows/test.yml
TotallyGamerJet: True!

internal/fakecgo/go_linux_amd64.go
hajimehoshi: Let's leave Go's, and remove ours.

.github/workflows/test.yml
hajimehoshi: > There are also examples for those GOARCHs for windows which could be used as inspiration.

I missed the word 'Windows'. Testing Windows for arm seems a little tricky.

Anyway I'm looking forward to adding more architectures in the future 😄

README.md
```diff
@@ -13,4 +13,5 @@ This is a list of the copied files:
 * `internal/abi/abi_*.h` from package `runtime/cgo`
 * `internal/fakecgo/callbacks.go` from package `runtime/cgo`
 * `internal/fakecgo/iscgo.go` from package `runtime/cgo`
-* `internal/fakecgo/setenv.go` from package `runtime/cgo`
\ No newline at end of file
+* `internal/fakecgo/setenv.go` from package `runtime/cgo`
+* `internal/fakecgo/go_GOOS_GOARCH.go` from package `runtime/cgo`
```
hajimehoshi, line 17: Let's reorder the items in the alphabetical order:

```
* `internal/fakecgo/callbacks.go` from package `runtime/cgo`
* `internal/fakecgo/go_GOOS_GOARCH.go` from package `runtime/cgo`
* `internal/fakecgo/iscgo.go` from package `runtime/cgo`
* `internal/fakecgo/setenv.go` from package `runtime/cgo`
```

internal/cgo/syscall_cgo_linux.go
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package cgo
+
+// this file moved to package cgo since Cgo and assembly files can't be in the same package
```
hajimehoshi, line 6: This file 'was' moved? When is the original place?

syscall_linux.go
hajimehoshi: Does this mean everything with RegisterFunc doesn't work on Linux?

TotallyGamerJet: Yes, callbacks won't work and floats won't work on Cgo but do without Cgo. Everything else that works on Darwin works on Linux

hajimehoshi: OK I understand.

hajimehoshi: Can we have comments somewhere about what works and what not so that purego users can understand?

TotallyGamerJet: I added the documentation to RegisterFunc!

hajimehoshi: Thanks!

func.go
```diff
@@ -59,6 +61,10 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
 // it will be expanded into a call to the C function as if it had the arguments in that slice.
 // This means that using arg ...interface{} is like a cast to the function with the arguments inside arg.
 // This is not the same as C variadic.
+//
+// There are some limitations when using RegisterFunc on Linux. First, there is no support for function arguments.
+// Second, float32 and float64 arguments and return values do not work when CGO_ENABLED=1. Otherwise, Linux
+// has the same feature parity as Darwin.
```
hajimehoshi, line 67: Is there a difference between Darwin and Windows here?

TotallyGamerJet: Windows doesnt implement RegisterFunc since we just use windows.Syscall on that platform. Otherwise, we'd have to copy all the windows files over into purego and make some modifications to match Darwin and Linux. Doable but seemed like double work to me since I don't think RegisterFunc is really needed for windows.

hajimehoshi: OK, we also want a comment that RegisterFunc doesn't work on Windows.

internal/cgo/syscall_cgo_linux.go
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package cgo
+
+// this file was placed inside internal/cgo and not package purego
```
hajimehoshi, line 6: `this file is placed...`

syscall_linux.go
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
```
hajimehoshi, line 1: The filename should be `syscall_nocgo_linux.go` for consistency

hajimehoshi (approved):
LGTM % one request

Thanks!

PR #67, opened 2022-11-10
purego: SyscallN r2 return float

TotallyGamerJet:
We originally discussed this change (#48) and decided against doing it because r2 usually returns the upper 32bits of a 64bit return value on 32bit platforms. However, on Windows amd64 r2 is actually the float return value.

https://github.com/golang/go/blob/7717ac151ae1556541dddc6a817ac04733f1af44/src/runtime/sys_windows_amd64.s#L66

And there is a TODO for arm64 as well.

https://github.com/golang/go/blob/2cea6cdb6016708e89c9472cdb0504731699681d/src/runtime/sys_windows_arm64.s#L91

Doing this makes supporting float returns easier without having to implement a full C calling convention for RegisterFunc.

### DIFF
--- a/func_darwin.go
+++ b/func_darwin.go
@@ -136,6 +136,8 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 		case reflect.String:
 			v.SetString(strings.GoString(r1))
 		case reflect.Float32, reflect.Float64:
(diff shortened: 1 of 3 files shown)


hajimehoshi (approved):
LGTM

PR #68, opened 2022-11-13
internal/fakecgo: use the same libSystem.dylib version

TotallyGamerJet:
This commit uses the same version of libSystem (B) in all usages.


hajimehoshi (approved):
LGTM

PR #69, opened 2022-11-14
purego: support float arguments in RegisterFunc

TotallyGamerJet:
This PR adds proper float support into RegisterFunc. It puts the first 8 in registers and the 9th onto the stack. It also checks beforehand that there aren't too many arguments.

### DIFF
--- a/func_darwin.go
+++ b/func_darwin.go
@@ -71,7 +71,37 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 	if cfn == 0 {
 		panic("purego: cfn is nil")
 	}
-	// TODO: check to make sure there aren't too many arguments
+	{
+		// this code checks how many registers and stack this function will use
+		// to avoid
+		var ints = 0
+		var floats = 0
+		var stack = 0
+		for i := 0; i < ty.NumIn(); i++ {
+			arg := ty.In(i)
+			switch arg.Kind() {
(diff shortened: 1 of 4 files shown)


func_darwin.go
```diff
@@ -68,6 +68,40 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
 	}
+	if cfn == 0 {
+		panic("purego: cfn is nil")
+	}
+	{
+		// this code checks how many registers and stack this function will use
+		// to avoid crashing with too many arguments
+		var ints = 0
+		var floats = 0
+		var stack = 0
```
hajimehoshi, line 79:
```go
var ints int
var floats int
var stack int
```

func_darwin.go
```diff
@@ -81,37 +115,69 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 				args = tmp
 			}
 		}
-		var sysargs = make([]uintptr, len(args))
+		var sysargs [maxArgs]uintptr
+		var stack = sysargs[numOfIntegerRegisters():]
+		var floats [8]float64
+		numInts := 0
+		numFloats := 0
+		numStack := 0
```
hajimehoshi, line 123:
```go
var numInts int
var numFloats int
var numStack int
```

hajimehoshi (approved):
LGTM

hajimehoshi:
Will we have tests with float numbers later?

TotallyGamerJet:
I do plan on having a test once I finish support for callbacks with float arguments!

PR #70, opened 2022-11-28
internal/fakecgo: generate symbols.go and trampolines_stubs.s

TotallyGamerJet:
These files are very repetitive so it is better to just generate them. It will make adding functions in the future easier.

This also standardizes all the assembly trampoline functions to have zero frame and zero arguments.


internal/fakecgo/symbols.go
```diff
@@ -1,3 +1,4 @@
+// THIS FILE IS GENERATED BY 'go generate' with gen.go. DO NOT EDIT!!!
```
hajimehoshi, line 1: Let's use the standardized comment (see https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source):

```
// Code generated by 'go generate' with gen.go. DO NOT EDIT.

```

internal/fakecgo/trampolines_stubs.s
```diff
@@ -1,3 +1,4 @@
+// THIS FILE IS GENERATED BY 'go generate' with gen.go. DO NOT EDIT!!!
```
hajimehoshi, line 1: ditto

internal/fakecgo/doc.go
```diff
@@ -28,3 +28,6 @@
 // Purego will pick which ever Cgo runtime is available and prefer the one that
 // comes with Go (runtime/cgo).
 package fakecgo
+
+//go:generate go run gen.go
+//go:generate go fmt symbols.go
```
hajimehoshi, line 33: `go fmt -w symbols.go`

internal/fakecgo/gen.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+var funcs = map[string]any{
+	"hasPrefix": strings.HasPrefix,
+}
+
+func main() {
+	t, err := template.New("symbol.go").Funcs(funcs).Parse(templateSymbols)
+	if err != nil {
+		log.Fatal(err)
```
hajimehoshi, line 132: Have a separate function like `run` returning error, and let `main` call `run` to integrate the error handling.

internal/fakecgo/doc.go
hajimehoshi: Ah, I found `gofmt` and `go fmt` are different. I usually do

```
gofmt -s -w symbols.go
```

What do you think?

TotallyGamerJet: `go fmt` doesn't support `-s` so I replaced it with `gofmt` directly.

TotallyGamerJet: Didn't see you had commented! I agree tho

internal/fakecgo/gen.go
```diff
@@ -0,0 +1,163 @@
...
+	t, err = template.New("trampolines_stubs.s").Funcs(funcs).Parse(templateTrampolinesStubs)
+	if err != nil {
+		return err
+	}
+	f, err = os.Create("trampolines_stubs.s")
+	if err != nil {
+		return err
+	}
+	err = t.Execute(f, symbols)
+	if err != nil {
```
hajimehoshi, line 153:
```go
if err := t.Execute(t, symbols); err != nil {
```

internal/fakecgo/gen.go
```diff
@@ -0,0 +1,163 @@
...
+	t, err := template.New("symbol.go").Funcs(funcs).Parse(templateSymbols)
+	if err != nil {
+		return err
+	}
+	f, err := os.Create("symbols.go")
+	if err != nil {
+		return err
+	}
+	err = t.Execute(f, symbols)
+	if err != nil {
```
hajimehoshi, line 141: if err := t.Execute(t, symbols); err != nil {

internal/fakecgo/gen.go
```diff
@@ -0,0 +1,163 @@
...
+
+func run() error {
+	t, err := template.New("symbol.go").Funcs(funcs).Parse(templateSymbols)
+	if err != nil {
+		return err
+	}
+	f, err := os.Create("symbols.go")
+	if err != nil {
+		return err
+	}
```
hajimehoshi, line 139:
```go
defer f.Close()
```

internal/fakecgo/gen.go
```diff
@@ -0,0 +1,163 @@
...
+	t, err = template.New("trampolines_stubs.s").Funcs(funcs).Parse(templateTrampolinesStubs)
+	if err != nil {
+		return err
+	}
+	f, err = os.Create("trampolines_stubs.s")
+	if err != nil {
+		return err
+	}
```
hajimehoshi, line 151:
```go
defer f.Close()
```

internal/fakecgo/trampolines_stubs.s
```diff
@@ -31,7 +33,7 @@ TEXT _pthread_create(SB), NOSPLIT, $0-0
 	JMP purego_pthread_create(SB)
 	RET
 
-TEXT _pthread_detach(SB), NOSPLIT, $0-12
+TEXT _pthread_detach(SB), NOSPLIT, $0-0
```
hajimehoshi, line 36: Is this change fine?

TotallyGamerJet: Yes. It was a lay over from #47 but should have been replaced with `0-0`. It's because the 12 means 12 bytes of arguments but these trampolines don't need to make any room for the arguments since call5 already does that.

hajimehoshi (approved):
LGTM

PR #73, opened 2022-12-01
internal/fakecgo: circumvent Go 1.20's race code

TotallyGamerJet:
Go 1.20 rewrote the race detector on darwin to be entirely Go. This means that fakecgo was having race functions placed inside of it which made it break. This PR circumvents the race functions by adding `//go:norace` and using closures to skip the ABI wrappers.

Closes #71

PR #75, opened 2022-12-02
purego: get ABIInternal wrapper using closure

TotallyGamerJet:
The function runtime.cgocallback takes an ABIInternal function pointer. However, the assembly tag `<ABIInternal>` is only allowed in the runtime. This PR gets the ABIInternal function pointer by using the closure technique from #73. It also removes the hacky trampoline.

### DIFF
--- a/syscall_darwin.go
+++ b/syscall_darwin.go
@@ -114,6 +114,12 @@ const callbackMaxFrame = 64 * ptrSize
 // callbackasmABI0 is implemented in zcallback_GOOS_GOARCH.s
 var callbackasmABI0 uintptr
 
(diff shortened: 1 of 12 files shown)


syscall_darwin.go
```diff
@@ -114,6 +114,12 @@ const callbackMaxFrame = 64 * ptrSize
 // callbackasmABI0 is implemented in zcallback_GOOS_GOARCH.s
 var callbackasmABI0 uintptr
 
+// This closure allows the calling of the ABIInternal wrapper
```
hajimehoshi, line 117: Let's start the comment with `callbackWrap_call`

hajimehoshi (approved):
LGTM

PR #80, opened 2022-12-14
use crosscall2 from runtime/cgo

TotallyGamerJet:
The code originally inlined crosscall2 because at the time there was no internal/fakecgo package to place the implementation. Now that this package exists crosscall2 can be put into it. The implementation is copied from runtime/cgo. When CGO_ENABLED=1 it will use the runtime/cgo implementation of crosscall2 which should always be identical.

Less assembly to maintain is good! :D


hajimehoshi (approved):
LGTM!

> Less assembly to maintain is good! :D

Doesn't this change increset the amount of assembly..?

hajimehoshi:
Ah, the assembly is just copied so the amount of our original assembly is decreased :)

TotallyGamerJet:
Well technically yes. However, the implementation comes from the runtime so it's easy to check if the implementation is correct - just by inspecting the runtime version.

PR #81, opened 2022-12-16
cleanup assembly

TotallyGamerJet:
On amd64, AX should be zero. This commit replaces AX with BX and then zeros AX.

On arm64, R26 is a callee-saved register. Replace R26 with R0 as that register is about to be overwritten anyways.


hajimehoshi (approved):
LGTM

PR #83, opened 2022-12-16
better error message and padding support

TotallyGamerJet:
To check that the Go struct has the same size as the Objective-C class subtract the current class's size from the super class's size. This is necessary because the super class's fields are embedded in the child class but are not present in the Go struct. It is fine that the Go struct does not contain the parent's fields because they are not accessible in Go. The +8 is because the child and super class both contain the isa field which the Go struct does have and needs to be counted in this comparison.

### DIFF
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -295,7 +295,6 @@ func RegisterClass(object Selector) (Class, error) {
 		if err != nil {
 			return 0, fmt.Errorf("objc: couldn't add Ivar %s: %w", f.Name, err)
 		}
-		fmt.Println(f.Name, size, alignment, enc)
 		if !class_addIvar(class, f.Name, size, alignment, enc) {
 			return 0, fmt.Errorf("objc: couldn't add Ivar %s", f.Name)
 		}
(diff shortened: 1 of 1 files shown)


hajimehoshi (approved):
LGTM, thanks!

TotallyGamerJet:
@hajimehoshi This seems to be incorrect.

It appears that the parent class's fields are placed before the child's in the struct. This means that padding is required to properly align any fields if the parent also has fields. Take for example this class:

```go
type contentView struct {
	isa    objc.Class `objc:"GLFWContentView : NSView <NSTextInputClient>"`
	_      [552]byte
	window *Window
	//    NSTrackingArea* trackingArea;
	//    NSMutableAttributedString* markedText;
}
```
The 552 bytes of padding are required because of NSView's fields. This padding could change at any update of the framework as far as I can tell. The good news is that if it does change there will be a runtime error that catches it. However, we probably don't want to be hardcoding the necessary padding. I'm not sure of a good fix without resorting to grabbing the offset from the instance variable and adding it to the `*contentView` which will be unsafe. Avoiding unsafe was the point of adding RegisterClass.

hajimehoshi:
The current implementation is also not correct, right? Does this cause a crash issue?

TotallyGamerJet:
The current implementation is fine bc it will properly check all the fields and sizes to make sure they match. Any issues will be caught at runtime without corrupting any memory. The strong error checking is actually what brought the issue to my attention.

TotallyGamerJet:
Will it cause a crash? Unlikely, since the only classes that use this inherit from NSObject which doesn't have any fields and most likely will never have fields added to it. Plus my previous comment still stands.

hajimehoshi:
I see. So even without this PR, a runtime error will happen if the struct size doesn't match, but this PR improves the situation by checking this explicitly. Is my understanding correct?

TotallyGamerJet:
Yes a runtime error will happen if the structures don't match. However, no this PR actually tries to circumvent the error because I thought that it would ok. I now know that it is not. 

I do however, have a branch that does improves the error message. It also makes my padding example valid. It still has a runtime error if the offsets or sizes don't match but it's not going to introduce mysterious bugs. 

Do you want me to close this and open that one?

hajimehoshi:
> Do you want me to close this and open that one?

Yes, please!

TotallyGamerJet:
Well I accidentally pushed it to this branch. Review and let me know if it is still good?

hajimehoshi:
Still LGTM

PR #89, opened 2023-01-07
objc: add comment about Objective-C global lock

### LINKED ISSUE #86: Can we have a string instead of SEL for `Send`?
I think we have already discussed this, but I was wondering why we have SEL as the first argument of `Send`, not a string.

In most cases, uses make a SEL by `objc.RegisterName` at that place, so isn't this redundant? Sorry if you have already explained the reason.

@TotallyGamerJet

TotallyGamerJet:
Closes #86


hajimehoshi (approved):
LGTM

PR #90, opened 2023-01-07
objc: generic Send and SendSuper

### LINKED ISSUE #87: How can I get a float64 value from ID?
https://developer.apple.com/documentation/foundation/nsnumber/1414104-doublevalue?language=objc

Some method returns a double value, but how can I get a float64 value from ID that is returned from Send?

TotallyGamerJet:
ID.Send and ID.SendSuper can only return integer types (such as ID). These two new functions can take any return type as long as it is a type that RegisterFunc supports.

Closes #87


hajimehoshi (approved):
Thanks!

Will the function registered by RegisterFunc GCed? In other words, if we call `Send` too many times, would this cause memory leak?

hajimehoshi (changes requested):
(I mark this 'request changes' until I'm convinced with the memory issue)

TotallyGamerJet:
The [reflect.MakeFunc](https://pkg.go.dev/reflect#MakeFunc) doc does not say anything about the function being permanent. The function is GCable because it only ever exists in Go code and is not called by C code. This is unlike NewCallback which is made to be called by C code and is therefore impossible to know when it should be collected. Which is why it has the comment that the function pointer is never deallocated.

hajimehoshi (approved):
Thanks!

PR #93, opened 2023-02-12
README.md: add simple example

TotallyGamerJet:


README.md
```diff
@@ -3,6 +3,42 @@
...
+
+	"github.com/ebitengine/purego"
+)
+
+func getSystemLibrary() string {
+	switch runtime.GOOS {
+	case "darwin":
+		return "/usr/lib/libSystem.B.dylib"
+	case "linux":
+		return "libc.so.6"
```
hajimehoshi, line 23: Do we need `.6`?

TotallyGamerJet: On Ubuntu where I tested, yes

TotallyGamerJet: Also [this](https://stackoverflow.com/questions/6841311/using-dlopen-to-get-handle-of-libc-memory-allocation-functions) StackOverflow issue says libc.so is a linker script not an elf file.

hajimehoshi: You mean `dlopen("libc.so")` without `.6` did not work?

hajimehoshi: I remembered that there is an environment that such a symbolic link doesn't exist. NVM

hajimehoshi (approved):
LGTM

README.md
TotallyGamerJet: Without the .6 it fails on Ubuntu

PR #96, opened 2023-02-12
add tests for dlfcn

TotallyGamerJet:
This commit adds two tests. One comes from an example which wasn't really an example of using purego but more of a test to make sure fakecgo was functioning properly. Now it's an actual test. The other is a simple test that calls into libc


dlfcn_test.go
```diff
@@ -0,0 +1,47 @@
+package purego_test
```
hajimehoshi, line 1:
```
```

dlfcn_test.go
```diff
@@ -0,0 +1,50 @@
...
+		return "libc.so.6", nil
+	default:
+		return "", fmt.Errorf("GOOS=%s is not supported", runtime.GOOS)
+	}
+}
+
+func TestRegisterFunc(t *testing.T) {
+	library, err := getSystemLibrary()
+	if err != nil {
+		t.Errorf("couldn't get system library: %s", err)
```
hajimehoshi, line 43: `t.Skipf` is better.

dlfcn_test.go
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 The Ebitengine Authors
+
+//go:build darwin || linux
```
hajimehoshi, line 4: If we use `Skipf`, we don't need this restriction. (Rather, it is better to test TestOS for Windows)

TotallyGamerJet: I believe we still would because it won't compile since there is no `purego.Dlopen` or `purego.RegisterLibFunc` for windows. I think adding stubs is a bad idea since the documentation states that purego.RegisterLibFunc `is not supported on Windows`.

TotallyGamerJet: I can move TestOS into a different file that can be ran on all platforms.

hajimehoshi: Ah I see, that makes sense.

dlfcn_test.go
hajimehoshi: As we discussed, this test runs only on Darwin and Linux, so Errorf is the best.

dlfcn_test.go
hajimehoshi: > I can move TestOS into a different file that can be ran on all platforms.

This sounds better.

hajimehoshi:
I have to go to bed so I'll take a look tomorrow. Thank you for your contribution!

dlfcn_test.go
```diff
@@ -40,7 +27,7 @@ func getSystemLibrary() (string, error) {
 func TestRegisterFunc(t *testing.T) {
 	library, err := getSystemLibrary()
 	if err != nil {
-		t.Errorf("couldn't get system library: %s", err)
+		t.Skipf("couldn't get system library: %s", err)
```
hajimehoshi: Now we have a go:build restriction, this should never come. So Errorf is fine (sorry for going back and forth)

hajimehoshi: Confirmed this is now Errorf, thanks

hajimehoshi (approved):
LGTM, thanks!

PR #98, opened 2023-02-13
purego: simplify code by merging darwin and linux

TotallyGamerJet:
This commit merges the darwin and linux code because they are both unix systems. This is important because it limits the amount of assembly. I found two bugs in the linux implementation while merging the two. Callbacks are still not supported on Linux


syscall_sysv.go
```diff
@@ -37,6 +39,9 @@ func syscall_syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2,
 // for these callbacks is never released. At least 2000 callbacks can always be created. Although this function
 // provides similar functionality to windows.NewCallback it is distinct.
 func NewCallback(fn interface{}) uintptr {
```
hajimehoshi, line 41: We need a comment about Linux.

hajimehoshi (approved):
LGTM

PR #103, opened 2023-02-18
Return errors from `Dlopen`, `Dlsym`, and `Dlclose`

### LINKED ISSUE #100: Return errors from `Dlerror()`
It would be better if `Dlerror()` returned errors, rather than strings as it does now. That would allow users to check errors using a nil check, rather than an empty string check, as well as allow them to return an error from `purego` as an `error` type from their own function. It would also provide the ability to wrap `purego` errors, such as with `%w` in `fmt.Errorf()`, as well as with the new functions added in Go 1.20, such as `errors.Join()`.

`Dlerror()` could simply use `errors.New()` to create a new error with the given string, but I feel it would be even better to define a custom error type that just returns a string. That way, the type information tells the caller that the error came from `purego`.

This would look something like:

```go
type Error struct {
	s string
}

func (e Error) Error() string {
	return e.s
}
```

If this seems like a good idea, I could submit a PR to implement it.

Arsen6331:
This PR returns errors from the `Dlopen`, `Dlsym`, and `Dlclose` functions in order to provide a more idiomatic API.

Fixes #100


README.md
```diff
@@ -41,12 +41,15 @@ func getSystemLibrary() string {
...
 func main() {
-	libc := purego.Dlopen(getSystemLibrary(), purego.RTLD_NOW|purego.RTLD_GLOBAL)
-	if err := purego.Dlerror(); err != "" {
+	libc, err := purego.Dlopen(getSystemLibrary(), purego.RTLD_NOW|purego.RTLD_GLOBAL)
+	if err != nil {
 		panic(err)
 	}
 	var puts func(string)
-	purego.RegisterLibFunc(&puts, libc, "puts")
+	err = purego.RegisterLibFunc(&puts, libc, "puts")
```
hajimehoshi, line 49: `if err := purego.RegisterLibFunc(&puts, libc, "puts"); err != nil`

objc/objc_runtime_darwin.go
```diff
@@ -38,27 +38,72 @@ var (
...
-	purego.RegisterLibFunc(&sel_registerName, objc, "sel_registerName")
-	purego.RegisterLibFunc(&class_getSuperclass, objc, "class_getSuperclass")
-	purego.RegisterLibFunc(&class_getInstanceVariable, objc, "class_getInstanceVariable")
-	purego.RegisterLibFunc(&class_addMethod, objc, "class_addMethod")
-	purego.RegisterLibFunc(&class_addIvar, objc, "class_addIvar")
-	purego.RegisterLibFunc(&class_addProtocol, objc, "class_addProtocol")
-	purego.RegisterLibFunc(&class_getInstanceSize, objc, "class_getInstanceSize")
-	purego.RegisterLibFunc(&ivar_getOffset, objc, "ivar_getOffset")
+	err = purego.RegisterLibFunc(&object_getClass, objc, "object_getClass")
+	if err != nil {
```
hajimehoshi, line 56: ditto

error.go
```diff
@@ -0,0 +1,10 @@
+package purego
+
+// Error represents an error value returned from purego
+type Error struct {
```
hajimehoshi, line 4: @TotallyGamerJet Don't we need to distinguish libdl errors or others? If so, we might want a special type for libdl error

objc/objc_runtime_darwin.go
Arsen6331: Done

error.go
Arsen6331: I added this error type specifically for libdl errors, so I it should already be distinguished from other errors. I can make the name more specific, like `LibdlError`, if you think that would be better.

dlfcn.go
```diff
@@ -33,22 +33,34 @@ func init() {
 // A second call to Dlopen with the same path will return the same handle, but the internal
 // reference count for the handle will be incremented. Therefore, all
 // Dlopen calls should be balanced with a Dlclose call.
-func Dlopen(path string, mode int) uintptr {
-	return fnDlopen(path, mode)
+func Dlopen(path string, mode int) (uintptr, error) {
+	u := fnDlopen(path, mode)
+	errStr := fnDlerror()
+	if errStr != "" {
```
hajimehoshi, line 39:
```go
if errStr := fnDlerror(); errStr != "" {
```

dlfcn.go
```diff
@@ -33,22 +33,34 @@ func init() {
...
 // Dlsym takes a "handle" of a dynamic library returned by Dlopen and the symbol name.
 // It returns the address where that symbol is loaded into memory. If the symbol is not found,
 // in the specified library or any of the libraries that were automatically loaded by Dlopen
 // when that library was loaded, Dlsym returns zero.
-func Dlsym(handle uintptr, name string) uintptr {
-	return fnDlsym(handle, name)
+func Dlsym(handle uintptr, name string) (uintptr, error) {
+	u := fnDlsym(handle, name)
+	errStr := fnDlerror()
+	if errStr != "" {
```
hajimehoshi, line 52: ditto

dlfcn.go
```diff
@@ -33,22 +33,34 @@ func init() {
...
+	return u, nil
 }
 
 // Dlerror returns a human-readable string describing the most recent error that
 // occurred from Dlopen, Dlsym or Dlclose since the last call to Dlerror. It
 // returns an empty string if no errors have occurred since initialization or
 // since it was last called.
+//
+// Deprecated: Use error values returned from Dlopen, Dlsym, and Dlclose instead.
 func Dlerror() string {
```
hajimehoshi, line 64: I feel like we can remove this function instead of deprecating.

@TotallyGamerJet, thoughts?

error.go
```diff
@@ -0,0 +1,10 @@
+package purego
```
hajimehoshi, line 1: Add a license comment at the top of this file

```go
```

dlfcn.go
Arsen6331: Done

error.go
hajimehoshi: `DlError` sounds better (if we decide to remove the `Dlerror` function)

func.go
```diff
@@ -18,12 +18,13 @@ import (
...
 //
 // Windows does not support this function.
-func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
-	sym := Dlsym(handle, name)
-	if sym == 0 {
-		panic("purego: couldn't find symbol: " + Dlerror())
+func RegisterLibFunc(fptr interface{}, handle uintptr, name string) error {
+	sym, err := Dlsym(handle, name)
+	if err != nil {
+		return err
```
hajimehoshi, line 24: Hmm, I'm not confident whether we should return an error here...

error.go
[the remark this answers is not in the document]
TotallyGamerJet: I saw the comment above

TotallyGamerJet: I think Dlerror sounds good

func.go
```diff
@@ -18,12 +18,13 @@ import (
 // It panics if Dlsym fails.
 //
 // Windows does not support this function.
-func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
-	sym := Dlsym(handle, name)
-	if sym == 0 {
-		panic("purego: couldn't find symbol: " + Dlerror())
+func RegisterLibFunc(fptr interface{}, handle uintptr, name string) error {
```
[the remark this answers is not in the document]
hajimehoshi: > Should this still panic instead of return an error? 

I agree.

dlfcn.go
Arsen6331: Done

objc/objc_runtime_darwin.go
```diff
@@ -38,13 +38,19 @@ var (
...
+	objc_msgSend_fn, err = purego.Dlsym(objc, "objc_msgSend")
+	if err != nil {
+		panic("objc: " + err.Error())
 	}
-	objc_msgSend_fn = purego.Dlsym(objc, "objc_msgSend")
 	purego.RegisterFunc(&objc_msgSend, objc_msgSend_fn)
-	objc_msgSendSuper2_fn = purego.Dlsym(objc, "objc_msgSendSuper2")
+	objc_msgSendSuper2_fn, err = purego.Dlsym(objc, "objc_msgSendSuper2")
+	if err != nil {
+		panic("objc: " + err.Error())
```
TotallyGamerJet, line 52: @hajimehoshi I think this error should be wrapped. Do you agree?

error.go
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 The Ebitengine Authors
+
+package purego
+
+// Dlerror represents an error value returned from purego
```
TotallyGamerJet, line 6: This error isn't a generic purego error. It only occurs when using dlopen, dlsym, or dlclose. I think the doc should express that

TotallyGamerJet (changes requested):

objc/objc_runtime_darwin.go
hajimehoshi: I don't understand. Isn't err already a DlError?

hajimehoshi:
Please check the test results.

objc/objc_runtime_darwin.go
TotallyGamerJet: Yes but if we wanted the context that this error happens in package objc. Otherwise, just panic with the error

Arsen6331:
> Please check the test results.

The `examples` directory it mentions doesn't exist in my PR, it was only added to main yesterday, along with the `libc` example.

hajimehoshi:
> The examples directory it mentions doesn't exist in my PR, it was only added to main yesterday, along with the libc example.

OK so please rebase or merge your PR and fix the example to pass the test.

objc/objc_runtime_darwin.go
hajimehoshi: I don't understand your suggestion yet. Are you suggesting to create a new error type for Objc? Even if so, this line would not change, right?

Arsen6331:
> OK so please rebase or merge your PR and fix the example to pass the test.

Done

Arsen6331:
Fixed the failing test

TotallyGamerJet (commented):
LGTM!

I'll let @hajimehoshi take one last look before merge.

hajimehoshi (approved):
LGTM!

objc/objc_runtime_darwin.go
hajimehoshi: Was this resolved?

image

TotallyGamerJet (approved):
LGTM!

hajimehoshi:
As this is a breaking change,

* I tagged the latest commit `0befdc3778764be7c9997dbe0326b2761eeafcd1` as `v0.3.0-alpha`
* I tagged the previous commit `0c858ac9f9ff921a4d24b833c0354cda7fdafa1b` as `v0.2`
* I created a branch from `v0.2`.

Issue #107, opened 2023-03-16
Stubs for other platforms?


klauspost:
I would like to use this for cross-platform development, and use purego where supported, but keep code compiling on all platforms.

However most functions, like `Dlopen(...)` aren't available for Windows, which is fair enough, but I would need to guard all code with build tags. 

In these cases it is very nice to have stubs available, so code compiles, but just returns errors:

```
var ErrUnsupported = errors.New("purego not available on this platform")

func Dlopen(path string, mode int) (uintptr, error) {
	return 0, ErrUnsupported
}

func Dlclose(handle uintptr) error {
	return ErrUnsupported 
}

func Dlsym(handle uintptr, name string) (uintptr, error) {
	return 0, ErrUnsupported
}

// etc
```

For my projects this would mean a lot less code duplication. I don't really see any significant downside to adding stubs.

klauspost:
Not really. In my case I must have a "software" fallback even for Linux, where the library may or may not be installed.

For me that would mean that I can keep the fallback codepath operating the same, and the code would be...

```
	handle, err := purego.Dlopen("libxyz.so", purego.RTLD_NOW|purego.RTLD_GLOBAL)
	if err != nil {
		return err  // error triggers fallback upstream.
	}
```

So in this case it can all be done without build tags at all.

TotallyGamerJet:
That makes sense. I can't think of a reason not to add the stubs. @hajimehoshi can you? I know we do stub NewCallback on Linux but that panics. Perhaps it would use the unsupported error. But perhaps that a non-backwards compatible change?

hajimehoshi:
I still think that build tags should work instead of stubs. Is the Dlopen code path used in Windows?

hajimehoshi:
> I know we do stub NewCallback on Linux but that panics.

Perhaps we should not have defined this for Linux in the first place...

TotallyGamerJet:
The stub was necessary for RegisterFunc since it calls it if the argument is a function. NewCallback can be implemented on Linux. There just hasn't been much reason for Ebitengine. Compared to Darwin where there are tons of OS APIs that need a callback.

hajimehoshi:
On second thought, having the stub for NewCallback for POSIX OSes makes sense so that we can integrate the same code for POSIX OSes including macOS and Linux.

I still don't think we need stubs for Dlopen for Windows though...

TotallyGamerJet:
[...]

PR #109, opened 2023-03-17
purego: undefine Dlerror for Windows

hajimehoshi:
Let's wait for the discussion at #107

### DIFF
--- a/dlerror.go
+++ b/dlerror.go
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023 The Ebitengine Authors
 
+//go:build darwin || linux
+
 package purego
 
 // Dlerror represents an error value returned from Dlopen, Dlsym, or Dlclose.


dlerror.go
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023 The Ebitengine Authors
 
+//go:build darwin || linux
```
TotallyGamerJet, line 4: This actually removes all platforms except Darwin and Linux. Shouldn't it be `!windows` to just remove windows.

TotallyGamerJet (commented):
Obviously this is a backwards breaking change. How are we handling that?

Will this PR include the stubs or should I make that a separate PR

dlerror.go
hajimehoshi: Dl* is supported only by darwin or linux. If we want to use `!windows`, we should update all go:build for consistency. What do you think?

hajimehoshi:
> Obviously this is a backwards breaking change. How are we handling that?

As purego is still in v0, I'm fine with this change. I guess nobody uses Dlerror for Windows.

> Will this PR include the stubs or should I make that a separate PR

As I'm against adding stubs, this PR doesn't aim to add any other stubs.

TotallyGamerJet (approved):
LGTM!

hajimehoshi:
As I commented first, let's wait for the conclusion at #107 :-) (That's why this PR is draft)

hajimehoshi:
OK we have reached an agreement :-)

PR #104, opened 2023-02-20
internal/fakecgo: separate Linux and Darwin trampolines

TotallyGamerJet:
#73 broke CGO_ENABLED=0 on Linux amd64 because Linux doesn't need the workaround for //go:norace. This commit separates Darwin and Linux implementations.

Closes #101

PR #111, opened 2023-03-19
internal/fakecgo: align stack on amd64 for SSE

### LINKED ISSUE #106: SSE instruction segfaults fakecgo init
Some SSE instructions have requirements that make the fakecgo package segfault. In my case, the call to `malloc` during `x_cgo_init` in [`internal/fakecgo/go_linux_amd64.go`](https://github.com/ebitengine/purego/blob/main/internal/fakecgo/go_linux_amd64.go#L81) makes use of a [`movaps`](https://c9x.me/x86/html/file_module_x86_id_180.html) instruction on `0x10(%rsp)`. But since `%rsp` was not aligned on 16 bits (as the instruction requires it), the cgo initialization segfaults. Fixing the register manually using gdb does the trick but of course it messes with the stack and segfault laters.

My team and I are really fond and the project and really want it to work on linux so I will probably drop a PR if I find a fix in the following days. If you have a definitive idea on how to fix this or if you want more debug information, please let me know.

TotallyGamerJet:
Some SSE instructions on amd64 require that the stack be aligned. The stack was not properly aligned and would SIGSEV inside fakecgo. This patch aligns the stacks properly. It also merges the darwin and linux trampolines. This is a reverse of #104 which I believe was actually caused by this issue. It also explicitly tests the behavior with and without Cgo (CGO_ENABLED=0/1).

Closes #106


hajimehoshi (approved):
LGTM

hajimehoshi:
I cherry-picked this to 0.3 (the current stable branch).

PR #117, opened 2023-03-24
purego: fix callback crash for arm64

erdichen:
crosscall2 clobbers FP in the frame record. Use NOFRAME flag for callbackasm1 to disable the assembler generated epilogue that restores the corrupted FP. Save and restore FP manually instead.

### DIFF
--- a/callback_test.go
+++ b/callback_test.go
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 The Ebitengine Authors
+
+//go:build darwin
+
+package purego_test
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"unsafe"
+
+	"github.com/ebitengine/purego"
+)
+
+// TestCallGoFromSharedLib is a test that checks for stack corruption on darwin/arm64
+// when C calls Go code from a non-Go thread in a dynamically loaded share library.
+func TestCallGoFromSharedLib(t *testing.T) {
+	libFileName := filepath.Join(t.TempDir(), "libcbtest.so")
+	t.Logf("Build %v", libFileName)
+
+	if err := buildSharedLib(libFileName, filepath.Join("libcbtest", "callback.c")); err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(libFileName)
+
+	lib, err := purego.Dlopen(libFileName, purego.RTLD_NOW|purego.RTLD_GLOBAL)
+	if err != nil {
+		t.Fatalf("Dlopen(%q) faled: %v", libFileName, err)
+	}
+
+	var callCallback func(p uintptr, s string) int
+	purego.RegisterLibFunc(&callCallback, lib, "callCallback")
+
+	goFunc := func(cstr *byte, n int) int {
+		s := string(unsafe.Slice(cstr, n))
+		t.Logf("FROM Go: %s\n", s)
+		return 1
+	}
+
+	const want = 10101
+	cb := purego.NewCallback(goFunc)
+	for i := 0; i < 10; i++ {
+		got := callCallback(cb, "a test string")
+		if got != want {
+			t.Fatalf("%d: callCallback() got %v want %v", i, got, want)
+		}
+	}
+}
+
+func buildSharedLib(libFile string, sources ...string) error {
+	out, err := exec.Command("go", "env", "CC").Output()
+	if err != nil {
+		return fmt.Errorf("go env error: %w", err)
+	}
+
+	cc := strings.TrimSpace(string(out))
+	if cc == "" {
+		return errors.New("CC not found")
+	}
+
+	args := append([]string{"-shared", "-Wall", "-Werror", "-o", libFile}, sources...)
+	cmd := exec.Command(cc, args...)
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("compile lib: %w\n%q\n%s", err, cmd, string(out))
+	}
+
+	return nil
+}
--- a/libcbtest/callback.c
+++ b/libcbtest/callback.c
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 The Ebitengine Authors
+
+#include <string.h>
+
+typedef int (*callback)(const char *, int);
+
+int callCallback(const void *fp, const char *s) {
+    // If the callback corrupts FP, this local variable on the stack will have incorrect value.
+    int sentinel = 10101;
+    ((callback)(fp))(s, strlen(s));
+    return sentinel;
+}
--- a/sys_arm64.s
+++ b/sys_arm64.s
@@ -63,7 +63,7 @@ TEXT syscall9X(SB), NOSPLIT, $0
 	FMOVD F0, syscall9Args_r2(R2) // save r2
 	RET
 
-TEXT callbackasm1(SB), NOSPLIT, $208-0
+TEXT callbackasm1(SB), NOSPLIT|NOFRAME, $0
 	NO_LOCAL_POINTERS
 
 	// On entry, the trampoline in zcallback_darwin_arm64.s left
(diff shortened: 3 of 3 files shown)


hajimehoshi:
Can we add a test which crashes without the fix?

TotallyGamerJet:
I did run this PR with Oto and Ebitengine and all the tests passed which is good. I do agree that a test that fails without this should be added.

callback_test.go
```diff
@@ -0,0 +1,50 @@
+//go:build manual_test
```
hajimehoshi, line 1: Isn't it possible to remove the build tag by compiling the C program on the fly (`os/exec`)?

callback_test.go
```diff
@@ -0,0 +1,50 @@
...
+	purego.RegisterLibFunc(&initLib, lib, "initLib")
+	initLib()
+
+	var printString func(p uintptr, s string)
+	purego.RegisterLibFunc(&printString, lib, "printString")
+
+	var wg sync.WaitGroup
+
+	goFunc := func(cstr *byte, n int) int {
+		a := (*[1<<30 - 1]byte)(unsafe.Pointer(cstr))[:n]
```
TotallyGamerJet, line 36: This could be `string(unsafe.Slice(cstr, n))`

callback_test.go
erdichen: Done

callback_test.go
erdichen: Done. Since `unsafe.String` requires 1.20, I falsely assumed the same for `unsafe.Slice`.

callback_test.go
```diff
@@ -0,0 +1,83 @@
+//go:build darwin
```
hajimehoshi, line 1: Add a copy right comment

```
```

hajimehoshi, line 1: Does this test only for Darwin? What about Linux?

libcbtest/Makefile
```diff
@@ -0,0 +1,2 @@
+libcbtest.so: callback.c
+	$(CC) -shared -Wall -Werror -o $@ $^
```
hajimehoshi, line 2: We no longer need this file

libcbtest/callback.c
```diff
@@ -0,0 +1,55 @@
+#include <stdint.h>
```
hajimehoshi, line 1: Add a copy right comment

```
```

callback_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/ebitengine/purego"
+)
+
+// TestCallGoFromSharedLib is a test that checks for stack corruption on darwin/arm64
+// when C calls Go code from a non-Go thread in a dynamically loaded share library.
+func TestCallGoFromSharedLib(t *testing.T) {
+	libFileName := filepath.Join(t.TempDir(), "libcbtest.so")
+	t.Logf("Build %v", libFileName)
+
+	if err := buildSharedLib(libFileName, "libcbtest/callback.c"); err != nil {
```
hajimehoshi, line 25: `filepath.Join("libctest", "callback.c")`

callback_test.go
```diff
@@ -0,0 +1,83 @@
...
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		printString(cb, "a test string")
+	}
+
+	wg.Wait()
+}
+
+func buildSharedLib(libFile string, sources ...string) error {
+	out, err := exec.Command("go", "env", "-json").Output()
```
hajimehoshi, line 61: running `go env CC` and trimming the result by `strings.TrimSpace` should be fine. We don't need parsing JSON.

libcbtest/callback.c
```diff
@@ -0,0 +1,55 @@
...
+            head = head->next;
+
+            printf("FROM C: %s\n", n->s);
+            int size = ((callback)(n->fp))(n->s, strlen(n->s));
+            printf("Back from callback: %d\n", size);
+
+            free(n->s);
+            free(n);
+        }
+        pthread_mutex_unlock(&mp);
```
hajimehoshi, line 36: Do we need manipulating nodes in order to reproduce the issue? Or would just using a mutex be fine?

hajimehoshi (commented):
Thank you!

libcbtest/callback.c
hajimehoshi, line 1: We need `//go:build ignore` for this file, as this file is unexpectedly included to compile the test.

hajimehoshi: Never mind, the directory was separated.

callback_test.go
erdichen: Done

erdichen: Linux does not support callback. I don't have arm64 Windows.

libcbtest/Makefile
erdichen: Done

libcbtest/callback.c
erdichen: Done

callback_test.go
erdichen: Done

callback_test.go
erdichen: Done

libcbtest/callback.c
erdichen: The crash happens after the Go callback returns and the C code access local variables because the frame pointer is incorrect. It was strange when the thread start function is compiled by cgo and is part of the main Go binary, it did not crash. I checked that FP was still overwritten but did not crash. FP was replaced by LR since the crosscall2 frame overlaps by one machine word, so it may just be a coincidence. Access a local variable that points to heap memory is more likely to trigger the crash.

callback_test.go
TotallyGamerJet: Windows doesn't go through this so just Darwin is fine

callback_test.go
```diff
@@ -0,0 +1,81 @@
...
+		s := string(unsafe.Slice(cstr, n))
+		t.Logf("FROM Go: %s\n", s)
+		wg.Done()
+		return 1
+	}
+
+	cb := purego.NewCallback(goFunc)
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		printString(cb, "a test string")
```
TotallyGamerJet, line 57: Is this supposed to be spawning a goroutine? Since you have WaitGroup

erdichen: It is waiting for the async callback from the C pthread.

libcbtest/callback.c
hajimehoshi: If we can get the minimized test to reproduce the issue, this would be the best. I'm still not sure the algorithm for the nodes and/or printf outputs. If these are likely needed, please leave a comment to explain why we need these code.

callback_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	cc := strings.TrimSpace(string(out))
+	if cc == "" {
+		return errors.New("CC not found")
+	}
+
+	args := append([]string{"-shared", "-Wall", "-Werror", "-o", libFile}, sources...)
+	cmd := exec.Command(cc, args...)
+	if _, err := cmd.Output(); err != nil {
```
hajimehoshi, line 76: This ignores stderr. CombinedOutput seems fine

erdichen: Done. Also log the output.

libcbtest/callback.c
erdichen: Turns out just checking a local variable against a known value is good enough. Just need to make sure to disable optimization so the C compiler does not optimize out the local variable.

libcbtest/callback.c
```diff
@@ -0,0 +1,9 @@
+#include <string.h>
+
+typedef int (*callback)(const char *, int);
+
+int printString(const void *fp, const char *s) {
```
hajimehoshi, line 5: What about just `callCallback`? This function itself doesn't call printf.

libcbtest/callback.c
```diff
@@ -0,0 +1,9 @@
...
+    int sentinel = 10101;
```
hajimehoshi, line 6: What is this?

libcbtest/callback.c
erdichen: Done

libcbtest/callback.c
erdichen: Added comment. If the FP is corrupted, the local variable on the stack will have a different value.

hajimehoshi (approved):
LGTM, thank you very much!

I would like to wait for @TotallyGamerJet's review.

TotallyGamerJet:
Does callback.c need a license stub? If not, LGTM.

PR #119, opened 2023-03-30
dlfcn: rework error check

### LINKED ISSUE #118: Support for static initialization dlopen/dlsym calls
Hello,

I noticed that in [`dlopen`](https://github.com/ebitengine/purego/blob/main/dlfcn.go#L37) and [`dlsym`](https://github.com/ebitengine/purego/blob/ebd8567e49db0b4514b146f222a59dd1acd618b7/dlfcn.go#L49) in `dlfcn.go` , the return value is never checked against nil. Is there a reason for this ? This blocks the use of any call to dlopen or any similar function in the static initialization of the library loaded. In the end, if the inner call fails, the dlerror value set by the inner call would make the top call fail.

Do tell me if you need more precisions or if you want me to do a PR or an MRE.

eliottness:
This PR implements the fix described [here](https://github.com/ebitengine/purego/issues/118#issuecomment-1489992006)

The test added needs the testing environment to possess a C++ compiler, feel free to tell me if you have another idea to reproduce this behaviour without this new dependency.

Closes #118

### DIFF
--- a/dlfcn_test.go
+++ b/dlfcn_test.go
@@ -7,7 +7,11 @@ package purego_test
 
 import (
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 	"unsafe"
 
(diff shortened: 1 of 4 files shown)


dlfcn_test.go
```diff
@@ -90,3 +94,33 @@ func Test_qsort(t *testing.T) {
 		}
 	}
 }
+
+
```
hajimehoshi, line 98: One empty line is ifne (Probably you need gofmt)

eliottness: Applied, thanks.

hajimehoshi (approved):
LGTM, thanks!

I'd like to wait for @TotallyGamerJet's review

PR #122, opened 2023-04-03
purego: enable testing of amd64 with Rosseta on arm64 macOS

TotallyGamerJet:
With Rosetta on arm64 it was possible to test amd64 just by setting `GOARCH=amd64`. However, since the shared library tests were added it is no longer possible because the default compiler will build for the host architecture and not the cross-compiled (amd64) architecture. This commit tells the C/C++ compiler to build for amd64 when running an amd64 binary. This only effects macOS.

### DIFF
--- a/dlfcn_test.go
+++ b/dlfcn_test.go
@@ -6,6 +6,7 @@
 package purego_test
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


dlfcn_test.go
```diff
@@ -123,3 +112,43 @@ func TestNestedDlopenCall(t *testing.T) {
...
+		switch runtime.GOARCH {
+		case "arm64":
+			arch = "arm64"
+		case "amd64":
+			arch = "x86_64"
+		default:
+			// if GOARCH is unknown then build the shared library
+			// with the compiler's normal target.
+			archFlag = ""
+			arch = ""
```
hajimehoshi, line 144: Is this reachable? If not, what about returning an error?

TotallyGamerJet: It probably wouldn't

hajimehoshi (approved):
LGTM

PR #126, opened 2023-04-10
purego: check for nil callback

TotallyGamerJet:
It was possible to create a C callback for a nil function but this shouldn't be possible. This commit makes it so that NewCallback will panic on a nil function.

### DIFF
--- a/syscall_sysv.go
+++ b/syscall_sysv.go
@@ -6,7 +6,6 @@
 package purego
 
 import (
-	"fmt"
 	"reflect"
 	"runtime"
 	"sync"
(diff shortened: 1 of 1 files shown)


syscall_sysv.go
```diff
@@ -75,6 +75,9 @@ func compileCallback(fn interface{}) uintptr {
 	if val.Kind() != reflect.Func {
 		panic("purego: type is not a function")
 	}
+	if val.IsNil() {
+		panic("purego: function is nil")
```
hajimehoshi, line 79: `function must not be nil`?

syscall_sysv.go
```diff
@@ -75,6 +75,9 @@ func compileCallback(fn interface{}) uintptr {
 	if val.Kind() != reflect.Func {
 		panic("purego: type is not a function")
```
hajimehoshi, line 76: `the type must be a function but was not`?

hajimehoshi (commented):
I feel like panic messages should suggest what action the users should take in general.

hajimehoshi (approved):
LGTM

Issue #121, opened 2023-04-01
Add GoString and CString functions to match C package API


TotallyGamerJet:
When using the the special C package there are a few functions that convert from Go types to C types and vice versa. When porting code that uses C it's annoying not having an equivalent to these functions. Especially, `GoString` and `CString` which convert a C string to a Go string and vice versa. Even [ebitengine](https://github.com/hajimehoshi/ebiten/blob/0d1f990338331361de268f4912b26eaac783d198/internal/goglfw/context_windows.go#L594) had to have it's own implementation and writing these over and over is begging for mistakes. Purego already uses it's own internal [versions](https://github.com/ebitengine/purego/blob/main/internal/strings/strings.go).

I purpose adding these two functions to purego with the following signatures.
```go
// Cstring converts a Go string to a null-terminated C string.
// The C string is allocated in the C heap using malloc.
// It is the caller's responsibility to arrange for it to be
// freed, such as by calling free.
func CString(string) *byte

// GoString converts a null-terminated C string and copies it into
// a Go allocated string.
func GoString(*byte) string
```

I believe these are the most used functions so should be added first. If there is strong demand later for the other ones a new issue can be created.

hajimehoshi:
Does `CString` add a null-termination automatically?

TotallyGamerJet:
> Does `CString` add a null-termination automatically?

Yes. That's what the C package does. Matching it's behavior would be the least surprising

> How does a user call free with purego?

```
var free func(unsafe.Pointer)
purego.RegisterFunc(&free, freeHandle)
free(unsafe.Pointer(cStr))
```

And freeHandle can be gotten using `Dlsym`

hajimehoshi:
> Yes. That's what the C package does. Matching it's behavior would be the least surprising

Sure, I'm fine if this is explicitly documented.

> And freeHandle can be gotten using Dlsym

So do we dlopen libc?

TotallyGamerJet:
> So do we dlopen libc?

I don't think purego should export free nor libc in its API. How someone wants to free it is up to the user

hajimehoshi:
> I don't think purego should export free nor libc in its API. How someone wants to free it is up to the user

Doesn't this depend on how `CString` allocate the string data, right?

hajimehoshi:
If `CString` uses `malloc` from libc, a user would have to use `free` from libc. I'd like to know your plan how to implement `CString`.

TotallyGamerJet:
[...]

PR #128, opened 2023-04-10
purego: decrease number of allocs in RegisterFunc

TotallyGamerJet:
This rewrites part of the RegisterFunc to have less code escape to the heap. Running the benchmark inside #121 shows a speed increase and alloc decrease.

```
goos: darwin
goarch: arm64
pkg: github.com/ebitengine/purego
BenchmarkCStringC
BenchmarkCStringC-10    	 2294838	       500.3 ns/op	     472 B/op	       8 allocs/op
PASS
```


hajimehoshi (approved):
LGTM

PR #131, opened 2023-04-15
purego: Windows support for RegisterLibFunc and RegisterFunc

### LINKED ISSUE #130: examples/libc: doesn't work on Windows
Is Windows O.S. support planned ?

TotallyGamerJet:
It implements these two functions to make the code between the two more similiar. It also updates the libc example to work on Windows.

Closes #130

P.S. This was quite a lot more changes than I expected so please take a good look since I may have made a mistake. Also please test this since I only tested it on Windows/arm64.


examples/libc/main.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+// SPDX-FileCopyrightText: 2023 The Ebitengine Authors
```
hajimehoshi, line 2: We don't have to update the year for an existing file. (This is the same rule as Go and Chromium)

hajimehoshi:
In the ideal world, we should make it available on 32bit Windows, but the priority should be pretty low.

hajimehoshi:
Confirmed `libc` worked on Windows/amd64!

```
purego> go run .\examples\libc\
go: downloading golang.org/x/sys v0.7.0
Calling C from Go without Cgo!
```

TotallyGamerJet:
> In the ideal world, we should make it available on 32bit Windows, but the priority should be pretty low.

It should work for the majority of cases since it falls back to the syscall9 implementation.

hajimehoshi:
```
 go run .\examples\libc\
panic: purego: unknown GOARCH (386)

goroutine 1 [running]:
github.com/ebitengine/purego.numOfIntegerRegisters(...)
        //Mac/Home/purego/func.go:261
github.com/ebitengine/purego.RegisterFunc({0x4a96c0, 0x940a2d8}, 0x76b848d0)
        //Mac/Home/purego/func.go:92 +0x28f
github.com/ebitengine/purego.RegisterLibFunc({0x4a96c0, 0x940a2d8}, 0x76ad0000, {0x4ba08a, 0x4})
        //Mac/Home/purego/func.go:24 +0x5d
main.main()
        //Mac/Home/purego/examples/libc/main.go:34 +0x85
exit status 2
```

GOARCH=386 didn't work.

TotallyGamerJet:
Try now

func.go
```diff
@@ -257,6 +257,8 @@ func numOfIntegerRegisters() int {
 		return 8
 	case "amd64":
 		return 6
+	case "386":
+		return 0
```
hajimehoshi, line 261: What about `arm` (32bit)?

hajimehoshi:
`GOARCH=386` worked, thanks! (I cannot test GOARCH=arm on my machine. Please try `$env:GOARCH = "arm"`)

hajimehoshi:
Let's add tests for 32bit Windows (386):

```yaml
      - name: go test (Windows 386)
        if: runner.os == 'Windows'
        run: |
          env GOARCH=386 CGO_ENABLED=0 go test -shuffle=on -v -count=10 ./...
          env GOARCH=386 CGO_ENABLED=0 go test -shuffle=on -v -count=10 ./...
```

TotallyGamerJet:
> `GOARCH=386` worked, thanks! (I cannot test GOARCH=arm on my machine. Please try `$env:GOARCH = "arm"`)

Did I do it right? I only have a window/arm64 not arm so I'm not sure if there is some Rosetta equivalent.

```
PS Z:\Documents\GolandProjects\purego> $env:GOARCH = "arm"
PS Z:\Documents\GolandProjects\purego> go test ./...
fork/exec C:\Users\JARRET~1\AppData\Local\Temp\go-build312348259\b001\purego.test.exe: This version of %1 is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher.
FAIL    github.com/ebitengine/purego    8.022s
?       github.com/ebitengine/purego/cmd        [no test files]
?       github.com/ebitengine/purego/examples/libc      [no test files]
?       github.com/ebitengine/purego/internal/strings   [no test files]
FAIL
```

hajimehoshi:
> Did I do it right? I only have a window/arm64 not arm so I'm not sure if there is some Rosetta equivalent.

Not sure. Would a simple Go program like hello-world work with GOARCH=arm on your Windows Arm for example?

TotallyGamerJet:
It appears the 32bit tests failed. Did the tests pass on your system or just the libc example?

hajimehoshi:
> It appears the 32bit tests failed. Did the tests pass on your system or just the libc example?

I don't have Arm Windows so both arm and arm64 don't work.

Have you tried a simple Go program with GOARCH=arm by the way?

TotallyGamerJet:
> I don't have Arm Windows so both arm and arm64 don't work.

The 386 GitHub test is what failed not arm. The arm64 tests are passing on my windows machine.

> Have you tried a simple Go program with GOARCH=arm by the way?

Tried and it does not work. Only an arm64 binary runs on my Windows/arm64 machine.

hajimehoshi:
Apparently GOARCH=386 tests failed.

If the fix is not obvious, it's OK to work on 32bit Windows on another machine. What do you think?

hajimehoshi:
> Tried and it does not work. Only an arm64 binary runs on my Windows/arm64 machine.

I see. Let's give up GOARCH=arm so far.

TotallyGamerJet:
Yeah it's not clear to me why it's failing. I pushed a commit to remove 32bit in test and remove them from the numOfIntRegisters so it will panic again on 32bit windows.

For my reference later, here is what ChatGPT says about the exception:

> The exception code you provided, 0xc0000005, indicates that the program has encountered an access violation error. This means that the program attempted to access a memory location that it was not allowed to access, which could be caused by a variety of issues such as null pointers, uninitialized variables, or memory corruption.
>
> The other values in the exception code are memory addresses that can help pinpoint the location of the error. The second value, 0x0, is the address of the instruction that caused the exception. The third value, 0xdd3a8e2c, is the memory address that the program attempted to access, and the fourth value, 0x771fa519, is the address of the Windows kernel function that raised the exception.

hajimehoshi (approved):
LGTM

hajimehoshi:
I hope ChatGPT would create a PR to fix our issues in the future 😁

TotallyGamerJet:
> I hope ChatGPT would create a PR to fix our issues in the future 😁

That would be great! haha

calvin2021y:
hi @TotallyGamerJet  @hajimehoshi 

After this merged,  this mean `RegisterFunc` only work for `Windows extern functions` ?

There is also "extern C" function from windows dynamic library(like some MingGW or cygwin projects),  so I think should add one new Function for windows,  and just keep the   `RegisterFunc` not changed.

Or add function like `RegisterCFunc`,  `RegisterWindowsFunc`.

TotallyGamerJet:
If you are proposing a change create a new issue and describe in detail what you want. I'm unclear what you want added. This commit adds support for [C ABI functions](https://www.agner.org/optimize/calling_conventions.pdf).

PR #135, opened 2023-05-18
dlfcn: add RTLD_DEFAULT macro value

### LINKED ISSUE #134: Give the ability to perform dlopen(NULL)
As per the man page, `dlopen(NULL)` is supposed to return the equivalent of the macro `RTLD_DEFAULT` which not accessible from go code obviously and is implementation defined. This function is especially useful to run already loaded symbols like the ones loaded with the pragma `go:cgo_import_dynamic`. 

But since we pass a string to Dlopen, a nil value cannot be passed as parameter. How could we give access to this feature ?

eliottness:
Closes #134

### DIFF
--- a/dlfcn_test.go
+++ b/dlfcn_test.go
@@ -18,6 +18,12 @@ import (
 	"github.com/ebitengine/purego"
 )
 
+func TestSimpleDlsym(t *testing.T) {
+	if _, err := purego.Dlsym(purego.RTLD_DEFAULT, "dlsym"); err != nil {
+		t.Errorf("Dlsym with RTLD_DEFAULT failed: %v", err)
+	}
+}
+
 func TestNestedDlopenCall(t *testing.T) {
 	libFileName := filepath.Join(t.TempDir(), "libdlnested.so")
 	t.Logf("Build %v", libFileName)
--- a/dlfcn_darwin.go
+++ b/dlfcn_darwin.go
@@ -6,10 +6,11 @@ package purego
 // Source for constants: https://opensource.apple.com/source/dyld/dyld-360.14/include/dlfcn.h.auto.html
 
 const (
-	RTLD_LAZY   = 0x1 // Relocations are performed at an implementation-dependent time.
-	RTLD_NOW    = 0x2 // Relocations are performed when the object is loaded.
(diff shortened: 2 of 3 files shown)


dlfcn_test.go
```diff
@@ -14,10 +14,18 @@ import (
 	"runtime"
 	"strings"
 	"testing"
+	_ "unsafe" // for go:linkname
```
TotallyGamerJet, line 17: I don't see the go linkname

TotallyGamerJet (changes requested):

dlfcn_test.go
eliottness: It's an artifact from a previous version, removing it
```suggestion
```

TotallyGamerJet (commented):
LGTM! @hajimehoshi, do you approve the API addition?

dlfcn_test.go
```diff
@@ -18,6 +18,13 @@ import (
 	"github.com/ebitengine/purego"
 )
 
+func TestSimpleDlsym(t *testing.T) {
+	_, err := purego.Dlsym(purego.RTLD_DEFAULT, "dlsym")
+	if err != nil {
```
hajimehoshi, line 23:
```go
if _, err := purego.Dlsym(purego.RTLD_DEFAULT, "dlsym"); err != nil {
```

dlfcn_test.go
```diff
@@ -18,6 +18,13 @@ import (
...
+		t.Errorf("failed to dlsym: %s", err)
```
hajimehoshi, line 24: `Dlsym with RTLD_DEFAULT failed: %v`

dlfcn_test.go
```diff
@@ -18,6 +18,12 @@ import (
 	"github.com/ebitengine/purego"
 )
 
+func TestSimpleDlsym(t *testing.T) {
+	if _, err := purego.Dlsym(purego.RTLD_DEFAULT, "dlsym"); err != nil {
+		t.Errorf("Dlsym with RTLD_DEFAULT failed: %s", err)
```
hajimehoshi, line 23: %v, not %s

hajimehoshi (approved):
LGTM

TotallyGamerJet (approved):

PR #138, opened 2023-06-25
objc: add a test to load NSWindow

hajimehoshi:


objc/objc_runtime_darwin_test.go
```diff
@@ -77,8 +77,7 @@ func ExampleID_SendSuper() {
 
 func TestSend(t *testing.T) {
 	// NSNumber comes from Foundation so make sure we have linked to that framework.
-	_, err := purego.Dlopen("Foundation.framework/Foundation", purego.RTLD_GLOBAL)
-	if err != nil {
+	if _, err := purego.Dlopen("Foundation.framework/Foundation", purego.RTLD_GLOBAL); err != nil {
```
TotallyGamerJet, line 80: Should be `purego.RTLD_GLOBAL|purego.RTLD_LAZY` as NOW or LAZY is supposed to be used with GLOBAL/LOCAL

objc/objc_runtime_darwin_test.go
```diff
@@ -90,3 +89,15 @@ func TestSend(t *testing.T) {
 		t.Failed()
 	}
 }
+
+func TestNSWindow(t *testing.T) {
+	if cls := objc.GetClass("NSWindow"); cls != 0 {
+		t.Error("NSWindow must be nil before Cocoa.framework is loaded")
+	}
+	if _, err := purego.Dlopen("Cocoa.framework/Cocoa", purego.RTLD_GLOBAL); err != nil {
```
TotallyGamerJet, line 97: Ditto

TotallyGamerJet:
Something else is linking to AppKit. Maybe Foundation includes the NSWindow symbol?

hajimehoshi:
> Something else is linking to AppKit. Maybe Foundation includes the NSWindow symbol?

```
    objc_runtime_darwin_test.go:95: NSWindow must be nil before Cocoa.framework is loaded
```

Probably yes, but I couldn't reproduce this on my local machine.

TotallyGamerJet:
I too wasn't able to reproduce the issue on an M1 Mac.

PR #141, opened 2023-07-01
internal/fakecgo: generate more

TotallyGamerJet:
Before only the symbols.go and trampolines_stubs.s were generated. however, if more symbols are added symbols_GOOS.go also needs to be updated. Might as well generate that too.


hajimehoshi (approved):
LGTM, thanks!

PR #142, opened 2023-07-03
internal/fakecgo: match Go's implementation

TotallyGamerJet:
This code inlined the macros to avoid copying the header. However, the header was copied into the project for other reasons and this code never got updated to match the Go implementation.


hajimehoshi (approved):
LGTM

hajimehoshi:
> However, the header was copied into the project for other reasons and this code never got updated to match the Go implementation.

And this change fixes the situation by using the existing macro so that this code would get updated, right?

PR #147, opened 2023-08-02
Copy ABI Headers in the right packages

### LINKED ISSUE #144: Bazel build of purego failing
Bazel does not use the go usual build process for an increase in building speed. For this, they use their usual sandbox which makes the current working directory a random temporary directory and not the current package. This makes cross-package `#include` unable to work since the included file is not at the right place at build time. Here is the error:

```
external/com_github_ebitengine_purego/internal/fakecgo/asm_arm64.s:7: 
#include: open /var/folders/fn/j3_08pb94p9_hyz4myff84br0000gq/T/abi/abi_arm64.h: no such file or directory
compilepkg: error running subcommand external/go_sdk/pkg/tool/darwin_arm64/asm: exit status 1
```

The files where it fails are the following:
* [`internal/fakecgo/asm_amd64.h`](https://github.com/ebitengine/purego/blob/b178e7518254895d38a88b4bcdbc02cda4221263/internal/fakecgo/asm_amd64.s#L6)
* [`internal/fakecgo/asm_arm64.h`](https://github.com/ebitengine/purego/blob/b178e7518254895d38a88b4bcdbc02cda4221263/internal/fakecgo/asm_arm64.s#L6)
* [`sys_amd64.h`](https://github.com/ebitengine/purego/blob/b178e7518254895d38a88b4bcdbc02cda4221263/sys_amd64.s#L7C1-L7C1)
* [`sys_arm64.h`](https://github.com/ebitengine/purego/blob/b178e7518254895d38a88b4bcdbc02cda4221263/sys_arm64.s#L7C1-L7C1)

Purego is especially useful in Bazel environments because CGO is disabled by default. I checked on bazel side and this behaviour seems unavoidable. Do we know if cross-package usage of `#include` is really something that is officially supported (in which case bazel is at fault) and not a happy coincidence ? It is only a matter of copying the files in `purego/internal/abi` to `purego/internal/fakecgo` and `purego/`, in which case I can do the PR, or am I missing something ?

eliottness:
### Work done

* Copy `internal/abi/abi_*.h` to `abi_*.h` and `internal/fakecgo/abi_*.h`
* Remove `dummy.go`
* Change README to point correctly to the files
* Create a bazel workspace
* Add an action building purego with bazel

### Goal

Remove cross-package use `#include` in Go assembly files for bazel builds to complete successfully.

### References

Closes #144


sys_arm64.s
```diff
@@ -6,7 +6,7 @@
 #include "textflag.h"
 #include "go_asm.h"
 #include "funcdata.h"
-#include "internal/abi/abi_arm64.h"
+#include "abi_arm64.h"
```
TotallyGamerJet, line 9: Is this even used in this file? Can it be removed?

eliottness: You are right this is not needed here
```suggestion
```

eliottness:
I checked that this fix is indeed working [here](https://github.com/eliottness/bzl-asm-cross-package-include-mre/tree/copy-files-fix)

eliottness:
@TotallyGamerJet This version should be working. I had to add the bazel WORKSPACE and BUILD file at the repo root to test this. Since you have write access to the repo you should be able to remove the action cache entry named "bazel" and leave the other 2. It was a mistake on my side.

TotallyGamerJet:
The code LGTM @hajimehoshi any changes before I merge this?

WORKSPACE
```diff
@@ -0,0 +1,30 @@
+workspace(name = "com_github_ebitengine_purego")
```
hajimehoshi, line 1: What is this file? Is this necesasry on the top directory?

.github/workflows/test.yml
```diff
@@ -82,3 +82,24 @@ jobs:
         if: ${{ !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') }}
         run: |
           go test -race -shuffle=on -v -count=10 ./...
+  bazel:
```
hajimehoshi, line 85: Create bazel.yml

WORKSPACE
eliottness: I can move it somewhere else but it will have to be at the root of the repo for running gazelle and building any go package

eliottness: It has the same role in a bazel build than the go.mod and go.sum files

hajimehoshi: > I can move it somewhere else but it will have to be at the root of the repo for running gazelle and building any go package

Really? So another question is, do third party Go projects have to have this file for Bazel? I don't think so as I have never seen this file.

BUILD.bazel
```diff
@@ -0,0 +1,4 @@
+load("@bazel_gazelle//:def.bzl", "gazelle")
```
hajimehoshi, line 1: I'm also wondering why we need this file.

.gitignore
```diff
@@ -1 +1,6 @@
 *~
+bazel-bin
+bazel-out
+bazel-purego
+bazel-testlogs
+*.bazel
```
hajimehoshi, line 6: I don't think we need these items. Who generates these files?

eliottness:
Hey @hajimehoshi, All these files are here to add a test for building purego via bazel. it's a request from @TotallyGamerJet . I can remove all these files but it will be at the price of the test for the bazel build

TotallyGamerJet:
We talked on the discord and since neither of us are familiar with Bazel we don't want to be responsible for maintaining it. So I think we agreed to sacrifice the test.

eliottness:
Ok I understand. I am removing it in 30min

hajimehoshi:
Another suggestion is to test Bazel with another project using purego (I don't think there exists such a project now, so we don't have to work on this in this PR). PureGo itself doesn't have to be a Bazel project only for testing.

eliottness:
I already worked on something like this [here](https://github.com/ebitengine/purego/pull/147#issuecomment-1662162426)

README.md
```diff
@@ -61,8 +61,9 @@ Purego uses code that originates from the Go runtime. These files are under the
 License that can be found [in the Go Source](https://github.com/golang/go/blob/master/LICENSE).
 This is a list of the copied files:
 
+* `abi_*.h` from package `runtime/cgo`
 * `zcallback_darwin_*.s` from package `runtime`
-* `internal/abi/abi_*.h` from package `runtime/cgo`
+* `internal/fakecgo/abi_*.h` from package `runtime/cgo`
```
hajimehoshi, line 66: So `abi_*.h` and `internal/fakecgo/abi_*.h` are completely the same files, right? If so, can we use symbolic links at the top directory?

eliottness: In other words since the files from other packages are not available a symbolic link fixes nothing

eliottness: Done 👍

hajimehoshi (approved):
LGTM

hajimehoshi:
@TotallyGamerJet Could you merge this if you are OK? (If you are busy, I'm fine to merge this by me)

Issue #150, opened 2023-08-15
what's the corresponding char* type in go


TeavenX:
### base info
platform: darwin/arm64
CGO=1

### question
hello, I have c func like this
```c
ABCXXX const char* Do_Something(void* a, int b);
ABCXXX char* Process_pre_func_output(const char* row_data, int row_data_len);
```
I wonder what's the corresponding `char*` and `const char*` type in go
I've tried `*string` `*[]byte` `*[]rune` `unsafe.Pointer`, however none of them worked

ps: not work means, c func received an invalid argument OR can not set correct value to return type OR wrong mem address error

btw, it's an awesome project!🫡

TeavenX:
found a tortuous solution, takes me hours, :(
```go
var doSomething func(a unsafe.Pointer, b int) unsafe.Pointer
var processPreFuncOutput func(row_data unsafe.Pointer, row_data_len int) unsafe.Pointer

row, rowLen := doSomething(a, b)
tmp := processPreFuncOutput(row, rowLen)

res := (*string)(unsafe.Pointer(&reflect.StringHeader{Data: uintptr(tmp), Len: tmp_len}))
```
well, it's even better if there is a type mapping

TotallyGamerJet:
char* is a *byte bc both are a byte big. If you know the length u can convert that pointer into a slice using unsafe.Slice.

hajimehoshi:
https://github.com/ebitengine/purego/commit/e5b87e976b386838bf71ee85f2eb03f753bae8b9

As this is already documented with an example, there is no further action item for this issue. What do you think? @TotallyGamerJet

TotallyGamerJet:
Agreed. The PR should probably also be closed since it doesn't run.

hajimehoshi:
Closed per e5b87e976b386838bf71ee85f2eb03f753bae8b9

PR #151, opened 2023-08-16
add char ptr convert example

TeavenX:
a simple supplement for #150 https://github.com/ebitengine/purego/issues/150

### DIFF
--- a/examples/type_convert/char_ptr.go
+++ b/examples/type_convert/char_ptr.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"reflect"
 	"unsafe"
 )
 
@@ -22,10 +21,10 @@ func main() {
 	output := Process(&outputLen)
 
 	// manually construct a string struct and set data ptr and length
-	s := *(*string)(unsafe.Pointer(&reflect.StringHeader{Data: uintptr(output), Len: outputLen}))
+	s := unsafe.String((*byte)(output), unsafe.IntegerType(outputLen))
 	println(s)
 
 	// to byte slice
-	bs := *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{Data: uintptr(output), Len: outputLen, Cap: outputLen}))
+	bs := unsafe.Slice((*byte)(output), outputLen)
 	println(bs)
 }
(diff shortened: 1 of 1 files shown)


hajimehoshi:
This example can be compiled but actually doesn't work.

If we want to add such a simple example like this, Example_* to purego_test would be fine.

TotallyGamerJet:
The comment is wrong. You need to use just plain old string. Internally, purego copies the *char into an allocated string object. *string can never happen bc C doesn't know the structure of a Go string. Also, I believe this is incorrectly using StringHeader and SliceHeader. You're not supposed to directly convert to and from those types.

hajimehoshi:
@TotallyGamerJet By the way, doesn't auto-type conversion work for return values? We already have a conversion for arguments: https://github.com/ebitengine/purego/blob/main/examples/libc/main.go

TeavenX:
> 

Use `unsafe.Slice` might be more suitable. (I'm too old fashion😂
However, the uintptr point to a C address, so it should not be garbage collected (if it is pure go case, that's a incorrectly usage). And I found a snippet code of google  https://github.com/google/gousb/blob/844d4b8a718f46eb580a7a84afa00eb15c1b7a9a/libusb.go%23L203 which is similar to my case

hajimehoshi:
Ahh, for return values, C and Go values' lifetimes would matter and automatic conversion would not work well...

hajimehoshi:
https://pkg.go.dev/github.com/ebitengine/purego#RegisterFunc

I failed to find a comment about the copying behavior. Does this happen only when converting a C char* to a Go string?

TotallyGamerJet:
True the copying part is not specified. I was referring to the fact that it does define that a char* can be returned from a function as a string. I should probably specify that when returning it does a copy. As well as define the behavior of arguments as well.

hajimehoshi:
@TotallyGamerJet I would appreciate if you could update the current documentation to explain with the current behavior. Thanks,

hajimehoshi:
https://github.com/ebitengine/purego/commit/e5b87e976b386838bf71ee85f2eb03f753bae8b9

As Purego converts a C string to a Go string automatically, I don't think this example is needed...

hajimehoshi:
Let me close this per e5b87e976b386838bf71ee85f2eb03f753bae8b9

PR #154, opened 2023-08-16
document memory for RegisterFunc

TotallyGamerJet:


func.go
```diff
@@ -35,7 +35,7 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
 // matches the C function. This also holds true for struct types where the padding
 // needs to be ensured to match that of C; RegisterFunc does not verify this.
 //
-// Type Conversions (Go => C)
+// # Type Conversions (Go => C)
```
hajimehoshi, line 38: `Go <=> C` ?

func.go
```diff
@@ -65,6 +65,23 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
...
+// However, there are some special cases. When passing a string as an argument if the string does not end in a null
+// terminated byte (\x00) then the string will be copied into Go memory maintained by purego. The memory is only valid for
+// that specific call. Therefore, if the C code keeps a reference to that string it may become invalid at some
+// undefined time. However, if the string does already contain a null-terminated byte then no copy is done.
+// It is then the responsibility of the caller to ensure the string stays alive as long as it's needed in C memory.
+// This can be done using runtime.KeepAlive. When a C function returns a null-terminated pointer to char a Go string
+// can be used. Purego will allocate a new string in Go memory and copy the data over. This string will
+// be garbage collected whenever Go decides it's no longer referenced. If the pointer to char is not null-terminated or
+// must continue to point to C memory (because it's a buffer for example) then use a pointer to byte and then convert
+// that to a slice using unsafe.Slice.
```
hajimehoshi, line 84: So, if we want to manage the lifetime of the returning value from C, we should not use this automatic conversion, right? Let's add this caveats.

hajimehoshi (commented):
I didn't know (or I just forgot) the rule of the automatic conversion rule, which depends on whether a Go string ends with a null char or not. So the documentation is very helpful!

I am not sure the current rule is the best and am a little worried that this might confuse users, but I don't have a good alternative idea. Let me think this more later.

TotallyGamerJet:
Well if issue #121 was solved than we could remove the automatic allocation since it would be mostly convenient and explicit to convert between the two.

hajimehoshi:
Any updates?

hajimehoshi:
> Well if issue https://github.com/ebitengine/purego/issues/121 was solved than we could remove the automatic allocation since it would be mostly convenient and explicit to convert between the two.

I acknowledged.

func.go
```diff
@@ -65,6 +65,42 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
...
+//
+//	// Let purego convert types
+//	var foo func(s string) string
+//	goString := foo("copied")
+//	// Go will garbage collect this string
+//
+//	// Manually, handle allocations
+//	var foo2 func(b string) *byte
+//	mustFree := foo2("not copied\x00")
+//	defer free(MustFree)
```
hajimehoshi, line 101: `mustFree`

func.go
```diff
@@ -65,6 +65,42 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
...
+// hold onto a reference to Go memory. This is the same as the [Cgo rules].
+//
+// However, there are some special cases. When passing a string as an argument if the string does not end in a null
+// terminated byte (\x00) then the string will be copied into memory maintained by purego. The memory is only valid for
+// that specific call. Therefore, if the C code keeps a reference to that string it may become invalid at some
+// undefined time. However, if the string does already contain a null-terminated byte then no copy is done.
+// It is then the responsibility of the caller to ensure the string stays alive as long as it's needed in C memory.
+// This can be done using runtime.KeepAlive or allocating the string in C memory using malloc. When a C function
+// returns a null-terminated pointer to char a Go string can be used. Purego will allocate a new string in Go memory
+// and copy the data over. This string will be garbage collected whenever Go decides it's no longer referenced.
```
hajimehoshi, line 82: Let's emphasis that purego doesn't care the C's string lifetime, though this might be obvious since purego just copies the data. What do you think?

hajimehoshi (commented):
Thanks!

func.go
TotallyGamerJet: True it should be made explicit that purego doesn't free the C string

func.go
```diff
@@ -65,6 +65,43 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
...
+// This C created string will not be freed by purego. If the pointer to char is not null-terminated or must continue
+// to point to C memory (because it's a buffer for example) then use a pointer to byte and then convert that to a slice
+// using unsafe.Slice. Doing this means that it becomes the responsibility of the caller to choose when the pointer
+// should be freed.
```
hajimehoshi, line 86: That's correct, but there is a case when the callee manages the string lifetime and the caller doesn't have to care. In this case, 'it becomes the responsibility of the caller to choose when the pointer should be freed' seems a little different. 'it becomes the responsibility of the caller to care the lifetime of the pointer' maybe?

hajimehoshi (approved):
LGTM, thanks!

PR #155, opened 2023-08-23
Add FreeBSD support

ldnetgate:
This change adds FreeBSD AMD64 and ARM64 targets.

Includes:
* github actions workflow with FreeBSD test unit:
  The FreeBSD test needs to run in its own space because it's
  executed as a virtualbox VM in macOS. vmactions/freebsd-vm only
  supports a single step block where it exits the VM process once
  the unit completse. All the run commands therefore need to be
  clumped together.

  The setup-go library doesn't support FreeBSD, so different
  versions of Go are manually fetched directly from the Go repo
  instead of using FreeBSD's pkg, which is limited to only a single
  version of Go.
* `environ` and `__progname` variables to enable cross-compilation
   which libc.so relies on. Cross-compile requires
  `-gcflags="github.com/ebitengine/purego/internal/fakecgo=-std"`
   to make the the symbols appear in the binary.

Sponsored by: Rubicon Communications LLC (Netgate)


dlfcn_freebsd.go
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package purego
+
+const (
```
TotallyGamerJet, line 6: Do you have a link that shows these constants in C?

TotallyGamerJet (commented):
Just want to say that I really appreciate this PR. I'll take a closer look tomorrow but it's very exciting to see a new platform supported

dlfcn_freebsd.go
ldnetgate: They are defined in https://github.com/freebsd/freebsd-src/blob/main/lib/csu/common/csu_common.h
The difference with FreeBSD is that the CSU components are not part of libc, and unfortunately libc.so depends on them to exist in the compiled executables.

ldnetgate: oops https://github.com/freebsd/freebsd-src/blob/main/include/dlfcn.h

hajimehoshi:
Nice work!

Can we add tests by https://github.com/vmactions/freebsd-vm ?

ldnetgate:
> Nice work!
> 
> Can we add tests by https://github.com/vmactions/freebsd-vm ?

Not sure if this will satisfy your request (vmactions is new to me and FreeBSD is not a first class citizen of github actions); I've made a custom unit specifically for FreeBSD because of this - `setup-go` and  different golang versions are not available for FreeBSD.

https://github.com/ldnetgate/purego/blob/freebsd-wftest/.github/workflows/test.yml

Test trigger passing: https://github.com/ldnetgate/purego/actions/runs/5951411806

I can merge it into this pull request if it's suitable.
thanks.

hajimehoshi:
> I can merge it into this pull request if it's suitable.

Yes, I have some requests on the YAML file, but it is very nice to have tests for FreeBSD! Thanks,

dlfcn_freebsd.go
TotallyGamerJet: Great! Can you leave this link as a comment in the code that way it's documented

internal/cgo/syscall_cgo_freebsd.go
```diff
@@ -0,0 +1,57 @@
...
+
+package cgo
+
+// this file is placed inside internal/cgo and not package purego
+// because Cgo and assembly files can't be in the same package.
+
+/*
+#include <stdint.h>
+
+int *__error(void);
```
TotallyGamerJet, line 12: #include <errno.h>?

internal/cgo/syscall_cgo_freebsd.go
```diff
@@ -0,0 +1,57 @@
...
+// symbol will then be made available to the linker and linked to inside dlfcn.go
+var (
+	_ = C.dlopen
+	_ = C.dlsym
+	_ = C.dlerror
+	_ = C.dlclose
+)
+
+//go:nosplit
+func Syscall9X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9 uintptr) (r1, r2, err uintptr) {
```
TotallyGamerJet, line 50: If two the changes above can be done can we just make the Linux and FreeBSD versions all one file?

internal/fakecgo/libcgo_freebsd.go
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package fakecgo
+
+type (
+	pthread_cond_t  [48]byte
```
TotallyGamerJet, line 7: Did you confirm these are the right sizes and macro definitions?

TotallyGamerJet (changes requested):

TotallyGamerJet:
As for the missing symbols we can do what Cgo actually does in the runtime. It has a file called [runtime/cgo/freebsd.go](https://github.com/golang/go/blob/master/src/runtime/cgo/freebsd.go). Unfortunately, this pragma "requires" Cgo but the compiler can be told to think that the fakecgo package is Cgo by using `-gcflags="github.com/ebitengine/purego/internal/fakecgo=-std"`. I think this is a better approach since it means the produced binary will run without any issues and doesn't require the preload at each invocation. A comment can also be added above the pragmas so that people trying to cross-compile know what flag and why it's needed.

internal/fakecgo/libcgo_freebsd.go
ldnetgate: PTHREAD_..._INITIALIZERs are NULL in FreeBSD, so I'll change it to that.

internal/cgo/syscall_cgo_freebsd.go
ldnetgate: consolidated to syscall_cgo_unix.go

ldnetgate:
Thanks for the gcflags hint, that was very useful and solves the problem we have here.

The github action passed in my fork: https://github.com/ldnetgate/purego/actions/runs/5957807111/job/16161095503

TotallyGamerJet:
Great! Did you create the freebsd.go file in fakecgo? I don't see it.

ldnetgate:
> Great! Did you create the freebsd.go file in fakecgo? I don't see it.

Apologies, missed it during push. It's in now as go_freebsd.go and test ran: https://github.com/ldnetgate/purego/actions/runs/5958576447/job/16162985604

TotallyGamerJet:
Thank you! We can keep the name just freebsd.go to match what it's called in runtime/cgo. Also can you add it to the README in the list of external code. The README also should have the example updated to show FreeBSD support.

TotallyGamerJet:
Actually maybe the code itself shouldn't be updated since then we have to complicate the example to explain the special flag. I think just update the line that mentions the full example to include FreeBSD

ldnetgate:
that still leaves a question mark as to to the suggestion of
```
Then to run: `CGO_ENABLED=0 go run main.go`
```

which requires the flag even when running on FreeBSD.

ldnetgate:
tweaked the two items you mentioned and tested (https://github.com/ldnetgate/purego/actions/runs/5958993688/job/16164058824)

.github/workflows/test.yml
```diff
@@ -82,3 +88,64 @@ jobs:
         if: ${{ !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') }}
         run: |
           go test -race -shuffle=on -v -count=10 ./...
+
+  freebsd:
```
hajimehoshi, line 92: Can we integrate this with the above?

```
    strategy:
      matrix:
        os: [linux, macos, windows, freebsd]
        go: ['1.18.x', '1.19.x', '1.20.x', '1.21.x']
    runs-on: ${{ matrix.os == 'linux' ? 'ubuntu-latest' : (matrix.os == 'macos' ? 'macos-latest' : ( matrix.os == 'windows' ? 'windows-latest' : 'macos-12')) }}
```

For each actions, you can use `if`:

```
- name: build in FreeBSD
   if: matrix.os == 'freebsd'
   run: |
```

ldnetgate: setup-go doesn't support freebsd, so there's only 1 go matrix possible, whatever is installed via `pkg add -y lang/go` in freebsd (which is 1.20 at the moment).

hajimehoshi: You can make a special set up section with `if`, right?

```
- name: Setup Go
   if: ${{.matrix.os != 'freebsd' }}
   ...
- name: Setup Go (FreeBSD)
   if ${{ matrix.os == 'freebsd'}}
```

> pkg add -y lang/go in freebsd (which is 1.20 at the moment).

Can't we use the official Go binaries for various Go version?

ldnetgate: they mentioned freebsd not supported: https://github.com/actions/setup-go/blob/main/src/system.ts#L5

hajimehoshi: I mean whether we can use https://go.dev/dl/

```
curl https://go.dev/dl/go1.20.7.freebsd-amd64.tar.gz
rm -rf /usr/local/go && tar -C /usr/local -xzf go1.21.0.freebsd-amd64.tar.gz
```

.github/workflows/test.yml
```diff
@@ -8,26 +8,50 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
```
hajimehoshi, line 9: Don't we need macos-12 here?

ldnetgate: I'm going to have to reverify the test, looks like there's a problem with running freebsd because it was running on the "outer" Darwin OS instead of the freebsd VM.

As for the matrix, I had to cluster the freebsd one's (macos-12) under the `-include` list separately. I did try to use your formula `${{ matrix.os == 'linux' ? 'ubuntu-latest' : (matrix.os == 'macos' ? 'macos-latest' : ( matrix.os == 'windows' ? 'windows-latest' : 'macos-12')) }}`, but github complained about the syntax.

hajimehoshi: Hm, I see...

I appreciate you for tackling this issue. Unfortunately, integrating the tests seems harder than expected and I don't want to bother you by fixing test files. If you find it hard to resolve, let's separate the FreeBSD test again like you did first. 

I will revisit this issue later.

ldnetgate: hmm... looks like freebsd-vm/vmactions is a single self-contained test step, ie it brings up the VM, performs `run` and then exits, causing the VM to be destroyed. https://github.com/vmactions/freebsd-vm/blob/main/index.js#L199

This is a problem for integrating FreeBSD with the other os targets...

.github/workflows/test.yml
```diff
@@ -82,3 +88,85 @@ jobs:
...
+            # vmactions/freebsd-vm only supports a single "step" where it
+            # brings down the VM at the end of the step, so all
+            # the commands to run need to be put into this single block.
+
+            echo "Running tests on $(uname -a) at $PWD"
+
+            # verify Go is available
+            go version
+
+            echo "=> go build"
```
hajimehoshi, line 123: My understanding is that it is impossible to have separate names like the above regular tests do, as these steps are executed in a VM. Is this correct?

.github/workflows/test.yml
ldnetgate: I've placed the freebsd matrix in its own block.
Results: https://github.com/ldnetgate/purego/actions/runs/5960355009

.github/workflows/test.yml
ldnetgate: As far as I can tell, it only supports one run unit so single named block: https://github.com/vmactions/freebsd-vm/blob/main/index.js#L163

Adding other `-name` blocks will run on the host OS (macOS).

ldnetgate: yes to your question

hajimehoshi: Acknowledged. Thanks!

.github/workflows/test.yml
```diff
@@ -82,3 +88,85 @@ jobs:
         if: ${{ !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') }}
         run: |
           go test -race -shuffle=on -v -count=10 ./...
+
+  freebsd:
+    strategy:
+      matrix:
+        os: [ macos-12 ]
+        go: ['1.18.10', '1.19.12', '1.20.7', '1.21.0']
+    name: Test with Go ${{ matrix.go }} on ${{ matrix.os }}
```
hajimehoshi, line 97: `Test with Go ${{ matrix.go }} on FreeBSD`

.github/workflows/test.yml
```diff
@@ -82,3 +88,85 @@ jobs:
         if: ${{ !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') }}
         run: |
           go test -race -shuffle=on -v -count=10 ./...
+
+  freebsd:
+    strategy:
+      matrix:
+        os: [ macos-12 ]
```
hajimehoshi, line 95: We don't need `os` for a matrix as there is only one choice.

.github/workflows/test.yml
```diff
@@ -82,3 +88,85 @@ jobs:
...
+        go: ['1.18.10', '1.19.12', '1.20.7', '1.21.0']
+    name: Test with Go ${{ matrix.go }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
```
hajimehoshi, line 98: Hardcode the OS:

```
run-os: macos-12
```

ldnetgate:
Not sure why github shows changes still requested; I can't see any items pending in my display.

Updated the PR description based on the changes applied.

hajimehoshi:
@TotallyGamerJet

TotallyGamerJet (approved):
LGTM

hajimehoshi (approved):
LGTM, thanks!

hajimehoshi:
@ldnetgate I would like to understand `-std` of `-gcflags` more. I think this is undocumented, so could you give me a pointer to a source code (probably in cmd/go)? I appreciate your hard work!

TotallyGamerJet:
I was the one who suggested the flag. Yes it's necessary for that pragma.

The flag is defined [here](https://github.com/golang/go/blob/a5ab4a9471d6745b604a0dc480524229106b0728/src/cmd/compile/internal/base/flag.go#L122).

hajimehoshi:
Thanks!

PR #157, opened 2023-08-27
Implement safe ObjC classes 

### LINKED ISSUE #143: enable `RegisterClass` for a struct with member variables
Now `RegisterClass` is available only for structs without member variables. If a struct had member variables, memory corruption might happen (`examples/objc` uses `RegisterClass` for a struct with a member variable, but this is a known issue).

As @TotallyGamerJet and Hajime discussed on Ebitengine Discord, would it be possible giving up mirroring member variables in the Go world to the Objective-C world? This makes the Objective-C struct size predictable and then memory corruption would no longer happen. Unfortunately, this would be a breaking change since `GetIvar` and `SetIvar` no longer works for Go member variables. However, the current major version is still 0, Hajime is fine to have such a breaking change.

Currently, a struct with members can use `AllocateClassPair`, `AddMethod`, and `AddIvar`, but this is quite painful. `RegisterClass` cares all of them, which is pretty nice.

TotallyGamerJet:
Closes #143


examples/objc/main_darwin.go
```diff
@@ -17,45 +17,42 @@ var (
...
+	class, err := objc.RegisterClass(
+		"BarObject",
+		objc.GetClass("NSObject"),
+		[]*objc.Protocol{
+			objc.GetProtocol("NSDelegateWindow"),
+		},
+		[]objc.FieldDef{
+			{
+				Name:      "bar",
+				Type:      reflect.TypeOf(int(0)),
```
TotallyGamerJet, line 41: Should the Type field just be an interface instead of reflect.Type and then we just obtain the type using reflection? That way there is no dependency on the caller to import reflect.

objc/objc_runtime_darwin.go
```diff
@@ -544,7 +506,7 @@ func (i Ivar) Offset() uintptr {
 }
 
 // Protocol is a type that declares methods that can be implemented by any class.
-type Protocol uintptr
+type Protocol struct{}
```
hajimehoshi, line 509: Is this really safe to change the type? There is no guarantee two pointers to an empty struct are different

https://go.dev/ref/spec#Size_and_alignment_guarantees

> Two distinct zero-size variables may have the same address in memory.

examples/objc/main_darwin.go
hajimehoshi: You mean we might want to write like this?

```go
{
    Name: "bar",
    Type: 0,
    Attribute: objc.ReadWrite
}
```

I slightly prefer `reflect.Type` as this is more explicit.

examples/objc/main_darwin.go
```diff
@@ -17,45 +17,42 @@ var (
...
 
 func main() {
-	class, err := objc.RegisterClass(&barObject{})
+	/*
+		This struct is equivalent to the following Objective-C definition.
+
+		@interface BarObject : NSObject <NSDelegateWindow>
+		@property (readwrite) bar int
+		@end
+	*/
```
hajimehoshi, line 31: Use `//` comments for consistency

objc/objc_runtime_darwin_test.go
```diff
@@ -5,42 +5,61 @@ package objc_test
...
+		},
+		[]objc.FieldDef{
+			{
+				Name:      "bar",
+				Type:      reflect.TypeOf(int(0)),
+				Attribute: objc.ReadWrite,
+			},
+		},
+		[]objc.MethodDef{
+			{sel_init, BarInit},
```
hajimehoshi, line 53: Add field names

examples/objc/main_darwin.go
```diff
@@ -17,45 +17,42 @@ var (
...
+		},
+		[]objc.FieldDef{
+			{
+				Name:      "bar",
+				Type:      reflect.TypeOf(int(0)),
+				Attribute: objc.ReadWrite,
+			},
+		},
+		[]objc.MethodDef{
+			{sel_init, BarInit},
```
hajimehoshi, line 46: Add field names

objc/objc_runtime_darwin.go
TotallyGamerJet: Pointers to this struct are never created on the Go side. They should only be created by calling GetProtocol. In C this struct isn't empty but Go doesn't need to access any of its fields which is why I changed it to an empty struct so that users don't think to dereference it.

hajimehoshi: Don't we have to distinguish Go's Protocol A and Go's Protocol B, for example? If so, why do we need this type?

hajimehoshi: OK so we usually don't use the struct{} value itself but use pointers to structs, and we don't have to compare Protocol values directly. Is this correct?

Another possible proposal would be `[0]func()` to prevent comparing Protcol values directly.

TotallyGamerJet: Yes we use pointers to structs. And yes we don't compare protocols directly there is a function protocol_isEqual that is used for that so perhaps that should be added too? 

I like making it explicit that it's not comparable so I'll use your suggestion

examples/objc/main_darwin.go
```diff
@@ -17,45 +17,42 @@ var (
...
+		},
+		[]objc.FieldDef{
+			{
+				Name:      "bar",
+				Type:      reflect.TypeOf(int(0)),
+				Attribute: objc.ReadWrite,
+			},
+		},
+		[]objc.MethodDef{
+			{Cmd: sel_init, Fn: BarInit},
```
hajimehoshi, line 46: I prefer indentations:

```go
{
    Cmd: sel_init,
    Fn: BarInit,
},
```

examples/objc/main_darwin.go
```diff
@@ -17,45 +17,42 @@ var (
...
-		return sel_init
-	case "SetBar":
-		return sel_setBar
-	case "Bar":
-		return sel_bar
-	default:
-		return 0
-	}
+func BarInit(id objc.ID, _cmd objc.SEL) objc.ID {
+	return id.SendSuper(_cmd)
```
hajimehoshi, line 21: Why `_cmd` rather than `cmd`?

objc/objc_runtime_darwin.go
```diff
@@ -148,203 +167,154 @@ func AllocateClassPair(super Class, name string, extraBytes uintptr) Class {
...
-						}
-						break outer
-					}
-				}
-				return 0, fmt.Errorf("objc: expected '>': %w", TagFormatError)
-			}
-		}
-	}
-	class := objc_allocateClassPair(GetClass(split[1]), split[0], 0)
+type FieldDef struct {
```
hajimehoshi, line 197: Let's add an explanation comment.

objc/objc_runtime_darwin.go
```diff
@@ -148,203 +167,154 @@ func AllocateClassPair(super Class, name string, extraBytes uintptr) Class {
...
+					}
+				case reflect.Pointer, reflect.UnsafePointer:
+					value = ID(args[2].Pointer())
+				default:
+					panic(fmt.Errorf("objc: unsupported kind %s", args[2].Kind()))
+				}
+				object_setIvar(args[0].Interface().(ID), instanceVariable, value)
+				return nil
+			}).Interface()
+			class.AddMethod(RegisterName("set"+strings.Title(ivar.Name)+":\x00"), NewIMP(val), encoding)
```
hajimehoshi, line 289: `strings.Title` is deprecated. https://pkg.go.dev/strings#Title . Also, we don't have to consider the case when `name` includes space chars.

What about making our own function to make a string a setter name?

```go
func toSetterName(name string) string {
    // This assumes name consists of ASCII characters.
    // I don't think we have to consider non-ASCII characters.
    return "set" + strings.ToUpper(string(name[0])) + name[1:]
}
```

objc/objc_runtime_darwin_test.go
```diff
@@ -5,42 +5,61 @@ package objc_test
...
+		},
+		[]objc.FieldDef{
+			{
+				Name:      "bar",
+				Type:      reflect.TypeOf(int(0)),
+				Attribute: objc.ReadWrite,
+			},
+		},
+		[]objc.MethodDef{
+			{Cmd: sel_init, Fn: BarInit},
```
hajimehoshi, line 53: ditto

objc/objc_runtime_darwin.go
hajimehoshi: Also, for a boolean member, `is` should be used as a prefix.

examples/objc/main_darwin.go
TotallyGamerJet: It was just a holdover from the previous code. We had decided to use cmd from now on.

objc/objc_runtime_darwin.go
```diff
@@ -148,203 +168,167 @@ func AllocateClassPair(super Class, name string, extraBytes uintptr) Class {
...
+					value = ID(args[2].Pointer())
+				default:
+					panic(fmt.Errorf("objc: unsupported kind %s", args[2].Kind()))
+				}
+				object_setIvar(args[0].Interface().(ID), instanceVariable, value)
+				return nil
+			}).Interface()
+			// this code only works for ascii but that shouldn't be a problem
+			if ivar.Type.Kind() == reflect.Bool && strings.HasPrefix(ivar.Name, "is") {
+				ivar.Name = ivar.Name[2:]
```
hajimehoshi, line 296: This doesn't work when the property name is for example `island`. Maybe detecting `'A' <= ivar.Name[2] && ivar.Name[2] <= 'Z'` is enough. (I am not sure when the name is `is12345` 🤔 )

TotallyGamerJet: how about only checking when the third letter is lower case then do nothing

hajimehoshi: Maybe that works.

In the first place, should `ivar.Name` include `is`? I am not sure.

hajimehoshi: Assuming `ivar.Name` does NOT include `is`, just adding `is` when making a getter name might make sense (though I am not sure which is more conventional). 

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html (editable vs isEditable)

TotallyGamerJet: So we say that Ivar.Name should not start with is and then just append is when the type is bool?

hajimehoshi: IMO yes. What do you think?

TotallyGamerJet: I think that works

objc/objc_runtime_darwin.go
```diff
@@ -148,203 +168,169 @@ func AllocateClassPair(super Class, name string, extraBytes uintptr) Class {
...
+// The name also cannot contain any spaces.
+// The type is the Go equivalent type of the Ivar.
+// Attribute determines if a getter and or setter method is generated for this field.
+type FieldDef struct {
+	Name      string
+	Type      reflect.Type
+	Attribute IvarAttrib
+}
+
+// The RegisterClass function takes the name of the class to create, the superclass, a list of protocols this class
```
hajimehoshi, line 210: Remove `The`?

`RegisterClass takes...`

objc/objc_runtime_darwin.go
```diff
@@ -544,19 +530,23 @@ func (i Ivar) Offset() uintptr {
...
 // Protocol is a type that declares methods that can be implemented by any class.
-type Protocol uintptr
+type Protocol [0]func()
 
 // GetProtocol returns the protocol for the given name or nil if there is no protocol by that name.
 func GetProtocol(name string) *Protocol {
 	return objc_getProtocol(name)
 }
 
+func (p *Protocol) Equals(p2 *Protocol) bool {
```
hajimehoshi, line 540: Add a comment.

objc/objc_runtime_darwin.go
```diff
@@ -148,203 +168,169 @@ func AllocateClassPair(super Class, name string, extraBytes uintptr) Class {
...
 		}
 	}
 	// Add Ivars
 	// Start at 1 because we skip the class object which is first
-	for i := 1; i < strct.NumField(); i++ {
-		f := strct.Field(i)
-		if f.Name == "_" {
-			continue
+	for _, ivar := range ivars {
+		if strings.Contains(ivar.Name, " ") {
```
hajimehoshi, line 248: Let's check the name with a regexp to avoid future troubles like this regexp expression:

```
[a-z][a-zA-Z0-9]*
```

As far as I know, the first character is always a small cap.

FYI, even if an ivar name starts with an abbreviation, the setter name is always like setFooBar, not setFOOBar. For example, Metal has `setRgbBlendOperation` for `rgbBlendOperation`, instead of `setRGBBlendOperation`. So the current implemention for setters seeems fine to me.

hajimehoshi (commented):
Overall good!

objc/objc_runtime_darwin.go
```diff
@@ -148,203 +168,171 @@ func AllocateClassPair(super Class, name string, extraBytes uintptr) Class {
...
 	}
 	// Add Ivars
-	// Start at 1 because we skip the class object which is first
-	for i := 1; i < strct.NumField(); i++ {
-		f := strct.Field(i)
-		if f.Name == "_" {
-			continue
+	for _, ivar := range ivars {
+		if !ivarRegex.MatchString(ivar.Name) {
+			return 0, fmt.Errorf("objc: Ivar cannot have space: '%s'", ivar.Name)
```
hajimehoshi, line 251: update this error message.

objc/objc_runtime_darwin.go
```diff
@@ -148,203 +168,171 @@ func AllocateClassPair(super Class, name string, extraBytes uintptr) Class {
...
+// The fields are still accessible using objc.GetIvar and objc.SetIvar regardless of the value of IvarAttrib.
+//
+// Take for example this Objective-C code:
+//
+//	@property (readwrite) float value;
+//
+// In Go, creates functions with these signatures:
+//
+//	func value() float32
+//	func setValue(float32)
```
hajimehoshi, line 188: This seems a little confusing. Such these functions are not generated, right?

examples/objc/main_darwin.go
```diff
@@ -17,45 +17,45 @@ var (
...
 
 func main() {
-	class, err := objc.RegisterClass(&barObject{})
+	//
+	// This struct is equivalent to the following Objective-C definition.
+	//
+	// @interface BarObject : NSObject <NSDelegateWindow>
+	// @property (readwrite) bar int
+	// @end
+	//
```
hajimehoshi, line 31: Remove `//` with empty lines

TotallyGamerJet: Even the one between the sentence and the objc code?

hajimehoshi: Sorry for my ambiguous comment. Let's keep the line between the sentence and the code.

objc/objc_runtime_darwin.go
TotallyGamerJet: No they aren't exactly created but are equivalently created on the ObjC side

hajimehoshi: Thanks. IIUC, a purego user can access the ivar via `Send` + selectors (`value` and `setValue:`), so I am not sure these Go function sigunatures help users. What about explaining a usage of `Send` and selectors?

hajimehoshi (approved):
LGTM, thanks!

Issue #124, opened 2023-04-07
Linux NewCallback


tmclane:
It would seem that NewCallback is not implemented on Linux.
Is this already in development?

TotallyGamerJet:
Not really at the moment. Most of the code is already there but isn't working yet. You can look at my linuxcb branch for a starting point. If you get it working create a PR

jwijenbergh:
> 

Could you expand upon what is not working exactly on that branch? All callback tests work when running `CGO_ENABLED=0 go test`.

I've also tested it with some gtk code and it also works there. Anything I'm missing? I am on amd64

TotallyGamerJet:
Well the GitHub tests failed. They use these commands:
```
env CGO_ENABLED=0 go test -shuffle=on -v -count=10 ./...
env CGO_ENABLED=1 go test -shuffle=on -v -count=10 ./...
```
They pass sometimes and fail other times. Exactly why I'm not 100% sure but might be related to threading

tmclane:
Ok I'll look into it and see if can figure out why it is crashing.

tmclane:
You know the code actually says that `float32` and `float64` are not supported as arguments when using `CGO_ENABLED=1` on linux.  Is that the error you are getting?   I am getting other errors that are definitely related to timing BUT they don't happen when `CGO_ENABLED=0` (which is the use-case I think most people would want on Linux).

TotallyGamerJet:
I don't have amd64 but I here's what I get on linux arm64. It fails both with and without Cgo but only some times.
<details>

<summary> terminal </summary>

```
$ CGO_ENABLED=0 go test -shuffle=on -v -count=10
-test.shuffle 1681241131463574146
=== RUN   Test_qsort
    dlfcn_test.go:83: got 3674159897003727796 wanted 2 at 0
    dlfcn_test.go:83: got 3674159897003727796 wanted 25 at 1
    dlfcn_test.go:83: got 3674159897003727796 wanted 56 at 2
    dlfcn_test.go:83: got 3674159897003727796 wanted 88 at 3
    dlfcn_test.go:83: got 88 wanted 100 at 4
--- FAIL: Test_qsort (0.00s)
=== RUN   TestNewCallbackFloat32AndFloat64
--- PASS: TestNewCallbackFloat32AndFloat64 (0.00s)
=== RUN   TestRegisterFunc
Calling C from from Go without Cgo!
--- PASS: TestRegisterFunc (0.00s)
[...]

PR #165, opened 2023-09-22
fix unintended R27 clobber on arm64

### LINKED ISSUE #162: Oto crashes with iOS 17 / Xcode 15 crashes (memory corruption?)
https://github.com/hajimehoshi/ebiten/issues/2773

Whenever an iOS application starts to play sound, the application crashes with this message:

```
objc[708]: autorelease pool page 0x104bf8000 corrupted
  magic     0x04bfc7c0 0x00000001 0x04bfc800 0x00000001
  should be 0xa1a1a1a1 0x4f545541 0x454c4552 0x21455341
  pthread   0x104bfc880
  should be 0x16ba9f000
```

* Ebitengine 2.7 (https://github.com/hajimehoshi/ebiten/commit/a3ff5f283ba393488cb6dc675155862d4e55799b) w/ Oto v3.2.0-alpha crashes
* Ebitengine 2.5 (https://github.com/hajimehoshi/ebiten/commit/7b8bdd0ee9f968b84e75d288ba52ef3ea3ae6982) w/ Oto v2.4.1 crashse
* Ebitengine 2.5 (the same as above) w/ Oto w/o purego (https://github.com/ebitengine/oto/commit/fa48350aba22b2cb4ee2cbe9a5930c6b68ed894e) did NOT crash

TotallyGamerJet:
For some reason the Go assembler uses R27 as a scratch register when loading in a variable's address. Since callbackasm1 is called from C and the arm64 ABI states that R27 is callee saved it was unexpectedly clobbering the value. This commit now saves R27 and restores it at the end of the function.

I also cleaned up some of the other parts of this function. I store the floats in pairs which saved 4 instructions. I also use the zero register to put zero in callbackArgs.result directly

Updates #124
Closes #162

### DIFF
--- a/sys_unix_arm64.s
+++ b/sys_unix_arm64.s
@@ -17,31 +17,37 @@ TEXT callbackasm1(SB), NOSPLIT|NOFRAME, $0
 	// Save callback register arguments R0-R7 and F0-F7.
 	// We do this at the top of the frame so they're contiguous with stack arguments.
 	SUB   $(16*8), RSP, R14
-	FMOVD F0, (0*8)(R14)
-	FMOVD F1, (1*8)(R14)
-	FMOVD F2, (2*8)(R14)
-	FMOVD F3, (3*8)(R14)
-	FMOVD F4, (4*8)(R14)
-	FMOVD F5, (5*8)(R14)
-	FMOVD F6, (6*8)(R14)
-	FMOVD F7, (7*8)(R14)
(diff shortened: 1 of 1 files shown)


TotallyGamerJet:
For future reference, Go's internal ABI is described [here](https://github.com/golang/go/blob/master/src/cmd/compile/abi-internal.md#arm64-architecture)

hajimehoshi (approved):
LGTM, thanks!

PR #166, opened 2023-09-23
.github/workflows: add tests for Windows 386

hajimehoshi:


TotallyGamerJet:
We tried this before and had issues. See commit #131

hajimehoshi:
Oh ya, that's true...

PR #168, opened 2023-09-23
Document supported platforms and status of project

### LINKED ISSUE #149: document package maturity
This project seems very cool!

However it's hard to assess the relative maturity of this package; which features it has or doesn't have, what are the largest caveats or roadblocks that one might encounter. 

It would be super great to have some high-level documentation about what consumers of this package can expect so that people can assess how suitable/viable this package is. 

Thanks! :)

TotallyGamerJet:
Closes #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,12 @@ born to bring that same vision to the other platforms supported by Ebitengine.
 
 ## Supported Platforms
 
-- **Darwin**: amd64, arm64
 - **FreeBSD**: amd64, arm64
 - **Linux**: amd64, arm64
-- **Windows**: amd64, arm64
+- **macOS**: amd64, arm64
+- **Windows**: 386*, amd64, arm*, arm64
+
+`*` These architectures only support SyscallN and NewCallback
 
 ## Example
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ born to bring that same vision to the other platforms supported by Ebitengine.
 
 - **FreeBSD**: amd64, arm64
 - **Linux**: amd64, arm64
-- **macOS**: amd64, arm64
+- **macOS / iOS**: amd64, arm64
 - **Windows**: 386*, amd64, arm*, arm64
 
 `*` These architectures only support SyscallN and NewCallback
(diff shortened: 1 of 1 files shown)


hajimehoshi (approved):
LGTM, thanks!

README.md
```diff
@@ -17,6 +21,13 @@ born to bring that same vision to the other platforms supported by Ebitengine.
 - **Dynamic Linking**: Load symbols at runtime and use it as a plugin system.
 - **Foreign Function Interface**: Call into other languages that are compiled into shared objects.
 
+## Supported Platforms
+
+- **Darwin**: amd64, arm64
```
hajimehoshi, line 26: Maybe macOS is better? (If we want to use GOOS names, let's unify them with lower cases)

README.md
```diff
@@ -17,6 +21,15 @@ born to bring that same vision to the other platforms supported by Ebitengine.
 - **Dynamic Linking**: Load symbols at runtime and use it as a plugin system.
 - **Foreign Function Interface**: Call into other languages that are compiled into shared objects.
 
+## Supported Platforms
+
+- **FreeBSD**: amd64, arm64
+- **Linux**: amd64, arm64
+- **macOS**: amd64, arm64
```
hajimehoshi, line 28: Sorry I forgot about iOS.

`macOS/iOS`?

hajimehoshi (commented):
Offtopic: has anyone tested purego with Android?

README.md
TotallyGamerJet: oh true.... How did I forget that!

TotallyGamerJet:
> Offtopic: has anyone tested purego with Android?

Does the Android version of Ebitengine use any purego?

hajimehoshi:
> Does the Android version of Ebitengine use any purego?

Not yet. Calling JNI without Cgo might be pretty nice.

TotallyGamerJet:
> Not yet. Calling JNI without Cgo might be pretty nice.

But would it suffer the same issue as iOS where you kind of have to use Cgo?

hajimehoshi:
> But would it suffer the same issue as iOS where you kind of have to use Cgo?

Yes Cgo is inevitable, but we can reduce them, like what we did for Oto.

hajimehoshi (approved):
LGTM!

PR #170, opened 2023-09-29
.github/workflows: reenable tests on branches

hajimehoshi:
Now tests run only on the main branch. This is not good as stable branches are not tested.


TotallyGamerJet (approved):
LGTM

PR #171, opened 2023-10-02
add test for mix of floats and int arguments in callback

TotallyGamerJet:
There wasn't a test for when integers and floats are interleaved in callbacks

### DIFF
--- a/callback_test.go
+++ b/callback_test.go
@@ -80,7 +80,7 @@ func TestNewCallbackFloat64(t *testing.T) {
 func TestNewCallbackFloat64AndIntMix(t *testing.T) {
 	// This tests interleaving float and integer arguments to NewCallback
 	const (
-		expectCbTotal = 43.74
+		expectCbTotal = 54.75
 	)
 	var cbTotal float64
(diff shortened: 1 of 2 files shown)


hajimehoshi (approved):
LGTM

callback_test.go
```diff
@@ -76,6 +77,24 @@ func TestNewCallbackFloat64(t *testing.T) {
...
+	const (
+		expectCbTotal = 43.74
+	)
+	var cbTotal float64
+	imp := purego.NewCallback(func(a1, a2 float64, a3, a4, a5 int, a6, a7, a8 float64, a9 int) {
+		cbTotal = a1 + a2 + float64(a3) + float64(a4) + float64(a5) + a6 + a7 + a8 + float64(a9)
+	})
+	var fn func(a1, a2 float64, a3, a4, a5 int, a6, a7, a8 float64, a9 int)
+	purego.RegisterFunc(&fn, imp)
+	fn(1.5, 3.7, 5, 9, 2, 8.9, 3.4, 4.24, 6)
```
hajimehoshi, line 91: Optional: I'm a little worried about precision. What about using .5 and .25?

PR #172, opened 2023-10-08
WIP: PoC of using a bit of generics and less reflection

Zyko0:


MatejMagat305:
is this plan or only example and hypotetical idea?

PR #175, opened 2023-11-17
Add struct argument support

TotallyGamerJet:
Add struct argument support to purego on arm64 and amd64 macOS.


progrium:
I didn't see any tests for struct return values, is that going to be in this PR?

TotallyGamerJet:
The focus of this PR passing structs.

func.go
```diff
@@ -133,6 +133,11 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 				} else {
 					stack++
 				}
+			case reflect.Struct:
+				if runtime.GOOS != "darwin" || (runtime.GOARCH != "amd64" && runtime.GOARCH != "arm64") {
+					panic("purego: struct arguments are only supported on macOS amd64 & arm64")
+				}
+				// TODO: implement
```
TotallyGamerJet, line 140: @hajimehoshi What are your thoughts about having functions with struct arguments panicking only when they are called? There is already ~250 LOC solely for placing struct arguments. This supports _most_ cases but not even all so more code would be necessary to support them. (Although IDK if that needs to be done in this PR). In order to do the same "does it fit" when creating a function like other arguments all the struct code would basically have to be duplicated and it is tricky code. I'm suggesting we loosen the check for structs to only panic when the C function is about to be called.

func.go
```diff
@@ -52,7 +52,7 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
 //	int64 <=> int64_t
 //	float32 <=> float (WIP)
 //	float64 <=> double (WIP)
```
hajimehoshi, line 54: Are floats still WIP?

func.go
hajimehoshi: Did you mean that it is pretty troublesome to check whether a struct is available for purego or not? Can we have a function to check this by refactoring the below code...?

TotallyGamerJet: That might work. I'll think about it

func.go
```diff
@@ -50,9 +50,9 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
 //	int16 <=> int16_t
 //	int32 <=> int32_t
 //	int64 <=> int64_t
-//	float32 <=> float (WIP)
-//	float64 <=> double (WIP)
-//	struct <=> struct (WIP)
+//	float32 <=> float
+//	float64 <=> double
```
hajimehoshi, line 54: Can we split this PR into supporting floats and supporitng structs?

structtest/struct.c
```diff
@@ -0,0 +1,306 @@
...
+struct BoolFloat {
+    _Bool b;
+    float f;
+};
+
+float BoolFloat(struct BoolFloat s) {
+    if (s.b)
+        return s.f;
+    return -s.f;
+}
```
hajimehoshi, line 306: Add a new line.

func.go
```diff
@@ -292,6 +317,343 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+			return true
+		default:
+			return false
+		}
+	default:
+		return false
+	}
+}
+
+func isHVA(t reflect.Type) bool {
```
hajimehoshi, line 629: Add a comment to explain what this is.

func.go
TotallyGamerJet: Floats are already supported in purego. I just changed the comment in this PR. Do you want the comment change in a separate PR? This PR only adds support for struct arguments not returns

hajimehoshi: > Do you want the comment change in a separate PR?

Yes. This would be less confusing.

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+			return true
+		default:
+			return false
+		}
+	default:
+		return false
+	}
+}
+
+// a HVA is a Homogeneous Aggregate with a Fundamental Data Type that is a Short-Vector type
```
hajimehoshi, line 632: `isHVA reports`...

func.go
```diff
@@ -133,6 +139,23 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 				} else {
 					stack++
 				}
+			case reflect.Struct:
+				if runtime.GOOS != "darwin" || (runtime.GOARCH != "amd64" && runtime.GOARCH != "arm64") {
+					panic("purego: struct arguments are only supported on macOS amd64 & arm64")
```
hajimehoshi, line 144: s/macOS/darwin/

func.go
```diff
@@ -52,7 +52,7 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
 //	int64 <=> int64_t
 //	float32 <=> float (WIP)
 //	float64 <=> double (WIP)
-//	struct <=> struct (WIP)
+//	struct <=> struct (WIP - macOS only)
```
hajimehoshi, line 55: Darwin only? (iOS is also supported, right?)

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 	fn.Set(v)
 }
 
+func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFloat, addStack func(uintptr), keepAlive []interface{}) []interface{} {
+	if v.Type().Size() == 0 {
+		return keepAlive
+	}
+	if runtime.GOARCH == "arm64" {
+		// https://student.cs.uwaterloo.ca/~cs452/docs/rpi4b/aapcs64.pdf
```
hajimehoshi, line 325: Isn't https://github.com/ARM-software/abi-aa a more relaiable and stable URL for the PDFs?

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 	fn.Set(v)
 }
 
+func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFloat, addStack func(uintptr), keepAlive []interface{}) []interface{} {
+	if v.Type().Size() == 0 {
+		return keepAlive
+	}
+	if runtime.GOARCH == "arm64" {
```
hajimehoshi, line 324: Use switch-case.

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+			// each element goes onto the stack
+			if hfa && *numFloats+v.NumField() > numOfFloats {
+				*numFloats = numOfFloats
+			} else if hva && *numInts+v.NumField() > numOfIntegerRegisters() {
+				*numInts = numOfIntegerRegisters()
+			}
+
+			var val uint64
+			var shift byte
+			flushed := false
```
hajimehoshi, line 342: `var flushed bool`

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+		var (
+			placedOnStack  = v.Type().Size() > 8*8 // if greater than 64 bytes place on stack
+			savedNumFloats = *numFloats
+			savedNumInts   = *numInts
+			savedNumStack  = *numStack
+		)
+
+		var val uint64
+		var shift byte // # of bits to shift
+		flushed := false
```
hajimehoshi, line 464: `var flushed bool`

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+		return keepAlive
+	}
+	panic("purego: struct has field that can't be allocated")
+}
+
+func roundUpTo8(val uintptr) uintptr {
+	return (val + 7) &^ 7
+}
+
+// isHVA reports a Homogeneous Floating-point Aggregate (HFA) which is a Fundamental Data Type that is a
```
hajimehoshi, line 601: isHFA

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+// Floating-Point type and at most four uniquely addressable members (5.9.5.1).
```
hajimehoshi, line 602: What is `5.9.5.1`?

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+		default:
+			return false
+		}
+	default:
+		return false
+	}
+}
+
+// isHVA reports a Homogeneous Aggregate with a Fundamental Data Type that is a Short-Vector type
+// and at most four uniquely addressable members (5.9.5.2).
```
hajimehoshi, line 633: What is `5.9.5.2`?

func.go
```diff
@@ -292,6 +317,351 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+			// Copy to heap and place the pointer in register
+			ptrStruct := reflect.New(v.Type())
+			ptrStruct.Elem().Set(v)
+			ptr := ptrStruct.Elem().Addr().UnsafePointer()
+			keepAlive = append(keepAlive, ptr)
+			addInt(uintptr(ptr))
+		}
+		return keepAlive // the struct was allocated so don't panic
+	} else if runtime.GOARCH == "amd64" {
+		// https://www.uclibc.org/docs/psABI-x86_64.pdf
```
hajimehoshi, line 445: Optional: is there an official PDF for this specification?

func.go
TotallyGamerJet: It the section in the spec where information about the type is

hajimehoshi: Let's add a link to the spec. Only from this comment, it is hard which PDF is referred.

struct_test.go
```diff
@@ -0,0 +1,415 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 The Ebitengine Authors
```
hajimehoshi, line 2: s/2023/2024/ (Happy New Year!)

structtest/struct.c
```diff
@@ -0,0 +1,306 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 The Ebitengine Authors
```
hajimehoshi, line 2: ditto

func.go
```diff
@@ -81,6 +81,12 @@ func RegisterLibFunc(fptr interface{}, handle uintptr, name string) {
 // using unsafe.Slice. Doing this means that it becomes the responsibility of the caller to care about the lifetime
 // of the pointer
 //
+// # Structs
+//
+// Purego can handle the most common structs that have fields of builtin types like int8, uint16, float32, etc. However,
+// it does not _currently_ support aligning fields properly. It is therefore the responsibility of the caller to ensure
```
hajimehoshi, line 87: The situation seems the same as Windows: adjusting the layout is a caller's responsibility. Do you think we should fix this?

func.go
```diff
@@ -133,6 +139,23 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+				addFloat := func(u uintptr) {
+					floats++
+				}
+				addStack := func(u uintptr) {
+					stack++
+				}
+				if arg.Size() == 0 {
+					continue
+				}
+				_ = addStruct(reflect.New(arg).Elem(), &ints, &floats, &stack, addInt, addFloat, addStack, nil)
```
hajimehoshi, line 158: What about returning the counts of ints, floats and stacks instead of passing closures?

structtest/struct.c
```diff
@@ -0,0 +1,306 @@
+// SPDX-License-Identifier: Apache-2.0
```
hajimehoshi, line 1: If this file is only for testing, rename this to `struct_test.c`

func.go
TotallyGamerJet: Perhaps, however it is not possible for purego to ensure it's doing the right thing since it can't see any C headers to know what the signature is.

func.go
TotallyGamerJet: Closures are needed for actually placing values into the registers down below. This is because addInt may place it on the stack if all the registers are full. It may be possible to return the ints, floats, stack instead of using pointers but is that as clear?

TotallyGamerJet: We could but would we return just the new ints, floats, stack variables or keep them pointers in the struct?

hajimehoshi: I think we would keep them pointers in the struct.

TotallyGamerJet: Do we gain anything then by putting that state together?

structtest/struct.c
TotallyGamerJet: It is only for testing. The other C files that are used for testing don't have `_test.c` so those should be changed in a different PR.

func.go
hajimehoshi: Readability would be improved (at least to me). What do you think?

structtest/struct.c
hajimehoshi: Sure, thanks

func.go
hajimehoshi: Isn't this `argumentState` as this is used not only for structs but other argument types?

TotallyGamerJet: Well it's only passed around for structs right?

hajimehoshi: My intention is replacing

```go
		var ints int
		var floats int
		var stack int

// ...

				addInt := func(u uintptr) {
					ints++
				}
				addFloat := func(u uintptr) {
					floats++
				}
				addStack := func(u uintptr) {
					stack++
				}
```

with

```go
type argumentState struct {
    ints int
    floats int
    stack int
}

func (a *argumentState) addInt(u uintptr) {
    a.ints++
}

func (a *argumentState) addFloat(u uintptr) {
    a.floats++
}

func (a *argumentState) addStack(u uintptr) {
    a.stack++
}

// We might be able to add the 'save' behavior as a method.

// ...

var state argumentState
```

and passing a pointer to `state` instead of 6 arguments. Would this make sense?

hajimehoshi: On second thought, I found that addInt/addFloat/addStack's implementation differ, so if we would do, we would have to use an interface like this?

```go
type argumentState interface {
     addInt(u uintptr)
     addFloat(u uintptr)
     addStack(u uintptr)
}
```

As this seems a little big refactoring, it's OK not to do this in this PR. Let's leave the current arguments as they are.

hajimehoshi, line 158: Swap the order like this:

```go
				if arg.Size() == 0 {
					continue
				}
				addInt := func(u uintptr) {
					ints++
				}
				addFloat := func(u uintptr) {
					floats++
				}
				addStack := func(u uintptr) {
					stack++
				}
```

struct_amd64.go
```diff
@@ -0,0 +1,163 @@
...
+				addFloat(uintptr(math.Float64bits(f.Float())))
+				class = NO_CLASS
+			case reflect.Array:
+				place(f)
+			default:
+				panic("purego: unsupported kind " + f.Kind().String())
+			}
+		}
+	}
+	place(v)
```
hajimehoshi, line 125: Should this be called even if `placedOnStack` is true?

TotallyGamerJet: It doesn't need to be

TotallyGamerJet: I'm not 100% sold on how I split the function since there is still a callback in one of them but I think it's more understandable now.

struct_amd64.go
```diff
@@ -0,0 +1,172 @@
...
+		// reset any values placed in registers
+		*numFloats = savedNumFloats
+		*numInts = savedNumInts
+		*numStack = savedNumStack
+		placeStack(v, addStack)
+	}
+	return keepAlive
+}
+
+func attemptPlaceRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) (placeOnStack bool) {
```
hajimehoshi, line 47: * `tryPlaceRegisters` seems a better name.
* Returning `ok bool` seems more natural than `placeOnStack`. `ok` would be true when values are put on registers.

struct_arm64.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		place(v)
+		if !flushed {
+			if class == FLOAT {
+				addFloat(uintptr(val))
+			} else {
+				addInt(uintptr(val))
+			}
+		}
+	} else {
```
hajimehoshi, line 124: This can also be split into the register version and the stack version for readability.

hajimehoshi (commented):
Almost LGTM!

structtest/struct_test.c
```diff
@@ -0,0 +1,306 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2024 The Ebitengine Authors
+
+// Empty is empty
+struct Empty {};
+
+//NoStruct tests that an empty struct doesn't cause issues
```
hajimehoshi, line 7: `// NoStruct...`

hajimehoshi (approved):
LGTM, thanks!

TotallyGamerJet:
@hajimehoshi Can you review this again since I fixed a bug

struct_arm64.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+
+		tryPlaceRegisters(v, addFloat, addInt)
+	} else {
+		keepAlive = placeStack(v, keepAlive, addInt)
+	}
+	return keepAlive // the struct was allocated so don't panic
+}
+
+func tryPlaceRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
```
hajimehoshi, line 39: Is this 'try'? Can this function fail?

structtest/struct_test.c
```diff
@@ -0,0 +1,319 @@
...
+      struct { double width, height; } size;
+};
+
+unsigned long InitWithContentRect(int *win, struct Content c, int style, int backing, _Bool flag) {
+  if (win == 0)
+      return 0xBAD;
+  if (!flag)
+      return 0xF1A6; // FLAG
+  return (unsigned long)(c.point.x + c.point.y + c.size.width + c.size.height) / (style - backing);
+}
```
hajimehoshi, line 319: Add a new line.

struct_arm64.go
TotallyGamerJet: It can't fail like amd64. We know if placing in registers is the correct thing to do before calling it.

hajimehoshi: Then let's remove 'try'.

hajimehoshi (approved):
LGTM

PR #163, opened 2023-09-22
Support Linux callbacks

jwijenbergh:
Fixes #124

Most of the work was already there, it's just some re-ordering. The main blocker was the ARM64 crash with callbacks which was nicely fixed by @TotallyGamerJet in https://github.com/ebitengine/purego/issues/124#issuecomment-1730540730 and merged in #165 

I had to split the callback tests into separate tests for floats as purego only supports float arguments with `CGO_ENABLED=0` currently.

Tested on my amd64 laptop, and an arm64 device. Both with `CGO_ENABLED=0` and `CGO_ENABLED=1`. The test suite passes and my GTK program leveraging https://github.com/jwijenbergh/puregotk/ also works on both architectures.

Edit: Floats are now supported with CGO_ENABLED=1 so this is basically just changing some build tags

PR #177, opened 2023-11-18
purego: fix float returns

### LINKED ISSUE #176: Float return Values dont work
This one always returns zero:
```golang
var getFrameTime func() float32
purego.RegisterLibFunc(&getFrameTime, raylibDll, "GetFrameTime")
delta := getFrameTime()
// delta is 0
```

When using the syscall package like this, it works:
```golang
getFrameTime, _ := syscall.GetProcAddress(handle, "GetFrameTime")
_, ret, _ := syscall.SyscallN(getFrameTime)
delta := *(*float32)(unsafe.Pointer(&ret))
// delta is 0.0166667
```

TotallyGamerJet:
Float32 was being converted into float64 which isn't right.

Closes #176

### DIFF
--- a/func_test.go
+++ b/func_test.go
@@ -5,6 +5,7 @@ package purego_test
 
 import (
 	"fmt"
+	"math"
 	"runtime"
 	"testing"
 	"unsafe"
@@ -83,6 +84,41 @@ func Test_qsort(t *testing.T) {
 	for i := range data {
 		if data[i] != sorted[i] {
 			t.Errorf("got %d wanted %d at %d", data[i], sorted[i], i)
+		}
+	}
+}
+
+func TestRegisterFunc_Floats(t *testing.T) {
+	library, err := getSystemLibrary()
+	if err != nil {
+		t.Errorf("couldn't get system library: %s", err)
+	}
+	libc, err := openLibrary(library)
+	if err != nil {
+		t.Errorf("failed to dlopen: %s", err)
+	}
+	{
+		var sinf func(arg float32) float32
+		purego.RegisterLibFunc(&sinf, libc, "sinf")
+		const (
+			arg float32 = 2
+		)
+		got := sinf(arg)
+		expected := float32(math.Sin(float64(arg)))
+		if got != expected {
+			t.Errorf("sinf failed. got %f but wanted %f", got, expected)
+		}
+	}
+	{
+		var sin func(arg float64) float64
+		purego.RegisterLibFunc(&sin, libc, "sin")
+		const (
+			arg float64 = 1
+		)
+		got := sin(arg)
+		expected := math.Sin(arg)
+		if got != expected {
+			t.Errorf("sin failed. got %f but wanted %f", got, expected)
 		}
 	}
 }
--- a/func_test.go
+++ b/func_test.go
@@ -5,7 +5,6 @@ package purego_test
 
 import (
 	"fmt"
-	"math"
 	"runtime"
 	"testing"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


JupiterRider:
@TotallyGamerJet I tested this fix. It seems to work 🎉

TotallyGamerJet:
Great! Just got to find a libc function that returns `float` so that there's test.

hajimehoshi (approved):
LGTM, thanks!

Should we backport this to stable branches?

TotallyGamerJet:
Yeah we can back port it since the fix is 5 lines.

hajimehoshi:
Oh the tests are failing now?

func_test.go
```diff
@@ -86,3 +87,38 @@ func Test_qsort(t *testing.T) {
 		}
 	}
 }
+
+func TestRegisterFunc_Floats(t *testing.T) {
+	library, err := getSystemLibrary()
+	if err != nil {
+		t.Errorf("couldn't get system library: %s", err)
```
hajimehoshi, line 93: t.Fatalf

func_test.go
```diff
@@ -86,3 +87,38 @@ func Test_qsort(t *testing.T) {
...
+	}
+	libc, err := openLibrary(library)
+	if err != nil {
+		t.Errorf("failed to dlopen: %s", err)
```
hajimehoshi, line 97: t.Fatalf

TotallyGamerJet:
Only because libc.so.6 on Ubuntu doesn't have a `sin` and `sinf` function for some reason. I changed the test to use `strtof` and `strtod` which I believe Ubuntu does have.

func_test.go
TotallyGamerJet: Don't we want to run both functions before failing? t.Fatalf ends the test so we don't know if the other one succeeded or not.

hajimehoshi: When a system library is not found, we cannot test the other tests, right?

hajimehoshi: Or, is it possible that a library is found and an error is returned at the same time...?

TotallyGamerJet: That is true bc it will panic if the symbol isn't found. However, if the symbol is found it may not return the correct value which is what is being tested. We'd want to test both float32 and float64 even if one fails.

hajimehoshi: Does `getSystemLibrary` search symbols? I don't suggest replacing all `Errorf` with `Fatalf`. Rather, I suggest replacing only two `Errorf` for `getSystemLibrary` and `openLibrary`.

TotallyGamerJet: Oh ok. That makes sense. Should we do that then for `Test_qsort` too?

hajimehoshi: Yeah. (Also TestRegisterFunc?)

hajimehoshi (approved):
Still LGTM, thanks!

Please merge this with comments when you like. I'll cherry-pick the fix to 0.5 later (maybe tomorrow). Thank you for your quick work!

hajimehoshi:
Didn't the tests pass yet?

JupiterRider:
@TotallyGamerJet Could you tag this or something, so the code is available at pkg.go.dev?

TotallyGamerJet:
I believe @hajimehoshi will do that

hajimehoshi:
I'll add a tag few days after the last fix, not now.

hajimehoshi:
> Please merge this with comments when you like.

My words were not good, but please wait for test results before merging, thanks!

TotallyGamerJet:
Whoops sorry about that

TotallyGamerJet:
@hajimehoshi  which did that commit fail in the 0.5 branch on Ubuntu? It passed in https://github.com/ebitengine/purego/pull/177/commits/877753f1a0d3782d7f5e281ed81fec0d0985d9fc

hajimehoshi:
I cherry-picked 63c6fec935c5d014be719ef339e930658eb39213

TotallyGamerJet:
I understand that. I was wondering why it had failed. It seems to be because 0.5 doesn't have #163 which added callbacks on Linux which the float tests need.

hajimehoshi:
Thanks! We should not cherry-pick 092d3a90dd8b57e94832c070f4e2d4c79e410605 for 0.5 tests since this is risky. 

Then, I'd cherry-pick the fix with this code snippet to avoid tests on Linux as other tests do:

```go
        if runtime.GOOS == "linux" {                                                                                                                                                                                                                                                  
                // TODO: enable once callbacks are working properly on Linux                                                                                                                                                                                                          
                t.Skip("callbacks are not supported on Linux")                                                                                                                                                                                                                        
        } 
```

What do you think?

hajimehoshi:
> What do you think?

I think we have reached an agreement :) https://github.com/ebitengine/purego/issues/176#issuecomment-1817585671

TotallyGamerJet:
Yes. Your solution is a good one and the test passed on Linux

PR #181, opened 2023-11-20
change how bool return values are converted

JupiterRider:
Sometimes C functions don't return 0 when false, but instead this conversion works.

### DIFF
--- a/func.go
+++ b/func.go
@@ -259,7 +259,7 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
 			v.SetInt(int64(r1))
 		case reflect.Bool:
-			v.SetBool(*(*bool)(unsafe.Pointer(&r1)))
+			v.SetBool(byte(r1) != 0)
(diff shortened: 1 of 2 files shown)


JupiterRider:
@TotallyGamerJet What do you think?

TotallyGamerJet:
It what case does the original not work? Isn't the go implementation of bools just a byte where 0 is false and other values true?

JupiterRider:
I am working on a purego version of raylib.
I figured out, that some functions didn't work. They always returned true.

Here is an example:
```golang
var isFileDropped func() uintptr
purego.RegisterLibFunc(&isFileDropped, raylib, "IsFileDropped")
```

`fmt.Println(isFileDropped())` returns `140021448219136`

https://github.com/raysan5/raylib/blob/master/src/rcore.c#L2205

I don't know, why this behavior affects only some functions.

TotallyGamerJet:
Please make an issue and then answer these questions: Well that number is indeed not zero. Do u expect it to return false? what number does the C version return?

TotallyGamerJet (approved):
LGTM @hajimehoshi ?

hajimehoshi (approved):
LGTM!

PR #182, opened 2023-11-21
Increase max argument to 15

TotallyGamerJet:
This increases the max number of arguments to be 15. This is necessary because amd64 puts almost everything on the stack when passing structs. This meant that there wasn't enough room with just 9 arguments to pass all the tests that arm64 has in the struct branch.

### DIFF
--- a/callback_test.go
+++ b/callback_test.go
@@ -126,18 +126,18 @@ func TestNewCallbackFloat32(t *testing.T) {
 func TestNewCallbackFloat32AndFloat64(t *testing.T) {
 	// This tests that calling a function with a mix of float32 and float64 arguments works
 	const (
-		expectedCbTotalF32 = float32(30)
+		expectedCbTotalF32 = float32(72)
 		expectedCbTotalF64 = float64(48)
 	)
 	var cbTotalF32 float32
 	var cbTotalF64 float64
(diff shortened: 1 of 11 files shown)


TotallyGamerJet:
I'm looking into why it's failing with `morestack on g0`. It doesn't fail on my machine. @hajimehoshi do you remember how you fixed that for the Switch?

hajimehoshi:
> `morestack on g0`

This should be fixed in 1.21.4 (https://github.com/golang/go/issues/62440). This doesn't occur in 1.20 and older.

TotallyGamerJet:
But the 1.19 test is the one that failed?

hajimehoshi:
> But the 1.19 test is the one that failed?

So unfortunately this seems a different thing from what I've seen in Switch.

TotallyGamerJet:
This is now ready for review!

hajimehoshi (approved):
LGTM, thanks!

Just out of curiosity, is there a limitation of float argument numbers?

TotallyGamerJet:
No there shouldn't be since once all the float registers are full they get placed on the stack. I'll add a test to this PR to prove that

progrium:
👀

hajimehoshi:
Let me know when this pr is ready

TotallyGamerJet:
@hajimehoshi it is ready to be merged if the new test also looks good

PR #120, opened 2023-04-01
purego: add float32 and float64 support to callbacks

TotallyGamerJet:
This commit adds support for float32 and float64 arguments in C callbacks. It also adds a test to check the maximum number of arguments a callback can have and that it is callback properly by purego. There is also some formatting done to the assembly.

PR #190, opened 2024-01-07
Remove WIP comment for float & double

TotallyGamerJet:
floats and doubles were already implemented in #67 #69 #120 #171 #177


hajimehoshi (approved):
LGTM

PR #192, opened 2024-01-08
Expose `Dlopen` / `Dlsym` / `Dlclose` on Windows

RomainMuller:
Make the API consistent across platforms, so that consumers can use purego wihtout having to bother with Windows-specific shenanigans. Simply wrap `windows.LoadLibrary`, `windows.GetProcAddress` and `windows.FreeLibrary` and cast the returned values to `purego` Api types.

This makes it tremendously easier ot consume `purego` in apps that target both Windows and *NIXes.

### DIFF
--- a/dlfcn_windows.go
+++ b/dlfcn_windows.go
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package purego
+
+import (
+	"golang.org/x/sys/windows"
+)
+
+const (
+	RTLD_DEFAULT = 0x00000 // Dlopen flags are not used on Windows.
+)
+
+// Dlopen examines the dynamic library or bundle file specified by path. If the file is compatible
+// with the current process and has not already been loaded into the
+// current process, it is loaded and linked. After being linked, if it contains
+// any initializer functions, they are called, before Dlopen
+// returns. It returns a handle that can be used with Dlsym and Dlclose.
+// A second call to Dlopen with the same path will return the same handle, but the internal
+// reference count for the handle will be incremented. Therefore, all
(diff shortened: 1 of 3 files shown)


TotallyGamerJet:
This was already discussed in #107 and rejected. I don't think there is any new information that would change the decision. Agreed? @hajimehoshi

hajimehoshi:
I'm still against adding them for Windows.

platypusguy:
I've read through this thread and both earlier and later threads on the same topic, and I understand your rationale. . As an OSS provider myself, I am wary of telling people who volunteer their time what more they should do to suit my needs. So please understand, this is just an observation (which I admit would suit my needs). 

If you look at the question only from the perspective of the present library, then sure, I see your point. But if you look at it from the point of view of a golang library, then the ability to have the same code run unchanged over all the major supported OSs is both a desirable and attractive feature. The golang team worked hard to make this kind of you-don't-have-to-test-for-the-OS portability a big part of the language and the core libraries. The requested change in this thread would be consistent with that work and, I believe, the expectation of most golang programmers. 

The frequency with which this thread comes up communicates this expectation, at least IMHO. 

Either way, thanks for a great library.

TotallyGamerJet:
I appreciate you taking to the time to read through and formulate your response. Yes, Go strives to have a portable API for its functions so you can build for any platform. This is one of my favorite things about the language. These portable packages though are higher level and abstract away the OS system calls. 

Purego on the other hand is more in line with the syscall package which are not cross platform without build tags. Also, the main use of purego is calling into system libraries which are mostly different between windows and macOS. Purego can be used with libraries built into shared object files but then one must figure out how to distribute it which losses Go's single binary distribution. Even if you do need the same API it is only two files to write which is simple enough to copy into your own Go code from the libc example.

PR #203, opened 2024-01-18
move C tests into testdata/

TotallyGamerJet:
These don't have public API so move them into internal to cleanup root folder


hajimehoshi:
Though I suggested using `internal` in the Discord server, but I realized that `testdata` directory sounds better.

https://pkg.go.dev/cmd/go#hdr-Test_packages

> The go tool will ignore a directory named "testdata", making it available to hold ancillary data needed by the tests.

Assuming `libcbtest`, `libdlnested`, and `structtest` are not Go packages, what about moving them into `testdata` instead of `internal`? Or, if there is a possibility that they become Go packages, `internal` is better. What do you think?

TotallyGamerJet:
There is no intention of them being go packages. Moving them into testdata is a great idea

hajimehoshi (approved):
LGTM, thanks!

PR #211, opened 2024-02-19
Fix Three Small Fields losing the last field

### LINKED ISSUE #210: Struct with float32s passed alongside other float32s as arguments is broken on GOARCH=amd64
https://gist.github.com/SealOfTime/f5a962a2182c1c9b264117ebac18fe50

When passing a structure, containing float32s, to a function alongside other float32 the resulting arguments are messed up.

Tested on GOOS=darwin GOARCH=amd64. Was able to reproduce with GOOS=linux by disabling panic, that restricts the feature only to `darwin`.

GOOS=darwin GOARCH=arm64 works well

SealOfTime:
There was a bug, that caused the last small field, that directly followed a set of small fields that combined into one full machine word, in a struct to not be added to the list of arguments.
It was caused by incorrectly marking that the field that is being processed right now is flushed if processing of previous small field resulted in `val` getting filled enough to be flushed.

In order to avoid similar errors and to have flushing logic at one place I've moved that out into an anonymous function. It should impact performance somewhat, I guess I could just inline it. Would like to hear your opinion on that!

Closes #210

### DIFF
--- a/struct_amd64.go
+++ b/struct_amd64.go
@@ -168,8 +168,7 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 
 			if shift == 64 {
 				flush()
-			}
-			if shift > 64 {
+			} else if shift > 64 {
 				// Should never happen, but may if we forget to reset shift after flush (or forget to flush),
 				// better fall apart here, than corrupt arguments.
 				panic("purego: tryPlaceRegisters shift > 64")
--- a/struct_amd64.go
+++ b/struct_amd64.go
@@ -94,13 +94,16 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 				f = v.Index(i)
 			}
 			if shift >= 64 {
+				// not marking flushed = true even though have flushed
+				// because now we're processing a new field
+				// and if we exit now - we have to flush that field
 				shift = 0
-				flushed = true
 				if class == _SSE {
 					addFloat(uintptr(val))
 				} else {
 					addInt(uintptr(val))
 				}
+				val = 0
 				class = _NO_CLASS
 			}
 			switch f.Kind() {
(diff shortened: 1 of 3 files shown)


struct_amd64.go
```diff
@@ -158,24 +157,29 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
...
 			case reflect.Array:
 				place(f)
 			default:
 				panic("purego: unsupported kind " + f.Kind().String())
 			}
+
+			if shift == 64 {
+				flush()
+			}
+			if shift > 64 {
```
TotallyGamerJet, line 172: Else if

TotallyGamerJet (changes requested):

TotallyGamerJet (approved):
LGTM. Separating the flushing code into another function is a good idea. I'll prolly do the same for arm64 in another PR @hajimehoshi

struct_amd64.go
```diff
@@ -76,6 +76,17 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 	var shift byte // # of bits to shift
 	var flushed bool
 	class := _NO_CLASS
+	flush := func() {
```
hajimehoshi, line 79: What about

```go
flushIfNeeded := func() {
    if flushed {
        return
    }
    flushed = true
    // ...
```

so that we don't need an if statement to call flushing?

SealOfTime: I don't see any benefits in it, could you please help me with it?

`if flushed` is used only once after we've iterated over all fields. It's only purpose is to flush the last small fields, that didn't trigger `if shift == 64` condition. But if we move it to the `flush` function, it'll be executed every flush. Therefore adding a tiny but still accumulating penalty to calls with structures as arguments.

hajimehoshi: This is just for readability and maintenability around callers.

For the penalty, yes, you are right. I was a little worried about the future when we need another flush call but forgetting `if flushed` check, but as long as this unlikely happens, the current change seems fine enough.

hajimehoshi (approved):
LGTM

PR #212, opened 2024-02-20
purego: refactor struct_amd64.go

hajimehoshi:


TotallyGamerJet (approved):
LGTM

hajimehoshi:
So, on second thought, I thought the cost of `if` branch would not matter compared to other costs we've added. For maintenability, I prefer this `flushIfNeeded` way. /CC @SealOfTime

SealOfTime:
Flush function is quite likely inlined at a call-site. Either way to reason about performance we'd better do benchmarks. Though I completely agree with maintainability claims regarding `flushIfNeeded` over just `flush`

PR #213, opened 2024-02-24
purego: Add struct return support for darwin amd64 and arm64

### LINKED ISSUE #195: Support returning struct
Similar to #88 but about returning structs.

TotallyGamerJet:
Add support for returning structs who's fields are floats or integer or struct or arrays containing floats or integers

Closes #195


func.go
```diff
@@ -134,6 +135,9 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 					stack++
 				}
 			case reflect.Float32, reflect.Float64:
+				if is32bit() {
+					panic("purego: floats only supported on 64bit platforms")
```
hajimehoshi, line 139: Is this panic related to returning structs?

func.go
```diff
@@ -317,10 +346,54 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
 	return (val + 7) &^ 7
 }
 
+func is32bit() bool {
+	switch runtime.GOARCH {
+	case "arm64", "amd64":
+		return false
+	default:
+		return true
+	}
```
hajimehoshi, line 394: What about 

```go
const is32bit = unsafe.Sizeof(uintptr(0)) == 4
```

?

testdata/structtest/structreturn_test.c
```diff
@@ -0,0 +1,193 @@
...
+     int32_t b;
+     float c;
+     int32_t d;
+};
+
+
+struct Mixed2 ReturnMixed2(float a, int32_t b, float c, int32_t d) {
+    struct Mixed2 e = {a, b, c, d};
+    return e;
+}
```
hajimehoshi, line 193: Add a new line.

struct_amd64.go
```diff
@@ -6,8 +6,48 @@ package purego
...
+		} else {
+			// up to 16 bytes is returned in RAX and RDX
+			v = reflect.NewAt(outType, unsafe.Pointer(&struct{ a, b uintptr }{syscall.a1, syscall.a2})).Elem()
+		}
+	} else {
+		// create struct from the Go pointer created above
+		// weird pointer dereference to circumvent go vet
+		v = reflect.NewAt(outType, *(*unsafe.Pointer)(unsafe.Pointer(&syscall.a1))).Elem()
+	}
+	return v
```
hajimehoshi, line 36: I prefer early returns like

```go
	if outSize == 0 {
		return reflect.New(outType).Elem()
	}

	if outSize <= 8 {
		if isAllFloats(outType) {
			// float64s are return in the float register
			return reflect.NewAt(outType, unsafe.Pointer(&struct{ a uintptr }{syscall.f1})).Elem()
		}
		// up to 8 bytes is returned in RAX
		return reflect.NewAt(outType, unsafe.Pointer(&struct{ a uintptr }{syscall.a1})).Elem()
	}

	if outSize <= 16 {
		if isAllFloats(outType) {
			return reflect.NewAt(outType, unsafe.Pointer(&struct{ a, b uintptr }{syscall.f1, syscall.f2})).Elem()
		}
		// up to 16 bytes is returned in RAX and RDX
		return reflect.NewAt(outType, unsafe.Pointer(&struct{ a, b uintptr }{syscall.a1, syscall.a2})).Elem()
	}

	// create struct from the Go pointer created above
	// weird pointer dereference to circumvent go vet
	return reflect.NewAt(outType, *(*unsafe.Pointer)(unsafe.Pointer(&syscall.a1))).Elem()
```

func.go
```diff
@@ -317,10 +346,54 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 	fn.Set(v)
 }
 
+func isAllSameFloat(ty reflect.Type) bool {
+	first := ty.Field(0).Type.Kind()
```
hajimehoshi, line 350: Doesn't this cause panic for an empty struct?

func.go
TotallyGamerJet: It's not. I had just realized it would just silently fail on 32bit performs. I can separate that if you want

func.go
TotallyGamerJet: It would but the size is always checked before entering this function so that doesn't happen. There is a test for empty return structs

func.go
hajimehoshi: Yeah, please separate this. The smaller the change is, the easier we can review.

hajimehoshi: So, will you be sending the PR for this part now, or later?

func.go
```diff
@@ -317,6 +343,41 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+		switch f.Type.Kind() {
+		case first:
+		default:
+			return false
+		}
+	}
+	return true
+}
+
+func checkStruct(ty reflect.Type) {
```
hajimehoshi, line 362: Can we have a better name like `areStructMembersSupported`?

struct_amd64.go
```diff
@@ -6,8 +6,47 @@ package purego
...
 	"math"
 	"reflect"
+	"unsafe"
 )
 
+func getStruct(outType reflect.Type, syscall syscall15Args) (v reflect.Value) {
+	outSize := outType.Size()
+	if outSize == 0 {
+		return reflect.New(outType).Elem()
+	} else if outSize <= 8 {
```
hajimehoshi, line 16: We don't need `else`, right?

struct_arm64.go
```diff
@@ -6,8 +6,59 @@ package purego
...
 	"math"
 	"reflect"
+	"unsafe"
 )
 
+func getStruct(outType reflect.Type, syscall syscall15Args) (v reflect.Value) {
+	outSize := outType.Size()
+	if outSize == 0 {
+		return reflect.New(outType).Elem()
+	} else if outSize <= 8 {
```
hajimehoshi, line 16: ditto

func.go
```diff
@@ -317,6 +343,41 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+		switch f.Type.Kind() {
+		case first:
+		default:
+			return false
+		}
+	}
+	return true
+}
+
+func areStructFieldsSupported(ty reflect.Type) {
```
hajimehoshi, line 362: Sorry I was misunderstanding that this function returned a bool, but this function just does assertions. So `checkStructFieldsSupported` might be better? Sorry for going back and forth.

struct_amd64.go
hajimehoshi: I mean, why not

```go
	if outSize == 0 {
		return reflect.New(outType).Elem()
	}
	if outSize <= 8 {
```

TotallyGamerJet: Are you saying just for the first `else if`? If so, idc either way but I was thinking grouping them made it clear that the conditionals build off of each other. ==0, <=8, <=16, >16

hajimehoshi: All the `else if`. If you want to make it explicit that all the conditions are exclusive, I prefer switch-case.

```go
switch {
case outSize == 0:
    return reflect.New(outType).Elem()
case outSize <= 8:
```

struct_test.go
```diff
@@ -439,3 +439,223 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+			c float32
+			d int32
+		}
+		var ReturnMixed2 func(a float32, b int32, c float32, d int32) Mixed2
+		purego.RegisterLibFunc(&ReturnMixed2, lib, "ReturnMixed2")
+		expected := Mixed2{1, 2, 3, 4}
+		if ret := ReturnMixed2(1, 2, 3, 4); ret != expected {
+			t.Fatalf("ReturnMixed2 returned %+v wanted %+v", ret, expected)
+		}
+	}
```
hajimehoshi, line 660: Can we test a struct including all the types: float32, float64, and others like int32?

TotallyGamerJet: I added one. Not sure how helpful it is since any struct bigger than 16 bytes is allocated in the Go heap and passed in for the C function to use.

hajimehoshi: Wouldn't this be <= 16 bytes?

```go
type Mixed struct {
    a float32
    b int32
    c float64
}
```

TotallyGamerJet: Yeah! Do that instead of Mixed3 or in addition to?

hajimehoshi: Doing it instead of Mixed3 seems enough

func.go
```diff
@@ -320,6 +346,41 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+		if f.Kind() == reflect.Array {
+			f = f.Elem()
+		} else if f.Kind() == reflect.Struct {
+			checkStructFieldsSupported(f)
+			continue
+		}
+		switch f.Kind() {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+			reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
+			reflect.Uintptr, reflect.Float64, reflect.Float32:
```
SealOfTime, line 377: No `reflect.Bool`?

func.go
```diff
@@ -163,6 +164,16 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 				panic("purego: unsupported kind " + arg.Kind().String())
 			}
 		}
+		if ty.NumOut() > 0 && ty.Out(0).Kind() == reflect.Struct {
+			if runtime.GOOS != "darwin" {
+				panic("purego: struct return values only supported on darwin arm64 & amd64")
+			}
```
SealOfTime, line 170: Why do you check for function having more than 0 return parameters, while at line 244 for precise equivalence to one?

Imo, since right now (and likely forever) we don't support more than 1 return parameter, it'd be better to be stricter about it and panic with a corresponding message, like `purego:  functions may return only one value`

func.go
```diff
@@ -163,6 +164,16 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+			outType := ty.Out(0)
+			checkStructFieldsSupported(outType)
+			if outType.Size() > 16 {
+				ints++
+			}
```
SealOfTime, line 175: It seems that for a structure like `struct{a, b bool}` we won't increment amount of occupied integer registers, is this intentional?

func.go
SealOfTime: > are floats or integer or struct or arrays containing floats or integers

Get it now. I don't understand tho, how is bool different from Int8?

func.go
SealOfTime: Oh, I get it now. We use additional parameter to pass a pointer for a struct more than 16 bytes in size. It's nitpicking, but I'd really love a comment about, why we use an int register to receive a return value

func.go
```diff
@@ -163,6 +164,16 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
 			}
 		}
+		if ty.NumOut() > 0 && ty.Out(0).Kind() == reflect.Struct {
+			if runtime.GOOS != "darwin" {
+				panic("purego: struct return values only supported on darwin arm64 & amd64")
+			}
+			outType := ty.Out(0)
+			checkStructFieldsSupported(outType)
+			if outType.Size() > 16 {
+				ints++
```
SealOfTime, line 174: Why don't we check for ints > numOfIntegerRegisters() here?

func.go
```diff
@@ -229,6 +240,21 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 			runtime.KeepAlive(keepAlive)
 			runtime.KeepAlive(args)
 		}()
+		var syscall syscall15Args
+		if ty.NumOut() == 1 && ty.Out(0).Kind() == reflect.Struct {
+			outType := ty.Out(0)
+			if runtime.GOARCH == "amd64" && outType.Size() > 16 {
```
SealOfTime, line 246: Maybe check for `outType.Size() > 16` and then specialize for specific platform? Also it's not critical right now, as `darwin` -exclusivity is hardcoded up the code, but how about we panic we GOARCH is neither `amd64`, not `arm64`?

func.go
SealOfTime: Took me a moment to realize that as well, wouldn't it be helpful to add a comment here, explaining why that's so?

func.go
```diff
@@ -320,6 +346,41 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+	if first != reflect.Float32 && first != reflect.Float64 {
+		return false
+	}
+	for i := 0; i < ty.NumField(); i++ {
+		f := ty.Field(i)
+		switch f.Type.Kind() {
+		case first:
+		default:
+			return false
+		}
```
SealOfTime, line 360: Why switch and not just if for one branch condition?

func.go
TotallyGamerJet: Checking for 1 is probably more clear. No need to panic here since it's already done at L115

func.go
```diff
@@ -260,25 +286,23 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
 			// Use the normal arm64 calling convention even on Windows
-			syscall := syscall15Args{
+			syscall = syscall15Args{
 				cfn,
 				sysargs[0], sysargs[1], sysargs[2], sysargs[3], sysargs[4], sysargs[5],
 				sysargs[6], sysargs[7], sysargs[8], sysargs[9], sysargs[10], sysargs[11],
 				sysargs[12], sysargs[13], sysargs[14],
 				floats[0], floats[1], floats[2], floats[3], floats[4], floats[5], floats[6], floats[7],
-				0, 0, 0,
+				0, 0, 0, syscall.arm64_r8,
```
SealOfTime, line 297: Is syscall_arm64_r8 passed as `err` field?

func.go
TotallyGamerJet: The check is done after this at L177

func.go
TotallyGamerJet: No reason to add more checks in the actual function call since it's done on creation.

func.go
TotallyGamerJet: No reason, it had multiple cases but then I changed it to one case and just left it as a switch

struct_arm64.go
```diff
@@ -6,8 +6,58 @@ package purego
...
+			switch outType.NumField() {
+			case 4:
+				r1 = syscall.f2<<32 | syscall.f1
+				r2 = syscall.f4<<32 | syscall.f3
+			case 3:
+				r1 = syscall.f2<<32 | syscall.f1
+				r2 = syscall.f3
+			case 2:
+				r1 = syscall.f1
+				r2 = syscall.f2
```
SealOfTime, line 38: That's so weird. So for two floats instead of cramming them into one reg, arm abi uses both regs?

struct_test.go
```diff
@@ -439,3 +439,240 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+		type TwoFloats struct{ a, b float32 }
+		var ReturnTwoFloats func(a, b float32) TwoFloats
+		purego.RegisterLibFunc(&ReturnTwoFloats, lib, "ReturnTwoFloats")
+		expected := TwoFloats{3, 10}
+		if ret := ReturnTwoFloats(5, 2); ret != expected {
+			t.Fatalf("ReturnTwoFloats returned %+v wanted %+v", ret, expected)
+		}
+	}
+	{
+		type ThreeFloats struct{ a, b, c float32 }
```
SealOfTime, line 542: How would arm getStruct handle `struct {a float64, b, c float32}`?

struct_arm64.go
SealOfTime: Oh I get it now, it's because they are more than 8 bytes in size, therefore double + float

struct_arm64.go
```diff
@@ -6,8 +6,58 @@ package purego
...
+			default:
+				panic("unreachable")
+			}
+		}
+		return reflect.NewAt(outType, unsafe.Pointer(&struct{ a, b uintptr }{r1, r2})).Elem()
```
SealOfTime, line 43: Oh wow, so for non-uniform structure ARM abi won't even use floating point registers?

struct_test.go
SealOfTime: I see it's checked below

sys_amd64.s
```diff
@@ -86,8 +86,10 @@ TEXT syscall15X(SB), NOSPLIT|NOFRAME, $0
 	CALL R10
 
 	MOVQ PTR_ADDRESS(BP), DI      // get the pointer back
-	MOVQ AX, syscall15Args_r1(DI) // r1
-	MOVQ X0, syscall15Args_r2(DI) // r2
+	MOVQ AX, syscall15Args_a1(DI) // r1
+	MOVQ DX, syscall15Args_a2(DI) // r3
+	MOVQ X0, syscall15Args_f1(DI) // f1
+	MOVQ X1, syscall15Args_f2(DI) // f1
```
SealOfTime, line 92: typo: f2

func.go
TotallyGamerJet: No, it goes r1, r2, err, arm64_r8

func.go
hajimehoshi: Is C's `_Bool` always 8bits?

func.go
hajimehoshi: Having a comment to explaing why 16 sounds nice.

hajimehoshi: Also, 16 is a magic number. Let's define a constant and use it everywhere.

struct_arm64.go
TotallyGamerJet: Yes they are either all the same float type or it's passed in registers

struct_amd64.go
```diff
@@ -6,8 +6,46 @@ package purego
...
 )
 
+func getStruct(outType reflect.Type, syscall syscall15Args) (v reflect.Value) {
+	outSize := outType.Size()
+	switch {
+	case outSize == 0:
+		return reflect.New(outType).Elem()
+	case outSize <= 8:
+		if isAllFloats(outType) {
+			// float64s are return in the float register
```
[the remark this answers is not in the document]
TotallyGamerJet: yes

func.go
TotallyGamerJet: I believe `_Bool` isn't used very often in C structs so I didn't bother since I'd have to add tests for all the different combinations of ints, floats, and bools.

sys_amd64.s
hajimehoshi: Is this resolved?

hajimehoshi (approved):
LGTM, thanks!

PR #220, opened 2024-03-28
all: allow running programs that import purego in the playground

TotallyGamerJet:
Currently, if a package imports purego in the playground even if none of the functions are called the run will fail with a complaint about unable to link in libdl.so.2. This commit circumvents this by stubbing out the dynamic linking pragmas for the go playground (`//go:build faketime`). This allows building and running apps in the playground that import purego but *doesn't* allow actually calling any of the functions. Currently, it will just die very ungracefully.

Playground: https://go.dev/play/p/t3PK-iTWSl-
Remove the replace in the go.mod to see old behavior.


hajimehoshi (approved):
LGTM

We can add kinder error messages instead of crash later.

PR #221, opened 2024-04-03
all: struct return fixes

### LINKED ISSUE #223: objc: returning  a struct only with floats from `Send` crashes on amd64 and doesn't work on arm64
The fix is #221

TotallyGamerJet:
On amd64, there was a crash when calling objc.Send with a struct return type. This was due to calling the wrong objc message send function.

On arm64, if a struct return contained structs that only contained floats it would expect it to be in R8 instead of the expected float registers

Closes #223

### DIFF
--- a/func.go
+++ b/func.go
@@ -250,7 +250,8 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 				keepAlive = append(keepAlive, val)
 				addInt(val.Pointer())
 			} else if runtime.GOARCH == "arm64" && outType.Size() > maxRegAllocStructSize {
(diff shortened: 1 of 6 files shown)


hajimehoshi (approved):
LGTM, thanks!

Let's add "Closes #223" to a commit message.

TotallyGamerJet:
@hajimehoshi if this does fix the issue in ebitengine should this be cherry-picked for 0.8?

hajimehoshi:
Confirmed this worked perfectly with Ebitengine. I'll update go.mod.

PR #226, opened 2024-04-05
objc: remove deprecated APIs

hajimehoshi:
Perhaps, do we need new examples?

### DIFF
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -193,13 +193,6 @@ func GetClass(name string) Class {
 	return objc_getClass(name)
 }
 
-// AllocateClassPair creates a new class and metaclass. Then returns the new class, or Nil if the class could not be created
-//
-// Deprecated: use RegisterClass instead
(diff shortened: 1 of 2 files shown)


hajimehoshi:
Updated with RegisterClass.

objc/objc_runtime_darwin.go
```diff
@@ -583,7 +554,7 @@ type IMP uintptr
 func NewIMP(fn interface{}) IMP {
 	ty := reflect.TypeOf(fn)
 	if ty.Kind() != reflect.Func {
-		panic("objc: not a function")
+		panic(fmt.Sprintf("objc: not a function"))
```
TotallyGamerJet, line 557: Why the change to Sprintf?

TotallyGamerJet (approved):
Other than comment LGTM

objc/objc_runtime_darwin.go
hajimehoshi: oops this was my mistake

TotallyGamerJet:
LGTM

PR #229, opened 2024-04-06
Consolidate struct definition

### LINKED ISSUE #228: Error on Android: `too few values in struct literal of type syscall15Args`
```
/Users/runner/go/pkg/mod/github.com/ebitengine/purego@v0.8.0-alpha.0.20240404024320-d0aedd0f4393/func.go:301:4: too few values in struct literal of type syscall15Args
```

This caused by my recent change: https://github.com/hajimehoshi/ebiten/commit/5701206229c345816511bae5ab466f113a2810e1

TotallyGamerJet:
The struct on Cgo non-amd64 and non-arm64 platforms didn't have the struct fields updated. To avoid forgetting to update all the structs I've moved it where every platform shares the same type def.

Closes #228

### DIFF
--- a/syscall_cgo_linux.go
+++ b/syscall_cgo_linux.go
@@ -13,14 +13,6 @@ import (
 
 var syscall15XABI0 = uintptr(cgo.Syscall15XABI0)
 
-// this is only here to make the assembly files happy :)
-type syscall15Args struct {
-	fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15 uintptr
-	f1, f2, f3, f4, f5, f6, f7, f8                                       uintptr
-	r1, r2, err                                                          uintptr
-	arm64_r8                                                             uintptr
-}
-
 //go:nosplit
 func syscall_syscall15X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15 uintptr) (r1, r2, err uintptr) {
 	return cgo.Syscall15X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15)
--- a/syscall.go
+++ b/syscall.go
@@ -10,6 +10,12 @@ const (
 	numOfFloats = 8 // arm64 and amd64 both have 8 float registers
 )
 
+type syscall15Args struct {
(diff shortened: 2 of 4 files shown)


TotallyGamerJet:
We should probably add a test for non-standard platforms but this would require a C compiler.

hajimehoshi:
Thanks.

> We should probably add a test for non-standard platforms but this would require a C compiler.

We would need a cross compiler or QEMU. It'd be nice to have.

hajimehoshi (approved):
LGTM, thanks!

syscall_cgo_linux.go
```diff
@@ -13,14 +13,6 @@ import (
 
 var syscall15XABI0 = uintptr(cgo.Syscall15XABI0)
 
-// this is only here to make the assembly files happy :)
-type syscall15Args struct {
-	fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15 uintptr
-	f1, f2, f3, f4, f5, f6, f7, f8                                       uintptr
-	r1, r2, err                                                          uintptr
```
hajimehoshi, line 20: Wouldn't this information (r1, r2, err) missing?

hajimehoshi (changes requested):
Let me ask a question...

syscall_cgo_linux.go
TotallyGamerJet: Do you mean the from the error message? func.go doesn't use the field marker but writes everything field unnamed

hajimehoshi: I mean, I was not sure it was really fine to remove these members. The new syscall15Args doesn't have r1, r2, and err, right?

TotallyGamerJet: They aren't used anywhere. Arguments are now in a1, f1 or R8

hajimehoshi: Thanks!

hajimehoshi (approved):
LGTM, thanks!

TotallyGamerJet:
So doing a separate commit for tests?

hajimehoshi:
> So doing a separate commit for tests?

If this is easy, it is better to have a test in this pr. If not, a sperate pr is fine. I thought it was not easy.

TotallyGamerJet:
I don't think it'll be easy

hajimehoshi:
Ok, then let's add a test later in a separate pr :-)

PR #234, opened 2024-04-16
Add RegisterFunc on unsupported platforms

### LINKED ISSUE #230: enable `RegisterFunc` and `RegisterLibFunc` for non-supported architectures
Now `RegisterFunc` and `RegisterLibFunc` don't work on non-supported architectures (e.g. Linux 386). If they could be available even with Cgo, it would be nice since we could share the same code using PureGo for any architectures.

There is an actual use case: I tried to fix `github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver/opengl/gl` to use PureGo for Linux, but failed due to this error:

```
panic: purego: unknown GOARCH (386)

goroutine 1 [running]:
github.com/ebitengine/purego.numOfIntegerRegisters(...)
	/home/runner/go/pkg/mod/github.com/ebitengine/purego@v0.8.0-alpha.0.20240407111646-0e6302132b78/func.go:422
github.com/ebitengine/purego.RegisterFunc({0x839c5a0, 0x86c4bdc}, 0xeac822b0)
	/home/runner/go/pkg/mod/github.com/ebitengine/purego@v0.8.0-alpha.0.20240407111646-0e6302132b78/func.go:132 +0x144
github.com/ebitengine/purego.init.0()
	/home/runner/go/pkg/mod/github.com/ebitengine/purego@v0.8.0-alpha.0.20240407111646-0e6302132b78/dlfcn.go:22 +0x3b
```

TotallyGamerJet:
Closes #230

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -109,8 +109,6 @@ jobs:
       - name: go test (Linux 386)
         if: runner.os == 'Linux'
         run: |
-           sudo dpkg --add-architecture i386
-           sudo apt-get update
            sudo apt-get install gcc-multilib
            sudo apt-get install g++-multilib
            env CGO_ENABLED=1 GOARCH=386 go test -shuffle=on -v -count=10 ./...


TotallyGamerJet:
Current outstanding issues:

1. Windows 386 qsort test fails. I don't have a Windows computer to investigate so I just disabled it.
2. Add more "unsupported" platforms to the GitHub workflows.

func_test.go
```diff
@@ -126,7 +134,7 @@ func TestRegisterLibFunc_Bool(t *testing.T) {
 	// this callback recreates the state where the return register
 	// contains other information but the least significant byte is false
 	cbFalse := purego.NewCallback(func() uintptr {
-		return 0x7F5948AE9A00
+		return uintptr(uint64(0x7F5948AE9A00) & uint64(^uintptr(0)))
```
TotallyGamerJet, line 137: This number is too big for 32bit platforms so I masked it off to what fits in a uintptr. Should I add a comment explaining this?

TotallyGamerJet: We could use smaller values but I wanted to keep the test unchanged on 64Bit platforms since they have larger registers and this test was found on a 64Bit platform

hajimehoshi:
Hmm, let's fix the issue with 386 and RegisterFunc first, and then address on #230.

TotallyGamerJet:
Well it is now confirmed it is something wrong in RegisterFunc as just directly calling SyscallN worked just fine

syscall.go
```diff
@@ -5,6 +5,10 @@
 
 package purego
 
+// Cdecl wraps a function that should be called using the cdecl calling convention when passed to NewCallback.
+// This is only used on windows and does nothing for other platforms.
+type Cdecl any
```
hajimehoshi, line 10: Rename this to `CDecl`

func_test.go
```diff
@@ -9,14 +9,11 @@ import (
 	"testing"
 	"unsafe"
 
+	"github.com/ebitengine/purego/internal/load"
+
 	"github.com/ebitengine/purego"
```
hajimehoshi, line 14:
```go
	"github.com/ebitengine/purego/internal/load"
```

TotallyGamerJet:
We should probably run the CI on other architectures as well.

hajimehoshi:
> We should probably run the CI on other architectures as well.

Which architectures? Like, Linux 386?

syscall.go
```diff
@@ -5,6 +5,13 @@
 
 package purego
 
+// CDecl marks a function as being called using the __cdecl calling convention as defined in
+// the [MSDocs] when passed to NewCallback. It must be the first argument to the function.
+// This is only useful on 386, but it is safe to use on other platforms.
```
hajimehoshi, line 10: Is this only for Windows, or also for the other OSes?

hajimehoshi (commented):
We want to have tests with Linux 386, if it is possible.

syscall_windows.go
```diff
@@ -25,15 +25,24 @@ func syscall_syscall15X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a
 // callbacks can always be created. Although this function is similiar to the darwin version it may act
 // differently.
 func NewCallback(fn interface{}) uintptr {
+	isCDecl := false
+	ty := reflect.TypeOf(fn)
+	for i := 0; i < ty.NumIn(); i++ {
+		in := ty.In(i)
+		if !in.AssignableTo(reflect.TypeOf(CDecl{})) {
```
hajimehoshi, line 32: Can this be true when the first argument is `_ any`?

syscall.go
TotallyGamerJet: It's visible on all OSs so that there is no need to write basically the same thing on each

syscall_windows.go
TotallyGamerJet: No because `any` is not assignable to Cdecl which is a struct

hajimehoshi: OK thanks!

syscall.go
hajimehoshi: So

> This is only useful on 386

This is only useful on 386 Windows, right?

TotallyGamerJet: Correct

hajimehoshi: Let's make this explicit in this comment. (Just s/386/386 Windows/ is fine to me)

.github/workflows/test.yml
```diff
@@ -100,7 +100,18 @@ jobs:
...
+        run: |
+          env CGO_ENABLED=0 GOARCH=386 go test -shuffle=on -v -count=10 ./...
+          env CGO_ENABLED=1 GOARCH=386 go test -shuffle=on -v -count=10 ./...
+
+      - name: go test (Linux 386)
+        if: runner.os == 'Linux'
+        run: |
+          go env -w CC=i386-linux-gnu-gcc
+          go env -w CXX=i386-linux-gnu-g++
+          env CGO_ENABLED=1 GOOS=linux GOARCH=386 go test -shuffle=on -v -count=10 ./...
```
hajimehoshi, line 114: Would this work?

https://github.com/hajimehoshi/ebiten/blob/af4638d83f6a4714d8e637ec4545fc58ce5ceb7a/.github/workflows/test.yml#L142-L151

TotallyGamerJet:
@hajimehoshi this is ready for review

.github/workflows/test.yml
```diff
@@ -100,7 +100,23 @@ jobs:
...
+        if: runner.os == 'Linux'
+        run: |
+           sudo dpkg --add-architecture i386
+           sudo apt-get update
+           sudo apt-get install ppa-purge
+           sudo ppa-purge -y ppa:ubuntu-toolchain-r/test # Hack for 32bit Linux (#2667)
+           sudo apt-get install gcc-multilib
+           sudo apt-get install g++-multilib
+           sudo apt-get install libasound2-dev:i386 libgl1-mesa-dev:i386 libxcursor-dev:i386 libxi-dev:i386 libxinerama-dev:i386 libxrandr-dev:i386 libxxf86vm-dev:i386
+           env CGO_ENABLED=1 GOARCH=386 go test -shuffle=on -v -count=10 ./...
```
hajimehoshi, line 119: I was wondering if your simpler way would work:

```
          go env -w CC=i386-linux-gnu-gcc
          go env -w CXX=i386-linux-gnu-g++
          env CGO_ENABLED=1 GOOS=linux GOARCH=386 go test -shuffle=on -v -count=10 ./...
```

dlfcn_test.go
```diff
@@ -59,7 +59,9 @@ func buildSharedLib(compilerEnv, libFile string, sources ...string) error {
 	} else {
 		args = []string{"-shared", "-Wall", "-Werror", "-o", libFile}
 	}
-
+	if runtime.GOARCH == "386" {
```
hajimehoshi, line 62: If `-m32` indicates a general 32bit architecture, this should be `unsafe.Sizeof(uintptr(0)) == 4`

func.go
```diff
@@ -117,6 +117,10 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 	if cfn == 0 {
 		panic("purego: cfn is nil")
 	}
+	if ty.NumOut() == 1 && (ty.Out(0).Kind() == reflect.Float32 || ty.Out(0).Kind() == reflect.Float64) &&
+		runtime.GOARCH != "arm64" && runtime.GOARCH != "amd64" {
+		panic("purego: float returns are not supported")
```
hajimehoshi, line 122: What about the number of returning value is 2 or more?

hajimehoshi (commented):
Thanks for the hard work! IIUC, this PR solves multiple problems at the same time: #230 and Linux 32bit. What about splitting this PR?

.github/workflows/test.yml
hajimehoshi: Or I prefer

```
env CGO_ENABLED=1 CC=i386-linux-gnu-gcc CXX=i386-linux-gnu-g++ GOOS=linux GOARCH=386 go test -shuffle=on -v -count=10 ./...
```

dlfcn_test.go
TotallyGamerJet: From my understanding it only indicates i386 on amd64 systems. It's listed under the x86_64 options here: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

func.go
TotallyGamerJet: L114 checks for that already

.github/workflows/test.yml
TotallyGamerJet: I believe CC and CXX are defaulted correctly by Go. I grab it in the buildShared function for tests and those test did pass on 386. Most of the code is for install the tools which Go doesn't do

hajimehoshi: I mean we might be able to omit apt-get

hajimehoshi: > I believe CC and CXX are defaulted correctly by Go.

Ah ok that's right. 

Can we still need the apt-get lines?

hajimehoshi (approved):
LGTM, thanks!

PR #235, opened 2024-04-19
internal/fakecgo: clean up gen.go

TotallyGamerJet:
The `go generate` command adds the `generate` build tag when searching for pragmas to run. This means the pragma doesn't need to be in a separate file.

I also moved formatting into the code itself using the `go/format` package and cleaned up the code into a template.


hajimehoshi (commented):
I didn't think including `//go:generate` in a generating script is a conventional way.

I found these patterns in the Go repository

* Use `ignore` (the same as the current way): https://cs.opensource.google/go/go/+/master:src/image/color/palette/generate.go
* Separate a generating script in another directory like `_gen`: https://cs.opensource.google/go/go/+/master:src/cmd/compile/internal/ssa/generate.go
* Use an external command: https://cs.opensource.google/go/go/+/master:src/html/template/context.go

So, I don't think we have to change the way. Rather, I'm worried that now `internal/fakecgo` cannot be compiled with `go build -tags=generate` due to the mix of package names ([`ignore` is conventional](https://pkg.go.dev/cmd/go#hdr-Build_constraints) for such a package name by the way). Thoughts?

TotallyGamerJet:
I don't think many people use `go:build generate` as a normal build tag outside of `go generate` but I don't know how to search all public repos for it.

I can put it back if you want. The only thing is doing it that way means only darwin and freebsd can generate it since the `doc.go` has a build tag. It has one so that it is a compilation error to import `fakecgo` on other platforms. I can add linux though.

hajimehoshi:
Sure, let's put it back to `//go:build ignore`. Thanks,

internal/fakecgo/doc.go
```diff
@@ -30,4 +30,3 @@
 package fakecgo
 
 //go:generate go run gen.go
-//go:generate gofmt -s -w symbols.go
```
hajimehoshi, line 33: Just to confirm: Don't we need this?

hajimehoshi (approved):
LGTM % comments

internal/fakecgo/doc.go
TotallyGamerJet: Correct it's formatted using `go/format` in `gen.go`

Issue #247, opened 2024-05-16
Check `//go:linkname` usage


TotallyGamerJet:
### Operating System

- [X] Windows
- [X] macOS
- [X] Linux
- [X] FreeBSD
- [X] Android
- [X] iOS

### What feature would you like to be added?

https://github.com/golang/go/issues/67401

### Why is this needed?

We want purego to continue to work

hajimehoshi:
For our `runtime.cgocall`, which we pulled, would `syscall_cgocaller` be available instead?

https://cs.opensource.google/go/go/+/master:src/runtime/cgocall.go;l=110;drc=9fa34d9fa25078423cdb484d39cdd62f067098ac?q=runtime%20cgocall&ss=go%2Fgo

TotallyGamerJet:
> For our `runtime.cgocall`, which we pulled, would `syscall_cgocaller` be available instead?

Perhaps for `purego.SyscallN` but not `RegisterFunc` since it needs the entire struct since any type of value could be returned

TotallyGamerJet:
Although this line isn't using `//go:linkname` perhaps we should make it one so that this function stays around?

https://github.com/ebitengine/purego/blob/6cd12240d3323028bec623dc01c4ec3d40340321/internal/fakecgo/asm_amd64.s#L35

Another possible solution is to just prebuild `runtime/cgo` for each GOOS,GOARCH, and GOVERSION pair that purego supports into a `cgo_GOOS_GOARCH_GOVERSION.syso` file. I'm not sure that'll work but it would limit the amount of `//go:linkname` to just  `runtime.cgocall` and means any changes to the `runtime/cgo` package would already be included. The hard part is trying to make the builds reproducible. I remember we tried for something else and eventually gave up.

hajimehoshi:
Of course it is ideal to have reproducible builds, but if there is no other way, we would have to allow such builds...

TotallyGamerJet:
> Of course it is ideal to have reproducible builds, but if there is no other way, we would have to allow such builds...

Are you saying you are in favor of using a prebuilt binary instead of maintaining our own Go port of `runtime/cgo`? 

Of course, I still need to test to make sure this will work

hajimehoshi:
> Are you saying you are in favor of using a prebuilt binary instead of maintaining our own Go port of runtime/cgo?

No, I mean, it is ok-ish to allow non-reproducible binaries if there is no other way. I have no idea which way is better so far.

[...]

PR #250, opened 2024-05-22
Remove linkname on playground

TotallyGamerJet:
Currently the playground just crashes with a nasty error. This makes it so the error message is nice. It also remove linkname usages which is also good.

Updates #247

### DIFF
--- a/dlfcn.go
+++ b/dlfcn.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-//go:build darwin || freebsd || linux
+//go:build darwin || freebsd || (linux && !faketime)
 
 package purego
 
--- a/dlfcn_stubs.s
+++ b/dlfcn_stubs.s
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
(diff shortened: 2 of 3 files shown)


dlfcn.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-//go:build darwin || freebsd || linux
+//go:build darwin || freebsd || (linux && !faketime)
```
hajimehoshi, line 4:
```go
//go:build (darwin || freebsd || linux) && !faketime
```

seems better. We don't have to depend on the fact that GOOS is linux on the playground.

dlfcn_stubs.s
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-//go:build darwin || !cgo && (freebsd ||  linux)
+//go:build darwin || !cgo && (freebsd || linux && !faketime)
```
hajimehoshi, line 4:
```go
//go:build darwin || !cgo && (freebsd || linux) && !faketime
```

hajimehoshi:
```go
//go:build (darwin || !cgo && (freebsd || linux)) && !faketime
```

is another suggestion, but I'm fine with both.

hajimehoshi (approved):
LGTM

PR #253, opened 2024-06-04
Remove some //go:linkname usages

TotallyGamerJet:
Updates #247


func.go
```diff
@@ -340,8 +340,7 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 			// We take the address and then dereference it to trick go vet from creating a possible miss-use of unsafe.Pointer
 			v.SetPointer(*(*unsafe.Pointer)(unsafe.Pointer(&syscall.a1)))
 		case reflect.Ptr:
-			// It is safe to have the address of syscall.r1 not escape because it is immediately dereferenced with .Elem()
```
hajimehoshi, line 343: Is this removing comment related to this PR?

TotallyGamerJet: Yes. It was talking about the use of `noescape` and how that was fine but as `noescape` is no longer being used the comment is not necessary

hajimehoshi (approved):
LGTM

PR #255, opened 2024-06-17
Save R15 before global access

### LINKED ISSUE #254: Dynamic linking error on amd64
### PureGo Version

v0.7.1

### Operating System

- [ ] Windows
- [ ] macOS
- [X] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go version go1.22.4 linux/amd64

Build the following code with `-buildmode=plugin` option. (ref. https://pkg.go.dev/plugin)

```go
package main

import _ "github.com/ebitengine/purego"

func main() {}
```

`purego@v0.5.1` succeeds building, but `purego@v0.6.1` and `purego@v0.7.1` raise the error.

The complete reproduction code is [here](https://github.com/zoncoen-sample/purego-plugin-issue).
[sample error log](https://github.com/zoncoen-sample/purego-plugin-issue/actions/runs/9463761342/job/26069658602)

`go build -buildmode=plugin -o plugin.so .` succeeds without error.

### What happens instead?

```
asm: sys_amd64.s:134: when dynamic linking, R15 is clobbered by a global variable access and is used here: 000[8](https://github.com/zoncoen-sample/purego-plugin-issue/actions/runs/9460766406/job/26060171519#step:4:9)6 (/home/runner/go/pkg/mod/github.com/ebitengine/purego@v0.7.1/sys_amd64.s:134)	MOVQ	R15, 32(SP)
asm: assembly failed
```

_No response_

TotallyGamerJet:
Closes #254

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,6 +46,9 @@ jobs:
           # The option '-gcflags=all=-N -l' is often used at Visual Studio Code.
           # See also https://go.googlesource.com/vscode-go/+/HEAD/docs/debugging.md#launch and the issue hajimehoshi/ebiten#2120.
           go build "-gcflags=all=-N -l" -v ./...
+          
+          # Make sure that plugin build mode works since we save the R15 register (#254)
+          go build -buildmode=plugin ./examples/libc
 
           # Check cross-compiling Windows binaries.
           env GOOS=windows GOARCH=386 go build -v ./...
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -106,7 +106,7 @@ jobs:
           
           echo "=> go build plugin"
           # Make sure that plugin buildmode works since we save the R15 register (#254)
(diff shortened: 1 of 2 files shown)


hajimehoshi (approved):
LGTM, thanks!

.github/workflows/test.yml
```diff
@@ -39,6 +39,12 @@ jobs:
         run: |
           go vet -v ./...
 
+      - name: go build plugin
+        if: runner.os == 'Linux'
```
hajimehoshi, line 43: plugin should work not only on Linux but also macOS and FreeBSD. Is there a reason why we specify Linux here?

TotallyGamerJet: I didn’t think it was necessary since they all behave the same but I can to be safe

hajimehoshi: Let's do this if it is easy.

hajimehoshi (approved):
Still LGTM

Issue #129, opened 2023-04-12
Mac Callback


leaanthony:
Hi there!

I'm trying to hook into Apple's GCD method `dispatch_async` to schedule callbacks on the main thread. I've gotten pretty far but the final call never returns. 

The obj-c code I'm trying to replace is:

```objective-c
	dispatch_async(dispatch_get_main_queue(), ^{
		// some code
	});
```

I discovered that `dispatch_get_main_queue` is actually a macro and returns the address of `_dispatch_main_q`. 
The following is where I'm up to. It feels like I'm missing just one final piece:

```go
func dispatchTest() {
	libSystem, err := purego.Dlopen("libSystem.dylib", purego.RTLD_NOW|purego.RTLD_GLOBAL)
	if err != nil {
		panic(err)
	}
	dispatchMainQueue, err := purego.Dlsym(libSystem, "_dispatch_main_q")
	if err != nil {
		panic(err)
	}
	var dispatch_async func(uintptr, uintptr)
	goFunc := func() {
		fmt.Println("Hello from the main thread")
	}
	callback := purego.NewCallback(goFunc)
	purego.RegisterLibFunc(&dispatch_async, libSystem, "dispatch_async")
	dispatch_async(dispatchMainQueue, callback)
	println("i never get here")
}

func main() {

	// Since the error wasn't checked before check it before calling any code.
	NSApp := NewNSSharedApplication()
	go func() {
		time.Sleep(1 * time.Second)
		dispatchTest()
	}()
	NSApp.Run()
}
```
The NSApp code is just the same as from last ticket so left out to simplify the example.
Nothing is crashing or suggesting anything bad is happening. Am I using `NewCallback` correctly? My guess is that it's to do with C vs Obj-C calling conventions...

Any help appreciated! 

chip: Apple M2 Pro
macOS: 13.2.1 (22D68)

TotallyGamerJet:
So it does crash on my system (M1 Max, 13.3.1).

[...]

PR #261, opened 2024-07-17
Add ObjC Block Support

K1000000:
Updates #129 

 feature

## What this PR does
Adds new type, objc.Block, which is an objc.ID referencing an Objective-C "block" function pointer.

Adds methods to create a Block from a Go function value, get a Go function value from a block, directly invoke a block function, and handle Objective-C memory management (e.g. Copy/Release).

Mitigates pressure on purego Callback limit by relying on the fact the first argument passed to a block implementation is the block itself. This allows for a single callback to handle every block instance that has the same signature, by way of keeping an association between the Go func value and the block instance it was used to create.

### DIFF
--- a/objc/objc_block_darwin.go
+++ b/objc/objc_block_darwin.go
@@ -88,13 +88,11 @@ func newBlockFunctionCache() *blockFunctionCache {
 	return &blockFunctionCache{functions: map[Block]reflect.Value{}}
 }
 
-/*
-blockCache is a thread safe cache of block layouts.
-
-It takes advantage of the block being the first argument of a block call being the block closure,
-only invoking purego.NewCallback() when it encounters a new function type (rather than on for every block creation.)
-This should mitigate block creations putting pressure on the callback limit.
-*/
+// blockCache is a thread safe cache of block layouts.
+//
+// It takes advantage of the block being the first argument of a block call being the block closure,
+// only invoking purego.NewCallback() when it encounters a new function type (rather than on for every block creation.)
+// This should mitigate block creations putting pressure on the callback limit.
 type blockCache struct {
 	sync.Mutex
 	descriptorTemplate blockDescriptor
@@ -140,7 +138,7 @@ func (*blockCache) encode(typ reflect.Type) *uint8 {
 	return &append([]uint8(encoding), 0)[0]
 }
 
-// get layout retrieves a blockLayout VALUE constructed with the supplied function type
+// GetLayout retrieves a blockLayout VALUE constructed with the supplied function type
 // It will panic if the type is not a valid block function.
 func (b *blockCache) GetLayout(typ reflect.Type) blockLayout {
 	b.Lock()
--- a/objc/objc_block_darwin.go
+++ b/objc/objc_block_darwin.go
@@ -1,0 +1,257 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+// SPDX-FileContributor: K1000000
+
+package objc
+
+import (
+	"fmt"
+	"reflect"
+	"sync"
+	"unsafe"
+
+	"github.com/ebitengine/purego"
+)
+
+const (
+	// end-goal of these defaults is to get an Objectve-C memory-managed block object,
+	// that won't try to free() a Go pointer, but will call our custom blockFunctionCache.Delete()
+	// when the reference count drops to zero, so the associated function is also unreferenced.
+
+	// blockBaseClass is the name of the class that block objects will be initialized with.
+	blockBaseClass = "__NSMallocBlock__"
+	// blockFlags is the set of flags that block objects will be initialized with.
+	blockFlags = blockHasCopyDispose | blockHasSignature
+
+	// blockHasCopyDispose is a flag that tells the Objective-C runtime the block exports Copy and/or Dispose helpers.
+	blockHasCopyDispose = 1 << 25
+	// blockHasSignature is a flag that tells the Objective-C runtime the block exports a function signature.
+	blockHasSignature = 1 << 30
+)
+
+// blockDescriptor is the Go representation of an Objective-C block descriptor.
+// It is a component to be referenced by blockDescriptor.
+type blockDescriptor struct {
+	_         uintptr
+	Size      uintptr
+	_         uintptr
+	Dispose   uintptr
+	Signature *uint8
+}
+
+// blockLayout is the Go representation of the structure abstracted by a block pointer.
+// From the Objective-C point of view, a pointer to this struct is equivalent to an ID that
+// references a block.
+type blockLayout struct {
+	Isa        Class
+	Flags      uint32
+	_          uint32
+	Invoke     uintptr
+	Descriptor *blockDescriptor
+}
+
+/*
+blockCache is a thread safe cache of block layouts.
+
+The function closures themselves are kept alive by caching them internally until the Objective-C runtime indicates that
+they can be released (presumably when the reference count reaches zero.) This approach is used instead of appending the function
+object to the block allocation, where it is out of the visible domain of Go's GC.
+*/
+type blockFunctionCache struct {
+	mutex     sync.RWMutex
+	functions map[Block]reflect.Value
+}
+
+// Load retrieves a function (in the form of a reflect.Value, so Call can be invoked) associated with the key Block.
+func (b *blockFunctionCache) Load(key Block) reflect.Value {
+	b.mutex.RLock()
+	defer b.mutex.RUnlock()
+	return b.functions[key]
+}
+
+// Store associates a function (in the form of a reflect.Value) with the key Block.
+func (b *blockFunctionCache) Store(key Block, value reflect.Value) Block {
+	b.mutex.Lock()
+	defer b.mutex.Unlock()
+	b.functions[key] = value
+	return key
+}
+
+// Delete removed the function associated with the key Block.
+func (b *blockFunctionCache) Delete(key Block) {
+	b.mutex.Lock()
+	defer b.mutex.Unlock()
+	delete(b.functions, key)
+}
+
+// newBlockFunctionCache initilizes a new blockFunctionCache
+func newBlockFunctionCache() *blockFunctionCache {
+	return &blockFunctionCache{functions: map[Block]reflect.Value{}}
+}
+
+/*
+blockCache is a thread safe cache of block layouts.
+
+It takes advantage of the block being the first argument of a block call being the block closure,
+only invoking purego.NewCallback() when it encounters a new function type (rather than on for every block creation.)
+This should mitigate block creations putting pressure on the callback limit.
+*/
+type blockCache struct {
+	sync.Mutex
+	descriptorTemplate blockDescriptor
+	layoutTemplate     blockLayout
+	layouts            map[reflect.Type]blockLayout
+	Functions          *blockFunctionCache
+}
+
+// encode returns a blocks type as if it was given to @encode(typ)
+func (*blockCache) encode(typ reflect.Type) *uint8 {
+	// this algorithm was copied from encodeFunc,
+	// but altered to panic on error, and to only accep a block-type signature.
+	if (typ == nil) || (typ.Kind() != reflect.Func) {
+		panic("objc: not a function")
+	}
+
+	var encoding string
+	switch typ.NumOut() {
+	case 0:
+		encoding = encVoid
+	default:
+		returnType, err := encodeType(typ.Out(0), false)
+		if err != nil {
+			panic(fmt.Sprintf("objc: %v", err))
+		}
+		encoding = returnType
+	}
+
+	if (typ.NumIn() == 0) || (typ.In(0) != reflect.TypeOf(Block(0))) {
+		panic(fmt.Sprintf("objc: A Block implementation must take a Block as its first argument; got %v", typ.String()))
+	}
+
+	encoding += encId
+	for i := 1; i < typ.NumIn(); i++ {
+		argType, err := encodeType(typ.In(i), false)
+		if err != nil {
+			panic(fmt.Sprintf("objc: %v", err))
+		}
+		encoding = fmt.Sprint(encoding, argType)
+	}
+
+	// return the encoding as a C-style string.
+	return &append([]uint8(encoding), 0)[0]
+}
+
+// get layout retrieves a blockLayout VALUE constructed with the supplied function type
+// It will panic if the type is not a valid block function.
+func (b *blockCache) GetLayout(typ reflect.Type) blockLayout {
+	b.Lock()
+	defer b.Unlock()
+
+	// return the cached layout, if it exists.
+	if layout, ok := b.layouts[typ]; ok {
+		return layout
+	}
+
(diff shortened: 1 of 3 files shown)


objc/objc_block_darwin.go
```diff
@@ -0,0 +1,257 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
```
hajimehoshi, line 2: 2024

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,257 @@
...
+// SPDX-FileContributor: K1000000
```
hajimehoshi, line 3: Do not add `// SPDX-FileContributor:`

objc/objc_block_darwin_test.go
```diff
@@ -0,0 +1,197 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+// SPDX-FileContributor: K1000000
```
hajimehoshi, line 3: ditto

objc/objc_runtime_darwin.go
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+// SPDX-FileContributor: Modified by K1000000
```
hajimehoshi, line 3: ditto

objc/objc_block_darwin.go
K1000000: OK Thank you. Should be fixed now.

objc/objc_block_darwin.go
K1000000: (ditto)

objc/objc_block_darwin_test.go
K1000000: (ditto)

objc/objc_runtime_darwin.go
K1000000: (ditto)

hajimehoshi:
Please address on test failures

objc/objc_block_darwin_test.go
```diff
@@ -0,0 +1,182 @@
...
+		argumentValues = append(argumentValues, reflect.Value{})
+		argumentTypes = append(argumentTypes, nil)
+		for index := 0; index < len(values); index++ {
+			argumentValues[len(argumentValues)-1] = values[index]
+			argumentTypes[len(argumentTypes)-1] = values[index].Type()
+			argumentRecurse(argumentValues, argumentTypes, execute)
+		}
+	}
+
+	for out := 0; out <= len(values); out++ {
```
hajimehoshi, line 101: `for out := range values`

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,256 @@
...
+	return &blockFunctionCache{functions: map[Block]reflect.Value{}}
+}
+
+/*
+blockCache is a thread safe cache of block layouts.
+
+It takes advantage of the block being the first argument of a block call being the block closure,
+only invoking purego.NewCallback() when it encounters a new function type (rather than on for every block creation.)
+This should mitigate block creations putting pressure on the callback limit.
+*/
```
hajimehoshi, line 97: Use `//` comments for consistency.

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,256 @@
...
+			panic(fmt.Sprintf("objc: %v", err))
+		}
+		encoding = fmt.Sprint(encoding, argType)
+	}
+
+	// return the encoding as a C-style string.
+	return &append([]uint8(encoding), 0)[0]
+}
+
+// get layout retrieves a blockLayout VALUE constructed with the supplied function type
```
hajimehoshi, line 143: `// GetLayout retrieves ...`

hajimehoshi (commented):
@TotallyGamerJet Do you have any insights?

objc/objc_block_darwin_test.go
K1000000: Sorry, this is not equivalent. Notice `<=`, which means this count wants to run one time outside the range.  This is used later to create "doesn't return anything" test cases. (But thank you.)

hajimehoshi: Oops sorry, you are right.

objc/objc_block_darwin.go
K1000000: Corrected. Thank you.

objc/objc_block_darwin.go
K1000000: Corrected. Thank you.

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,254 @@
...
+
+	// return the cached layout, if it exists.
+	if layout, ok := b.layouts[typ]; ok {
+		return layout
+	}
+
+	// otherwise: create a layout, and populate it with the default templates
+	layout := b.layoutTemplate
+	layout.Descriptor = &blockDescriptor{}
+	(*layout.Descriptor) = b.descriptorTemplate
```
TotallyGamerJet, line 155: nitpick: these parenthesis aren't necessary

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,254 @@
...
+// GetImplementation populates a function pointer with the implementation of a Block.
+// Function will panic if the Block is not kept alive while it is in use
+// (possibly by using Block.Copy()).
+func (b Block) GetImplementation(fptr any) {
+	// there is a runtime function imp_implementationWithBlock that could have been used instead,
+	// but experimentation has shown the returned implementation doesn't actually work as expected.
+	// also, it creates a new copy of the block which must be freed independently,
+	// which would have made this implementation more complicated than necessary.
+	// we know a block ID is actually a pointer to a blockLayout struct, so we'll take advantage of that.
+	if b != 0 {
```
TotallyGamerJet, line 216: I think it's better to write this as:

```
if b == 0 {
    return
}
if cfn := (*(**blockLayout)(unsafe.Pointer(&b))).Invoke; cfn == 0 {
    return
}
purego.RegisterFunc(fptr, cfn)
```

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	cache.descriptorTemplate.Dispose = purego.NewCallback(cache.Functions.Delete)
+	return cache
+}
+
+// blocks is the global block cache
+var blocks *blockCache
+
+// Block is an opaque pointer to an Objective-C object containing a function with its associated closure.
+type Block ID
```
TotallyGamerJet, line 199: What's the reason for keeping this as an opaque pointer instead of just *blockLayout?

objc/objc_block_darwin_test.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	var count = 0
+	block := objc.NewBlock(
+		func(block objc.Block, line objc.ID, stop *bool) {
+			count++
+			fmt.Printf("LINE %d: %s\n", count, objc.Send[string](line, objc.RegisterName("UTF8String")))
+			(*stop) = (count == 3)
```
TotallyGamerJet, line 27: nitpick: neither of these parentheses are necessary

objc/objc_block_darwin_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"reflect"
+	"testing"
+	"unsafe"
+
+	"github.com/ebitengine/purego"
+	"github.com/ebitengine/purego/objc"
+)
+
+func ExampleNewBlock() {
+	_, err := purego.Dlopen("/System/Library/Frameworks/Foundation.framework/Foundation", purego.RTLD_GLOBAL)
```
TotallyGamerJet, line 17: Although it isn't necessary it's good to add `purego.RTLD_NOW|` as that's the proper way to call dlopen

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,254 @@
...
+// Copy creates a copy of a block on the Objective-C heap (or increments the reference count if already on the heap.)
+// Use Block.Release() to free the copy when it is no longer in use.
+func (b Block) Copy() Block {
+	return _Block_copy(b)
+}
+
+// GetImplementation populates a function pointer with the implementation of a Block.
+// Function will panic if the Block is not kept alive while it is in use
+// (possibly by using Block.Copy()).
+func (b Block) GetImplementation(fptr any) {
```
TotallyGamerJet, line 210: Why should this be exported?

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,254 @@
...
+		encoding = fmt.Sprint(encoding, argType)
+	}
+
+	// return the encoding as a C-style string.
+	return &append([]uint8(encoding), 0)[0]
+}
+
+// GetLayout retrieves a blockLayout VALUE constructed with the supplied function type
+// It will panic if the type is not a valid block function.
+func (b *blockCache) GetLayout(typ reflect.Type) blockLayout {
```
TotallyGamerJet, line 143: This shouldn't be exported as it is a method on an unexpected type

objc/objc_block_darwin.go
```diff
@@ -0,0 +1,254 @@
...
+	descriptorTemplate blockDescriptor
+	layoutTemplate     blockLayout
+	layouts            map[reflect.Type]blockLayout
+	Functions          *blockFunctionCache
+}
+
+// encode returns a blocks type as if it was given to @encode(typ)
+func (*blockCache) encode(typ reflect.Type) *uint8 {
+	// this algorithm was copied from encodeFunc,
+	// but altered to panic on error, and to only accep a block-type signature.
```
TotallyGamerJet, line 107: I would just do this check first and then call the encodeFunc instead of copying the code.

```
if type.Kind() == reflect.Func && ((typ.NumIn() == 0) || (typ.In(0) != reflect.TypeOf(Block(0)))) {
		panic(fmt.Sprintf("objc: A Block implementation must take a Block as its first argument; got %v", typ.String()))
	}
encoding, err := encodeFunc(fn)
if err != nil {
   panic(fmt.Sprintf("objc: failed to encode Block %s", err))
}
``

**Review by TotallyGamerJet — CHANGES_REQUESTED** [review]

File: objc/objc_block_darwin.go (hunk 10 above)
Comment (K1000000) [reply objc/objc_block_darwin.go:199]: A "Block" in Objective-C is  first-class object (e.g. can be cast to an ID and sent messages like any other object.) The problem is, ID in purego's objc implementation has a base of `uintptr,` so defining a block as a pointer type would complicate down-casting.  I believe the choice to define ID as `uintptr` instead of `*struct{...}` (which would better align to what it actually is in Objective-C) was to allow receiver methods to be defined (since go does not allow receiver methods on pointer type aliases), but that's just speculation/intuition.

File: objc/objc_block_darwin_test.go (hunk 12 above)
Comment (K1000000) [reply objc/objc_block_darwin_test.go:17]: Okay, will fix. I actually copied the syntax from the existing load of  `libobjc.dylib`.

File: objc/objc_block_darwin.go (hunk 13 above)
Comment (K1000000) [reply objc/objc_block_darwin.go:210]: I wasn't actually happy with it either.  It was meant to provide provide synergy with purego's RegisterFunc in the block case, but has potential for misuse because of blocks having managed lifetimes.

File: objc/objc_block_darwin.go (hunk 15 above)
Comment (K1000000) [reply objc/objc_block_darwin.go:107]: The problem is, encodeFunc (as it stands) has a check that enforces an Objective-C method signature, which is not the same as a block signature. Maybe worth splitting up that functon from it's check first, and having both implementations share?

File: objc/objc_block_darwin_test.go (hunk 12 above)
Comment (TotallyGamerJet) [reply objc/objc_block_darwin_test.go:17]: Please fix it there too

File: objc/objc_block_darwin.go (hunk 13 above)
Comment (TotallyGamerJet) [reply objc/objc_block_darwin.go:210]: If someone wants to call a Block directly can't they just call RegisterFunc? It's just a C function with a objc.Block as the first argument current?

File: objc/objc_block_darwin.go (hunk 10 above)
Comment (TotallyGamerJet) [reply objc/objc_block_darwin.go:199]: Are people going to actually cast Block to ID? Block doesn't have any methods does it?

Of course if a class extended Block then you'd probably want ID but then are you using it as a Block still?

Perhaps a Block.ID() method and BlockFromID() function that does the unsafe cast? Perhaps that can be added later as I'm not confident people need to use a Block as an ID

File: objc/objc_block_darwin.go (hunk 15 above)
Comment (TotallyGamerJet) [reply objc/objc_block_darwin.go:107]: If you think you can find a good solution go for it otherwise it might be ok to just leave it how you have it now

File: objc/objc_block_darwin.go [hunk 16]
```diff
@@ -0,0 +1,254 @@
...
+	Descriptor *blockDescriptor
+}
+
+/*
+blockCache is a thread safe cache of block layouts.
+
+The function closures themselves are kept alive by caching them internally until the Objective-C runtime indicates that
+they can be released (presumably when the reference count reaches zero.) This approach is used instead of appending the function
+object to the block allocation, where it is out of the visible domain of Go's GC.
+*/
```
hajimehoshi, line 58: Use `//` style comments for consistency.

objc/objc_block_darwin.go
K1000000: Indeed. Perhaps then, just replace this method with `Implementation()` (to treat it like a read-only property) that returns a uintptr, and allow the user to call RegisterFunc() at their own discretion (keeping in mind, the lifetime caveats, of course?), with the option to use just use an `Invoke()` or `BlockInvoke[]()` that have already been implemented with lifetime-safety in mind?

objc/objc_block_darwin.go
K1000000: Will do. Thank you.

objc/objc_block_darwin.go
K1000000: It does have methods (most importantly `Copy()` and `Release()` for lifetime management), but taking your thoughts into account, this could be refactored to behave like `Protocol`, which is used as a struct pointer, even though it, too, is actually a first-class object?

objc/objc_block_darwin.go
K1000000: WIll fix. Note that the `cfn  :=` clause would need to be pulled out of the `if` block to actually work, but simple enough.

objc/objc_block_darwin.go
TotallyGamerJet: I mean there are no objc methods that users are trying to call.

> this could be refactored to behave like Protocol

Yes that's what I'm thinking

objc/objc_block_darwin.go
TotallyGamerJet: Do you have a use case where grabbing the raw uintptr to call directly is helpful? I want to limit the exposed API to only the necessary parts

irootyou:
This is for passing struct arguments that are not pointers?

TotallyGamerJet:
No this is for objc callbacks which are called Blocks

hajimehoshi:
Let me close this PR as there is no recent progress.

PR #263, opened 2024-08-02
Call C dlfcn functions directly on android

### LINKED ISSUE #262: purego.Dlopen fails on Galaxy A02s
### PureGo Version

v0.8.0-alpha.3

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [X] Android
- [ ] iOS

### Go Version (`go version`)

go version go1.22.5 darwin/arm64

Call `purego.Dlopen("libGLESv2.so", purego.RTLD_LAZY|purego.RTLD_GLOBAL)` on Android

An so file is loaded

### What happens instead?

Error:

```
invalid flags to dlopen: 101
```

[Galaxy A02s has 32bits Android OS though the CPU is 64bit](
https://community.infiniteflight.com/t/i-cannot-download-if-because-its-not-compatible/710435/4). Also, Android has [different definitions for `RTLD_*` for 32bits](https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/dlfcn.h).

[Even though I used C's `RTLD_*`, another crash happened](https://github.com/hajimehoshi/ebiten/issues/3052#issuecomment-2261184838). I am not sure when this happened. Maybe loading a symbol?

See also https://github.com/hajimehoshi/ebiten/issues/3052

TotallyGamerJet:
Closes #262

bug

## What this PR does | solves
Before we were link naming to the C symbols and hard coding the C constants on android. This PR calls the C functions directly.

### DIFF
--- a/dlfcn.go
+++ b/dlfcn.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-//go:build (darwin || freebsd || linux) && !faketime
+//go:build (darwin || freebsd || linux) && !faketime && !android
 
 package purego
 
--- a/dlfcn_android.go
+++ b/dlfcn_android.go
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package purego
+
+import "github.com/ebitengine/purego/internal/cgo"
+
+var (
+	RTLD_DEFAULT = cgo.RTLD_DEFAULT
+	RTLD_LAZY    = cgo.RTLD_LAZY
+	RTLD_NOW     = cgo.RTLD_NOW
+	RTLD_LOCAL   = cgo.RTLD_LOCAL
+	RTLD_GLOBAL  = cgo.RTLD_GLOBAL
+)
+
+func Dlopen(path string, mode int) (uintptr, error) {
+	return cgo.Dlopen(path, mode)
+}
+
+func Dlsym(handle uintptr, name string) (uintptr, error) {
+	return cgo.Dlsym(handle, name)
+}
+
+func Dlclose(handle uintptr) error {
+	return cgo.Dlclose(handle)
+}
+
+func loadSymbol(handle uintptr, name string) (uintptr, error) {
+	return Dlsym(handle, name)
+}
--- a/dlfcn_android.go
+++ b/dlfcn_android.go
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+// SPDX-FileCopyrightText: 2024 The Ebitengine Authors
 
 package purego
 
--- a/dlfcn_linux.go
+++ b/dlfcn_linux.go
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
+//go:build !android
+
 package purego
 
 // Source for constants: https://codebrowser.dev/glibc/glibc/bits/dlfcn.h.html
--- a/internal/cgo/dlfcn_cgo_unix.go
+++ b/internal/cgo/dlfcn_cgo_unix.go
@@ -23,11 +23,11 @@ import (
 )
 
 var (
-	RTLD_DEFAULT int = int(uintptr(C.RTLD_DEFAULT))
-	RTLD_LAZY    int = int(uintptr(C.RTLD_LAZY))
-	RTLD_NOW     int = int(uintptr(C.RTLD_NOW))
-	RTLD_LOCAL   int = int(uintptr(C.RTLD_LOCAL))
-	RTLD_GLOBAL  int = int(uintptr(C.RTLD_GLOBAL))
+	RTLD_DEFAULT = int(uintptr(C.RTLD_DEFAULT))
+	RTLD_LAZY    = int(uintptr(C.RTLD_LAZY))
+	RTLD_NOW     = int(uintptr(C.RTLD_NOW))
+	RTLD_LOCAL   = int(uintptr(C.RTLD_LOCAL))
+	RTLD_GLOBAL  = int(uintptr(C.RTLD_GLOBAL))
 )
 
 func Dlopen(filename string, flag int) (uintptr, error) {
@@ -43,7 +43,7 @@ func Dlopen(filename string, flag int) (uintptr, error) {
 func Dlsym(handle uintptr, symbol string) (uintptr, error) {
 	csymbol := C.CString(symbol)
 	defer C.free(unsafe.Pointer(csymbol))
-	symbolAddr := C.dlsym(unsafe.Pointer(handle), csymbol)
+	symbolAddr := C.dlsym(*(*unsafe.Pointer)(unsafe.Pointer(&handle)), csymbol)
 	if symbolAddr == nil {
 		return 0, errors.New(C.GoString(C.dlerror()))
 	}
@@ -51,7 +51,7 @@ func Dlsym(handle uintptr, symbol string) (uintptr, error) {
 }
 
 func Dlclose(handle uintptr) error {
-	result := C.dlclose(unsafe.Pointer(handle))
+	result := C.dlclose(*(*unsafe.Pointer)(unsafe.Pointer(&handle)))
 	if result != 0 {
 		return errors.New(C.GoString(C.dlerror()))
 	}
--- a/internal/cgo/dlfcn_cgo_unix.go
+++ b/internal/cgo/dlfcn_cgo_unix.go
@@ -12,11 +12,43 @@ package cgo
 */
 import "C"
 
-// all that is needed is to assign each dl function because then its
-// symbol will then be made available to the linker and linked to inside dlfcn.go
-var (
-	_ = C.dlopen
-	_ = C.dlsym
-	_ = C.dlerror
-	_ = C.dlclose
+import (
+	"errors"
+	"unsafe"
 )
+
+var (
+	RTLD_DEFAULT = C.RTLD_DEFAULT
+	RTLD_LAZY    = C.RTLD_LAZY
+	RTLD_NOW     = C.RTLD_NOW
+	RTLD_LOCAL   = C.RTLD_LOCAL
+	RTLD_GLOBAL  = C.RTLD_GLOBAL
+)
+
+func Dlopen(filename string, flag int) (uintptr, error) {
+	cfilename := C.CString(filename)
+	defer C.free(unsafe.Pointer(cfilename))
+	handle := C.dlopen(cfilename, C.int(flag))
+	if handle == nil {
+		return 0, errors.New(C.GoString(C.dlerror()))
+	}
+	return handle, nil
+}
+
+func Dlsym(handle uintptr, symbol string) (uintptr, error) {
+	csymbol := C.CString(symbol)
+	defer C.free(unsafe.Pointer(csymbol))
+	symbolAddr := C.dlsym(handle, csymbol)
+	if symbolAddr == nil {
+		return 0, errors.New(C.GoString(C.dlerror()))
+	}
+	return symbolAddr, nil
+}
+
+func Dlclose(handle uintptr) error {
+	result := C.dlclose(handle)
+	if result != 0 {
+		return errors.New(C.GoString(C.dlerror()))
+	}
+	return nil
+}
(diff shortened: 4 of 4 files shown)


TotallyGamerJet:
So to test this rollback https://github.com/hajimehoshi/ebiten/commit/297efea68b3f689687c134bd9f30fc1d9a7b91c0 and then test in Test Lab using a purego that has this PR in it.

dlfcn_android.go
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2024 The Ebitengine Authors
+
+package purego
+
+import "github.com/ebitengine/purego/internal/cgo"
+
+var (
```
hajimehoshi, line 8: If we use var here, we should also change other platforms to var for consistency. What do you think?

TotallyGamerJet: I prefer them being constants since it makes it clear they aren't supposed to be modified. Android is just "unique"

hajimehoshi: So this code would no longer be portable, though we would not do that

```go
var _ [purego.RTLD_GLOBAL]int
```

TotallyGamerJet: Sure but why would anyone do that? Making it var breaks compatibility for all the people who do currently do that instead of just on android which is likely less common anyways

hajimehoshi: >  Making it var breaks compatibility for all the people who do currently do that instead of just on android which is likely less common anyways

The version is still 0.x so it is still allowed to break compatibility.

Another example would be:

```go
_ = math.Abs(purego.RTLD_GLOBAL) // This works only when RTLD_GLOBAL is an untyped constant.
```

```go
type T int

func main() {
    var _ T = purego.RTLD_GLOBAL // This works only when RTLD_GLOBAL is an untyped constant.
}
```

At least, if we keep constants, we should add types (int) instead of untyped constants, so that the both cases would be portable and no longer be compiled.

TotallyGamerJet: Agreed they shouldn't be untype constants but constant int

TotallyGamerJet: Hmmm... This seems to break more things...

```
./dlfcn_test.go:23:28: cannot use purego.RTLD_DEFAULT (constant -2 of type int) as uintptr value in argument to purego.Dlsym
./dlfcn_test.go:91:31: cannot use purego.RTLD_DEFAULT (constant -2 of type int) as uintptr value in argument to purego.Dlsym
./dlfcn_test.go:94:40: cannot use purego.RTLD_DEFAULT (constant -2 of type int) as uintptr value in argument to purego.SyscallN
```

It's because Dlopen takes an int but Dlysm and SyscallN take uintptr. Not sure which to change since it currently matches the C header (int -> int, void* -> uintptr)

hajimehoshi: What about uintptr constants (and variables)? Wouldn't -2 work?

hajimehoshi: https://github.com/ebitengine/purego/blob/main/dlfcn_freebsd.go#L8

Yeah I feel like we should define uintptr constants (and variables)

hajimehoshi: And we would have to change `mode int` to `mode uintptr`, maybe...?

https://pkg.go.dev/github.com/ebitengine/purego#Dlopen

TotallyGamerJet: What do you mean constants and variables? Only RTLD_DEFAULT needs to uintptr. Wasn't it the case that was the only one on android that wasn't const so perhaps?

```
var
RTLD_DEFAULT = cgo.RTLD_DEFAULT
)
const(
	RTLD_LAZY    = cgo.RTLD_LAZY
	RTLD_NOW     = cgo.RTLD_NOW
	RTLD_LOCAL   = cgo.RTLD_LOCAL
	RTLD_GLOBAL  = cgo.RTLD_GLOBAL
)

```

hajimehoshi: Hm? Can we use consts for cgo values?

TotallyGamerJet: Yes as long as it's value is constant which it is for all of those except RTLD_DEFAULT

hajimehoshi: By the way, for RTLD_*, what about defining values by ourselves form [the header](https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/dlfcn.h)? I mean

```go
//go:build android && (386 || arm)

const (
	RTLD_DEFAULT = 0xffffffff
	RTLD_LAZY    = 0x00000001
	RTLD_NOW     = 0x00000000
	RTLD_LOCAL   = 0x00000000
	RTLD_GLOBAL  = 0x00000002
)
```

hajimehoshi: > Yes as long as it's value is constant which it is for all of those except RTLD_DEFAULT [at least on linux]

Interesting. So only `cgo.RTLD_DEFAULT` cannot be a constant, right?

TotallyGamerJet: Yes because it's being casted in the header to a `void*` although its value doesn't change

hajimehoshi:
The test has passed!
image

dlfcn_android.go
TotallyGamerJet: > By the way, for RTLD_*, what about defining values by ourselves form the header? I mean

I would, but I thought we determined that that doesn't always work on android. Am I mistaken?

hajimehoshi: > I thought we determined that that doesn't always work on android. Am I mistaken?

My understanding is that `purego.Dlopen` succeeded with the constants for 32bit Android, but a mysterious crash happens after that. The crash should be fixed by using C's dl functions directly by this PR.

TotallyGamerJet: Let's try with the constants hardcoded in dlfcn_android.go

hajimehoshi: Thanks!

TotallyGamerJet: Not sure if my approach is the best since now there are 3 extra files but the easiest thing I could come up with so we can test it

dlfcn_android_32bit.go
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2024 The Ebitengine Authors
+
+//go:build android && (386 || arm)
+
+package purego
+
```
hajimehoshi, line 7: Let's add a comment that these values are from https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/dlfcn.h

dlfcn_android_64bit.go
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2024 The Ebitengine Authors
+
+//go:build android && (amd64 || arm64)
+
+package purego
+
```
hajimehoshi, line 7: Ditto

dlfcn_android.go
hajimehoshi: We can reduce one file by merging `dlfcn_linux.go` and `dlfcn_android_64bit.go`, but I am not sure which is better.

hajimehoshi:
347a5810c121da4a1b667b199c22963008d9fdf3 also passed the tests (i.e. the hardcoded values worked!)

dlfcn_freebsd.go
```diff
@@ -5,9 +5,9 @@ package purego
 
 // Constants as defined in https://github.com/freebsd/freebsd-src/blob/main/include/dlfcn.h
 const (
-	RTLD_DEFAULT = ^uintptr(0) - 2 // Pseudo-handle for dlsym so search for any loaded symbol
-	RTLD_LAZY    = 0x00001         // Relocations are performed at an implementation-dependent time.
-	RTLD_NOW     = 0x00002         // Relocations are performed when the object is loaded.
-	RTLD_LOCAL   = 0x00000         // All symbols are not made available for relocation processing by other modules.
-	RTLD_GLOBAL  = 0x00100         // All symbols are available for relocation processing of other modules.
+	RTLD_DEFAULT = 0xffffffff_fffffffd // Pseudo-handle for dlsym so search for any loaded symbol
```
hajimehoshi, line 8: Does this really work with 32bit machines?

We can get an untyped 32 or 64 by https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/math/const.go;l=40

dlfcn_android_64bit.go
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2024 The Ebitengine Authors
+
+//go:build android && (amd64 || arm64)
```
hajimehoshi, line 4: I slightly prefer `android && !386 && !arm`

Optional: another option is integrate `dlfnc_android_32bit.go` and `dlfnc_android_64bit.go` and use a little complex calculation like https://cs.opensource.google/go/go/+/master:src/runtime/malloc.go;l=210:

```go
const (
    _64bit = 1 << (^uintptr(0) >> 63) / 2
    RTLD_DEFAULT = (1 - _64bit) * 0xffffffff
    RTLD_LAZY = 0x00000001
    RTLD_NOW = _64bit * 0x00002
    RTLD_LOCAL = 0x00000000
    RTLD_GLOBAL = _64bit * 0x00100 + (1 - _64bit) * 0x00000002
)
```

dlfcn.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-//go:build (darwin || freebsd || linux) && !faketime
+//go:build (darwin || freebsd || linux) && !faketime && !android
```
hajimehoshi, line 4: `//go:build (darwin || freebsd || linux) && !android && !faketime`

Let's keep alphabetical order.

dlfcn_freebsd.go
```diff
@@ -5,9 +5,10 @@ package purego
 
 // Constants as defined in https://github.com/freebsd/freebsd-src/blob/main/include/dlfcn.h
 const (
-	RTLD_DEFAULT = 0xffffffff_fffffffd // Pseudo-handle for dlsym so search for any loaded symbol
-	RTLD_LAZY    = 0x00000001          // Relocations are performed at an implementation-dependent time.
-	RTLD_NOW     = 0x00000002          // Relocations are performed when the object is loaded.
-	RTLD_LOCAL   = 0x00000000          // All symbols are not made available for relocation processing by other modules.
-	RTLD_GLOBAL  = 0x00000100          // All symbols are available for relocation processing of other modules.
+	intSize      = 32 << (^uint(0) >> 63) // 32 or 64
+	RTLD_DEFAULT = 1<<intSize - 3         // Pseudo-handle for dlsym so search for any loaded symbol
```
hajimehoshi, line 9: `1<<intSize - 2`?

TotallyGamerJet: 3 is correct. One lower than before because the first minus sets the bits to all 1s then -2 to clear the lower to bits

https://go.dev/play/p/7qQCqEPtFFz

TotallyGamerJet: Perhaps should be written as `(1<<intSize - 1) -2`?

hajimehoshi: https://github.com/freebsd/freebsd-src/blob/main/include/dlfcn.h#L64

Isn't `((void *) -2)` `0xfffffffe` (in 32bit)?

TotallyGamerJet: You are correct. I guess the constant was wrong before then?

dlfcn_darwin.go
```diff
@@ -6,11 +6,11 @@ package purego
 // Source for constants: https://opensource.apple.com/source/dyld/dyld-360.14/include/dlfcn.h.auto.html
 
 const (
-	RTLD_DEFAULT = ^uintptr(0) - 1 // Pseudo-handle for dlsym so search for any loaded symbol
-	RTLD_LAZY    = 0x1             // Relocations are performed at an implementation-dependent time.
-	RTLD_NOW     = 0x2             // Relocations are performed when the object is loaded.
-	RTLD_LOCAL   = 0x4             // All symbols are not made available for relocation processing by other modules.
-	RTLD_GLOBAL  = 0x8             // All symbols are available for relocation processing of other modules.
+	RTLD_DEFAULT = 0xffffffff_fffffffe // Pseudo-handle for dlsym so search for any loaded symbol
```
hajimehoshi, line 9: `1 << 64 - 2` might be better

hajimehoshi (commented):
I'll test this with 32bit Android again

dlfcn_freebsd.go
hajimehoshi: Yes, unfortunately...

hajimehoshi:
Building failed on Android

```
/Users/hajimehoshi/purego/func.go:155:8: non-boolean condition in if statement
/Users/hajimehoshi/purego/func.go:420:7: is32bit redeclared in this block
        /Users/hajimehoshi/purego/dlfcn_android.go:12:2: other declaration of is32bit

exit status 1
make: *** [buildAndroidRuntime] Error 1
```

hajimehoshi (approved):
I've confirmed this worked with Android 32bit and 64bit. LGTM, thanks!

Issue #270, opened 2024-08-22
Remove dependency on `golang.org/x/sys`


david50407:
### Operating System

- [X] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

### What feature would you like to be added?

Remove dependency on `golang.org/x/sys`. Aim to be zero dependencies.

### Why is this needed?

> The discussion was started from Discord chat: https://discord.com/channels/842049801528016967/1123106378731487345/1275998637826248744

For now, the only dependency is `golang.org/x/sys`.

If we can remove the dependency, we can embed purego into more situation like Golang toolchain.

Currently I'm working on a go fork and try to introduce our own C lib into Go toolchain, if purego can have zero dependencies, we can import the C lib without CGO (instead, use purego for sure). But in the toolchain, dependency chain has limitation that `golang.org/x/sys` is not allowed to be here. :(

hajimehoshi:
@david50407 Please try `5905af50730466425f56454a0491ae991d644beb`

texadactyl:
My partner (@platypusguy) and I are working on a large and complex Go project that is being amended to access JVM library functions. It runs on Linux, Unix, MacOS, and Windows. It would be very helpful if we only had to deal with a single generic purego API set (Dlopen, Dlclose, etc.). 

Is a single purego API set one of the goals of this issue?

hajimehoshi:
> It would be very helpful if we only had to deal with a single generic purego API set (Dlopen, Dlclose, etc.).

This is a different topic from this issue. The answer is no, since Windows doesn't have dl* functions.

TotallyGamerJet:
> Is a single purego API set one of the goals of this issue?

For the explanation why see #107

texadactyl:
I am in agreement with the points made by both @RomainMuller and @platypusguy in PR #192. APIs should not burden the user with O/S specifics wherever possible.

Reference: https://pkg.go.dev/github.com/ebitengine/purego

"Supported Platforms" states that the APIs are supported on the listed platforms with the exception of a ARM32 and Windows-32 environments. However, the "Example" states that the APIs only work on POSIX systems contradicts the "Motivation" and "Supported Platforms" sections. 

It would be helpful if an equivalent example for Windows was provided. Then, users would have some guidance without having to engage in significant trial-and-error activity.

[...]

PR #271, opened 2024-08-22
all: remove dependencies on golang.org/x/sys

hajimehoshi:
The example still has the dependency on golang.org/x/sys.

Updates #270

#270 

feature

## What this PR does | solves
Removes the dependency on golang.org/x/sys from the core.

The example still uses golang.org/x/sys.


TotallyGamerJet (approved):
LGTM

PR #272, opened 2024-08-23
remove golang.org/x/sys from examples and therefore completely #270

JupiterRider:
#270 

feature

## What this PR does | solves
Removes golang.org/x/sys  from the examples as well.


hajimehoshi:
Removing the dependency from the example doesn't make sense.

PR #274, opened 2024-08-25
Clarify Example section and Dl* windows equivalents

### LINKED ISSUE #273: Generic APIs are needed for multiple platforms (or documentation of differences with examples)
### Operating System

- [X] Windows
- [X] macOS
- [X] Linux
- [X] FreeBSD
- [X] Android
- [X] iOS

I am in agreement with the points made by both @RomainMuller and @platypusguy in PR https://github.com/ebitengine/purego/pull/192. APIs should not burden the user with O/S specifics wherever possible. The precise API names are not so important as long as they are universal.

However, at least, the project documentation needs an update.

Reference: https://pkg.go.dev/github.com/ebitengine/purego

"Supported Platforms" states that the APIs are supported on the listed platforms with the exception of a ARM32 and Windows-32 environments. However, the "Example" states that the APIs only work on POSIX systems contradicts the previous documentation sections.

It would be helpful if an equivalent documentation example for Windows was provided.  Also, under each API where Windows is not currently supported (E.g. Dlopen), depict the equivalent operation for Windows. Then, users authoring multiplatform projects would have some guidance without having to engage in significant trial-and-error activity.

### Why is this needed?

My goal is to be able to open and use a Java library Windows, Linux, Unix, and MacOS on 64-bit hardware platforms. I'll take care with providing the full paths. The dlfcn concept & code of @RomainMuller will be useful to me in modifying my own work.

TotallyGamerJet:
Closes #273

n/a

## What this PR does | solves
Clarification on example code for windows in README and adding equivalent windows functions to

### DIFF
--- a/dlfcn.go
+++ b/dlfcn.go
@@ -35,7 +35,8 @@ func init() {
 // Dlopen calls should be balanced with a Dlclose call.
 //
 // This function is not available on Windows.
-// Use [golang.org/x/sys/windows.LoadLibrary] behind a //go:build windows build tag.
+// Use [golang.org/x/sys/windows.LoadLibrary], [golang.org/x/sys/windows.NewLazyDLL], or
+// [golang.org/x/sys/windows.NewLazySystemDLL] for Windows instead.
 func Dlopen(path string, mode int) (uintptr, error) {
 	u := fnDlopen(path, mode)
 	if u == 0 {
@@ -50,7 +51,7 @@ func Dlopen(path string, mode int) (uintptr, error) {
 // when that library was loaded, Dlsym returns zero.
 //
 // This function is not available on Windows.
-// Use [golang.org/x/sys/windows.GetProcAddress] behind a //go:build windows build tag.
+// Use [golang.org/x/sys/windows.GetProcAddress] for Windows instead.
 func Dlsym(handle uintptr, name string) (uintptr, error) {
 	u := fnDlsym(handle, name)
 	if u == 0 {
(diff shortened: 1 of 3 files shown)


dlfcn.go
```diff
@@ -33,6 +33,9 @@ func init() {
 // A second call to Dlopen with the same path will return the same handle, but the internal
 // reference count for the handle will be incremented. Therefore, all
 // Dlopen calls should be balanced with a Dlclose call.
+//
+// This function is not available on Windows.
+// Use [golang.org/x/sys/windows.LoadLibrary] behind a //go:build windows build tag.
```
hajimehoshi, line 38: Maybe,

```
Use [golang.org/x/sys/windows.LoadLibrary] for Windows instead.
```

? It's because `//go:build` is not needed if the file has `_windows` suffix.

hajimehoshi: Also, `NewLazyDLL` and `NewLazySystemDLL` are other options.

TotallyGamerJet: Should I list all three options?

hajimehoshi: I think so. (Offtopic: Personally `NewLazySystemDLL` is the best option, but the comment doesn't have to say this is the best)

TotallyGamerJet:
Also, purego.Dlerror type is not available on Windows. I'm thinking I should add a comment to that too but there is no equivalent to it

hajimehoshi (approved):
LGTM, thanks!

hajimehoshi:
> Also, purego.Dlerror type is not available on Windows. I'm thinking I should add a comment to that too but there is no equivalent to it

A comment about the fact there is not couterpart would be fine.

hajimehoshi (approved):
Still LGTM, thanks!

PR #278, opened 2024-09-03
examples: remove depdencies on golang.org/x/sys/windows

### LINKED ISSUE #270: Remove dependency on `golang.org/x/sys`

hajimehoshi:
Closes #270

Closes #270

Bug

## What this PR does | solves
This PR moves external dependencies from the entire module.

### DIFF
--- a/examples/libc/main_windows.go
+++ b/examples/libc/main_windows.go
@@ -3,9 +3,11 @@
 
 package main
 
-import "golang.org/x/sys/windows"
+import "syscall"
 
 func openLibrary(name string) (uintptr, error) {
-	handle, err := windows.LoadLibrary(name)
+	// Use [syscall.LoadLibrary] here to avoid external dependencies (#270).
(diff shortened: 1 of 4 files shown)


TotallyGamerJet (approved):
LGTM

PR #282, opened 2024-10-18
all: improve memory usage

TotallyGamerJet:
Don't allocate when being called by objc.Send. This is done by just inlining the expansion of the parameters directly into the argument list instead of making a tmp slice.

Use a sync.Pool for syscall15Args struct to save on allocations

### DIFF
--- a/func.go
+++ b/func.go
@@ -16,7 +16,7 @@ import (
 	"github.com/ebitengine/purego/internal/strings"
 )
 
-var pool = sync.Pool{New: func() interface{} {
+var thePool = sync.Pool{New: func() any {
 	return new(syscall15Args)
 }}
 
@@ -253,8 +253,8 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
 			runtime.KeepAlive(keepAlive)
 			runtime.KeepAlive(args)
 		}()
-		syscall := pool.Get().(*syscall15Args)
-		defer pool.Put(syscall)
+		syscall := thePool.Get().(*syscall15Args)
+		defer thePool.Put(syscall)
 		if ty.NumOut() == 1 && ty.Out(0).Kind() == reflect.Struct {
 			outType := ty.Out(0)
 			if runtime.GOARCH == "amd64" && outType.Size() > maxRegAllocStructSize {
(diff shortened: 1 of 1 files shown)


func.go
```diff
@@ -10,11 +10,16 @@ import (
 	"math"
 	"reflect"
 	"runtime"
+	"sync"
 	"unsafe"
 
 	"github.com/ebitengine/purego/internal/strings"
 )
 
+var pool = sync.Pool{New: func() interface{} {
```
hajimehoshi, line 19: We can use `any` instead of `interface{}` since Go 1.18.

hajimehoshi, line 19: I prefer `thePool` over `pool`

hajimehoshi: On secont thought, we can use `interface{}` for this PR for consistency, and we can replace all `interface{}` with `any` in a later PR. So, using `any` here is an optional.

func.go
```diff
@@ -357,15 +332,54 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
+			args[0] = v
+			return args[:1]
+		} else {
+			return []reflect.Value{v}
+		}
 	})
 	fn.Set(v)
 }
 
+func addValue(v reflect.Value, keepAlive []interface{}, addInt func(x uintptr), addFloat func(x uintptr), addStack func(x uintptr), numInts *int, numFloats *int, numStack *int) []interface{} {
```
hajimehoshi, line 350: optional: `any`?

func.go
```diff
@@ -276,48 +270,29 @@ func RegisterFunc(fptr interface{}, cfn uintptr) {
...
-				addInt(v.Pointer())
-			case reflect.Func:
-				addInt(NewCallback(v.Interface()))
-			case reflect.Bool:
-				if v.Bool() {
-					addInt(1)
-				} else {
-					addInt(0)
+		for i, v := range args {
+			if variadic, ok := args[i].Interface().([]interface{}); ok {
```
hajimehoshi, line 274: Optional: `any`?

hajimehoshi (approved):
LGTM % comments

Thanks!

TotallyGamerJet:
I'll just make another PR that changes `interface{}` to `any`

Issue #305, opened 2025-03-09
build failure on FreeBSD w/o optimization


hajimehoshi:
### PureGo Version

009f4b0e473c90e938fda983e6945c5c00c4cca5

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [x] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go1.22.12 (but also other versions)

### What steps will reproduce the problem?

Run the test CGO_ENABLED=0 w/o optimization

### What is the expected result?

The test passes

### What happens instead?

The test fails

https://github.com/hajimehoshi/purego/actions/runs/13751525916/job/38453081556

```
  => go test CGO_ENABLED=0 w/o optimization
  # github.com/ebitengine/purego/internal/fakecgo
  internal/fakecgo/freebsd.go:23:3: //go:cgo_export_dynamic environ only allowed in cgo-generated code
  internal/fakecgo/freebsd.go:24:3: //go:cgo_export_dynamic __progname only allowed in cgo-generated code
  FAIL	github.com/ebitengine/purego [build failed]
  FAIL	github.com/ebitengine/purego/examples/libc [build failed]
  ?   	github.com/ebitengine/purego/internal/cgo	[no test files]
  FAIL	github.com/ebitengine/purego/internal/fakecgo [build failed]
  FAIL	github.com/ebitengine/purego/internal/load [build failed]
  ?   	github.com/ebitengine/purego/internal/strings	[no test files]
  FAIL
```

### Anything else you feel useful to add?

We should add `-e` to the shebang `#!/bin/sh`.

I don't think this is happening in the stable branch, where the test is not executed in a shell script.

Issue #304, opened 2025-03-09
suspicous segmentation fault on NetBSD


hajimehoshi:
### PureGo Version

https://github.com/hajimehoshi/purego/actions/runs/13750780953/job/38451482165

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [x] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go1.24.1 (but also all the others)

### What steps will reproduce the problem?

Run go test (CGO_ENABLED=0)

### What is the expected result?

The tests pass

### What happens instead?

The tests fail

https://github.com/hajimehoshi/purego/actions/runs/13750780953/job/38451482165

```
=> go test CGO_ENABLED=0
  signal: segmentation fault (core dumped)
  FAIL	github.com/ebitengine/purego	0.007s
  ?   	github.com/ebitengine/purego/examples/libc	[no test files]
  ?   	github.com/ebitengine/purego/internal/cgo	[no test files]
  ?   	github.com/ebitengine/purego/internal/fakecgo	[no test files]
  ?   	github.com/ebitengine/purego/internal/load	[no test files]
  ?   	github.com/ebitengine/purego/internal/strings	[no test files]
  FAIL
```

hajimehoshi:
The mysterious thing is, the test does not 'fail' with a red signal, but says 'fail' in the output.

TotallyGamerJet:
Perhaps the size of one of the structs in Cgo doesn't match. Perhaps the non failure is bc it's inside a script?

hajimehoshi:
For the test script, just adding `-e` should work.

```diff
diff --git a/.github/scripts/bsd_tests.sh b/.github/scripts/bsd_tests.sh
index d1503d4..4199bce 100755
--- a/.github/scripts/bsd_tests.sh
+++ b/.github/scripts/bsd_tests.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -e
 
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: 2025 The Ebitengine Authors
```

and unfortunately, I found another issue on FreeBSD...

https://github.com/hajimehoshi/purego/actions/runs/13751525916/job/38453080319

```
=> go test CGO_ENABLED=0 w/o optimization
  # github.com/ebitengine/purego/internal/fakecgo
[...]

PR #306, opened 2025-03-17
.github/workflows: add -e for FreeBSD test script and skip NetBSD tests

hajimehoshi:
Without `-e` in the shebang `#!/bin/sh`, the script passed even some commands failed and then we missed some errors happened. This change adds `-e` to abort tests with an error code explicitly.

As #305 says, FreeBSD tests w/o optimization fails. This change skips the test temporarilly.

This change also retracts the support of NetBSD temporarily. Let's revert README again after #304 is fixed.

## What this PR does | solves
#304 and #305

### DIFF
--- a/.github/scripts/bsd_tests.sh
+++ b/.github/scripts/bsd_tests.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -e
 
 # SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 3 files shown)


TotallyGamerJet (approved):
LGTM! Sorry I've been busy and haven't had time to look into these

hajimehoshi:
No worries!

PR #310, opened 2025-04-11
Fix alignment of stack args on macOS

### LINKED ISSUE #309: segfault w/ many args
### PureGo Version

0.9.0.alpha.2

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

1.24.1

👋 hey there, thanks for this project, its amazing! 

im experiencing a problem where the arguments from the go side of things are not mapping correctly in the zig side of things. for example, I have a function ive registered like this:

```go
	var notWorking func(
		driverPtr uintptr,
		operationID *uint32,
		cancel *bool,
		stream,
		filter,
		filterType,
		filterNamespacePrefix,
		filterNamespace string,
		period uint64,
		stopTime uint64,
		dscp uint8,
		weighting uint8,
		dependency uint32,
		encoding string,
	) int
```

calling this function like so:

```go
notWorking(1, &i, &t, "stream", "filter", "filter_type", "filter_ns_prefix", "filter_ns", 1, 2, 3, 4, 5, "encoding")
```

results in zig receiving these args like this:

```
	D PTR > 1
	OP ID > u32@14000112134
	CANCEL > bool@14000112130
	STREAM > stream
	FILTER > filter
	FILTER TYPE > filter_type
	FILTER NAMESPACE PREFIX > filter_ns_prefix
	FILTER NAMESPACE > filter_ns
	PERIOD > 1
	STOP TIME > 2
	DSCP > 3
	WEIGHTING > 0
	DEPENDENCY > 0
	ENCODING > SIGSEGV: segmentation violation
```

the first section of things looks correct -- up to the "dscp" arg, then we see where "weighting" should have a value of `4` and "dependency" a value of `5` we see zero, and attempting to access the value of the "encoding" argument causes a seg fault.

I've seen this happen tons when I've had my registration function not properly aligned to the args on the zig side, but I have checked this 900 times (at least :P) so I'm pretty sure thats not to blame.

doing a bit of investigation I can see that [here](https://github.com/ebitengine/purego/blob/main/func.go#L291) the arguments getting loaded up into the sys call *are in fact correct* -- in this case the period or stop time and onward are put into the stack view into the sysargs array and they all look correct. 

anything beyond that and it is very much dark magic to me how it "gets to" the c/zig things.

I was able to get this working by moving the args around to look like this:

```go
	var working func(
		driverPtr uintptr,
		operationID *uint32,
		cancel *bool,
		period uint64,
		stopTime uint64,
		dscp uint8,
		weighting uint8,
		dependency uint32,
		encoding,
		stream,
		filter,
		filterType,
		filterNamespacePrefix,
		filterNamespace string,
	) int
```

im out of my depth here but I can only assume/believe this has to do w/ alignment of the parameters? 

interestingly changing args around a little bit less to look like this:

```go
	var notWorkingDifferently func(
		driverPtr uintptr,
		operationID *uint32,
		cancel *bool,
		stream,
		filter,
		filterType,
		filterNamespacePrefix,
		filterNamespace string,
		period uint64,
		stopTime uint64,
		dscp uint32, // using all uint32 to maybe have better alignment?
		weighting uint32,
		dependency uint32,
		encoding string,
	) int
```

causes things to still not work, but to work a little differently :D 

I've created [this](https://github.com/carlmontanari/repro-purego-zig-issue) repo that shows the issue. there are three functions in zig that are exported, and registered. then three calls w/ comments showing what ive seen. 

also just as an additional note I was on 0.8.2 but have been using the latest tag -- both with the same results.

to run the repro you can (install zig > 0.14.0 [like so](https://ziglang.org/learn/getting-started/#direct)) then `zig build` and just `go run main.go`.

like I said totally possible this is user error and if so please feel free to close! but just wanted to share in case theres something else going on here. im happy to help test or provide any other info. thanks again for a great project!

for args in a registered function like this to be passed to zig/c as expected:

```go
	var notWorking func(
		driverPtr uintptr,
		operationID *uint32,
		cancel *bool,
		stream,
		filter,
		filterType,
		filterNamespacePrefix,
		filterNamespace string,
		period uint64,
		stopTime uint64,
		dscp uint8,
		weighting uint8,
		dependency uint32,
		encoding string,
	) int
```

### What happens instead?

segfault when trying to access the `encoding` value, and values of some of the stack uints missing.

_No response_

TotallyGamerJet:
Closes #309

bug

## What this PR does | solves
fixes issue with alignment on macOS

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,10 +34,13 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y gcc-aarch64-linux-gnu g++-aarch64-linux-gnu qemu-user
 
+      - name: Set up the prerequisites
+        if: runner.os == 'Windows'
+        uses: msys2/setup-msys2@v2
+
       - name: go vet
-        if: runner.os != 'Windows' && runner.os != 'macOS'
         run: |
-          go vet -v ./...
+          env CGO_ENABLED=0 go vet -v ./...
 
       - name: go build
         run: |
--- a/func.go
+++ b/func.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"reflect"
 	"runtime"
+	"strconv"
 	"sync"
 	"unsafe"
 
@@ -206,7 +207,6 @@ func RegisterFunc(fptr any, cfn uintptr) {
 	}
 	v := reflect.MakeFunc(ty, func(args []reflect.Value) (results []reflect.Value) {
 		var sysargs [maxArgs]uintptr
-		stack := sysargs[numOfIntegerRegisters():]
 		var floats [numOfFloats]uintptr
 		var numInts int
 		var numFloats int
@@ -215,7 +215,7 @@ func RegisterFunc(fptr any, cfn uintptr) {
 		if runtime.GOARCH == "arm64" || runtime.GOOS != "windows" {
 			// Windows arm64 uses the same calling convention as macOS and Linux
 			addStack = func(x uintptr) {
-				stack[numStack] = x
+				sysargs[numOfIntegerRegisters()+numStack] = x
 				numStack++
 			}
 			addInt = func(x uintptr) {
@@ -279,6 +279,29 @@ func RegisterFunc(fptr any, cfn uintptr) {
 					keepAlive = addValue(reflect.ValueOf(x), keepAlive, addInt, addFloat, addStack, &numInts, &numFloats, &numStack)
 				}
 				continue
+			}
+			if runtime.GOARCH == "arm64" && (numInts >= numOfIntegerRegisters() || numFloats >= numOfFloats) && v.Kind() != reflect.Struct { // hit the stack
+				fields := make([]reflect.StructField, len(args[i:]))
+
+				for j, val := range args[i:] {
+					if val.Kind() == reflect.String {
+						ptr := strings.CString(v.String())
+						keepAlive = append(keepAlive, ptr)
+						val = reflect.ValueOf(ptr)
+						args[i+j] = val
+					}
+					fields[j] = reflect.StructField{
+						Name: "X" + strconv.Itoa(j),
+						Type: val.Type(),
+					}
+				}
+				structType := reflect.StructOf(fields)
+				structInstance := reflect.New(structType).Elem()
+				for j, val := range args[i:] {
+					structInstance.Field(j).Set(val)
+				}
+				placeRegisters(structInstance, addFloat, addInt)
+				break
 			}
 			keepAlive = addValue(v, keepAlive, addInt, addFloat, addStack, &numInts, &numFloats, &numStack)
 		}
--- a/func.go
+++ b/func.go
@@ -280,7 +280,8 @@ func RegisterFunc(fptr any, cfn uintptr) {
 				}
 				continue
 			}
-			if runtime.GOARCH == "arm64" && (numInts >= numOfIntegerRegisters() || numFloats >= numOfFloats) && v.Kind() != reflect.Struct { // hit the stack
+			if runtime.GOARCH == "arm64" && runtime.GOOS == "darwin" &&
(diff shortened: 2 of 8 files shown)


TotallyGamerJet:
@hajimehoshi Have you seen this Windows error before with `t.TempDir()`?

```
func_test.go:141: Build C:\Users\USER~1\AppData\Local\Temp\TestABI524339155\001\abitest.so
    testing.go:1225: TempDir RemoveAll cleanup: remove C:\Users\USER~1\AppData\Local\Temp\TestABI524339155\001\abitest.so: Access is denied.
```

Appears similar to https://github.com/golang/go/issues/50051

testdata/abitest/abi_test.c
```diff
@@ -0,0 +1,27 @@
...
+    assert(b == 2);
+    assert(c == 1024);
+    return a | b | c;
+}
+
+uint32_t stack_string(uint32_t a, uint32_t b, uint32_t c, uint32_t d, uint32_t e, uint32_t f, uint32_t g, uint32_t h, const char * i) {
+    assert(i != 0);
+    assert(strcmp(i, "test") == 0);
+    return a | b | c | d | e | f | g | h;
+}
```
hajimehoshi, line 27: Add a new line.

.github/workflows/test.yml
```diff
@@ -34,10 +34,13 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y gcc-aarch64-linux-gnu g++-aarch64-linux-gnu qemu-user
 
+      - name: Set up the prerequisites
+        if: runner.os == 'Windows'
+        uses: msys2/setup-msys2@v2
```
hajimehoshi, line 39: Is this needed to run `go vet`?

hajimehoshi:
> Have you seen this Windows error before with t.TempDir()?

I don't think so. Is there any other process having file handlers in the directory?

hajimehoshi:
```go

	if err := buildSharedLib("CC", libFileName, filepath.Join("testdata", "abitest", "abi_test.c")); err != nil {
		t.Fatal(err)
	}
	defer os.Remove(libFileName)

	lib, err := load.OpenLibrary(libFileName)
```

I guess the test itself opens the lib in the current directory and keeps the handle to it. Would [`CloseHandle`](https://pkg.go.dev/syscall?GOOS=windows#CloseHandle) work?

.github/workflows/test.yml
TotallyGamerJet: No it's need to build the abitest library since it uses GCC to compile the C code.

func_test.go
```diff
@@ -143,13 +142,16 @@ func TestABI(t *testing.T) {
...
-	defer os.Remove(libFileName)
 
 	lib, err := load.OpenLibrary(libFileName)
 	if err != nil {
 		t.Fatalf("Dlopen(%q) failed: %v", libFileName, err)
 	}
-
+	defer func() {
+		if err := load.CloseLibrary(lib); err != nil {
+			t.Errorf("failed to close library: %s", err)
```
hajimehoshi, line 155: `t.Fatal` is better

.github/workflows/test.yml
hajimehoshi: Ack

func.go
```diff
@@ -280,6 +280,30 @@ func RegisterFunc(fptr any, cfn uintptr) {
 				}
 				continue
 			}
+			if runtime.GOARCH == "arm64" && runtime.GOOS == "darwin" &&
+				(numInts >= numOfIntegerRegisters() || numFloats >= numOfFloats) && v.Kind() != reflect.Struct { // hit the stack
```
hajimehoshi, line 284:
```go
numFloats >= numOfFloats
```

Can we rename either variable?

hajimehoshi (approved):
LGTM, thanks!

PR #317, opened 2025-05-13
Allow uintptr in struct on amd64

TotallyGamerJet:

Relates to https://github.com/hajimehoshi/ebiten/issues/3239

Fixes the bug that uintptr as the only field is not passed to uintptr field on amd64

### DIFF
--- a/struct_amd64.go
+++ b/struct_amd64.go
@@ -200,7 +200,7 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 				val |= f.Uint() << shift
 				shift += 32
 				class |= _INTEGER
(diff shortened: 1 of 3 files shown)


hajimehoshi (commented):
LGTM, but the mystery thing is that `reflect.Uintptr` doesn't exist in `struct_arm64.go`

TotallyGamerJet:
> LGTM, but the mystery thing is that `reflect.Uintptr` doesn't exist in `struct_arm64.go`

Yes it does

https://github.com/ebitengine/purego/blob/5047c08daa38ee2b067f695e1c671c0d122cde41/struct_arm64.go#L142

hajimehoshi:
I was looking at a wrong branch. Thanks!

PR #319, opened 2025-05-20
internal/fakecgo: disable race in fakecgo

### LINKED ISSUE #318: `go test` causes segmentation fault with `-race`
### PureGo Version

v0.9.0-alpha.5

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go version go1.24.3 darwin/arm64

```
cd go/src/github.com/ebitengine/purego
CGO_ENABLED=0 go test -race
```

Pass the tests.

### What happens instead?

```
% CGO_ENABLED=0 go test -race
signal: segmentation fault
FAIL    github.com/ebitengine/purego    0.231s
```

This might be related to [gopsutil issue](https://github.com/shirou/gopsutil/issues/1774).

TotallyGamerJet:
Closes #318

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -139,7 +139,8 @@ jobs:
       - name: go test race
         if: ${{ !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') }}
         run: |
-          go test -race -shuffle=on -v -count=10 ./...
+          env CGO_ENABLED=0 go test -race -shuffle=on -v -count=10 ./...
+          env CGO_ENABLED=1 go test -race -shuffle=on -v -count=10 ./...
 
   bsd:
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -139,7 +139,9 @@ jobs:
       - name: go test race
         if: ${{ !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') }}
         run: |
-          env CGO_ENABLED=0 go test -race -shuffle=on -v -count=10 ./...
+          if [[ "$(uname)" == "Darwin" ]]; then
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -139,7 +139,10 @@ jobs:
       - name: go test race
         if: ${{ !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') }}
         run: |
-          go test -race -shuffle=on -v -count=10 ./...
+          if [[ "$(uname)" == "Darwin" ]]; then
+            env CGO_ENABLED=0 go test -race -shuffle=on -v -count=10 ./...
```
hajimehoshi, line 143: Why is this working only with Darwin?

TotallyGamerJet: Only macOS allows building with the race detector and Cgo disabled. All other platforms fail to build with the message: `go: -race requires cgo`

hajimehoshi: I didn't know that. Actually Cgo was required even on Windows.

Let's separate the tasks like this

```yaml
      - name: go test race (no Cgo)
        if: ${{ runner.os == 'macOS' &&  !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') }}
        run: |
          # -race requires Cgo except for macOS. See https://go.dev/doc/articles/race_detector#Requirements
          env CGO_ENABLED=0 go test -race -shuffle=on -v -count=10 ./...
      - name: go test race (Cgo)
        run: |
          env CGO_ENABLED=1 go test -race -shuffle=on -v -count=10 ./...
```

hajimehoshi (approved):
LGTM, thanks!

PR #328, opened 2025-07-30
purego: use sync.Pool in syscall_syscall15X with wrapper

oliverbestmann:
This works around the stack growth issue in 05d8da4 while providing the same benefit. The syscall15Args does not escape to the heap for each purego call, reducing memory allocation rate **a lot** for ebitengine!

I am not sure why we need `go:nosplit` at the first place, though. Maybe that can just be removed and this optimization simplified.

Reduces allocations on the hot path

Performance

## What this PR does | solves
reduced garbage collection

### DIFF
--- a/syscall_sysv.go
+++ b/syscall_sysv.go
@@ -14,14 +14,17 @@ import (
 
 var syscall15XABI0 uintptr
 
-//go:nosplit
 func syscall_syscall15X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15 uintptr) (r1, r2, err uintptr) {
-	args := syscall15Args{
+	args := thePool.Get().(*syscall15Args)
+	defer thePool.Put(args)
+
+	*args = syscall15Args{
 		fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15,
 		a1, a2, a3, a4, a5, a6, a7, a8,
 		0,
 	}
-	runtime_cgocall(syscall15XABI0, unsafe.Pointer(&args))
+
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Context: https://github.com/ebitengine/purego/commit/450319b71c523c3716a29e3068ef333583744aea

I'm not sure the change is really safe. @TotallyGamerJet thoughts?

oliverbestmann:
It is working fine within ebitengine without optimizations and pointer checking. I don't see a reason for it not being unsafe, as the calls are all happening outside of the nosplit method.
And as long as the cgo call itself is not holding on to the args pointer, it should have the same behavior.

syscall_sysv.go
```diff
@@ -14,14 +14,28 @@ import (
...
-//go:nosplit
+var pool = sync.Pool{
+	New: func() any {
+		return &syscall15Args{}
+	},
+}
+
 func syscall_syscall15X(fn, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15 uintptr) (r1, r2, err uintptr) {
-	args := syscall15Args{
+	args := pool.Get().(*syscall15Args)
```
hajimehoshi: Why not using the existing `thePool`?

oliverbestmann: That was on me, i just did not see it. fixed.

TotallyGamerJet:
I don't think nosplit is actually needed. The uintptr are all forced onto the heap where a stack split won't affect their addresses since Go uses a non moving GC.

oliverbestmann:
I've removed the `noescape` directive and the wrapper function and pushed again.

oliverbestmann:
Yes

TotallyGamerJet (approved):
LGTM

hajimehoshi (approved):

PR #333, opened 2025-08-18
Add structs.HostLayout to Go representations of C structs

### LINKED ISSUE #259: Use Go 1.23's new structs.HostLayout
### Operating System

- [X] Windows
- [X] macOS
- [X] Linux
- [X] FreeBSD
- [X] Android
- [X] iOS

Use `structs.HostLayout` in all Platform specific structs

### Why is this needed?

In Go 1.23, a new package and type called `structs.HostLayout` was added in https://github.com/golang/go/issues/66408. Adding this field to a struct instructs the compiler to guarantee the struct matches the C ABI. 

We currently, rely on the fact that Go's current struct layout matches the C ABI. The following is an incomplete list of structs that must match C ABI.
* objc_super
* G
* ThreadStart
* pthread_mutex_t
* pthread_cond_t

As well as all the one's in `struct_test.go`, `objc_runtime_darwin_test.go` and in the `examples/window` platform files

This is not urgent as there is no current plan to change Go's struct layout

judah-caruso:
Closes #259 

Enhancement

## What this PR does | solves
This PR ensures internal structs (with C equivalents) match the memory layout of the host's C ABI.

### DIFF
--- a/func_test.go
+++ b/func_test.go
@@ -136,7 +136,7 @@ func TestRegisterLibFunc_Bool(t *testing.T) {
 }
 
 func TestABI(t *testing.T) {
-	if runtime.GOOS == "windows" && runtime.GOARCH == "386" {
+	if runtime.GOOS == "windows" || runtime.GOARCH != "386" {
 		t.Skip("need a 32bit gcc to run this test") // TODO: find 32bit gcc for test
 	}
 	libFileName := filepath.Join(t.TempDir(), "abitest.so")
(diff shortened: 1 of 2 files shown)


func_test.go
```diff
@@ -136,7 +136,7 @@ func TestRegisterLibFunc_Bool(t *testing.T) {
 }
 
 func TestABI(t *testing.T) {
-	if runtime.GOOS == "windows" && runtime.GOARCH == "386" {
+	if runtime.GOOS == "windows" || runtime.GOARCH != "386" {
```
judah-caruso, line 139: I made this change because the skip condition didn't make sense to me. 32bit windows wants to skip the test that requires 32bit gcc?

PR #334, opened 2025-08-19
Windows Struct Support: Organization + Tests can build a DLL

### LINKED ISSUE #237: Struct support for Windows
### Operating System

- [X] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

Support structs as return values and parameters of functions on Windows.

### Why is this needed?

Technically purego already supports structs, but you have to make some weird stunts:

If the struct's size is less than or equal 8 bytes, you just can convert them into a uintptr. Same goes for return values. Example:

C:
```c
RLAPI void ClearBackground(Color color); 
```
Golang:
```golang
var clearBackground func(col uintptr)

purego.RegisterLibFunc(&clearBackground, raylibDll, "ClearBackground")

func ClearBackground(col color.RGBA) {
	clearBackground(*(*uintptr)(unsafe.Pointer(&col)))
}
```

If the struct's size is lager then 8 bytes, you have to use a pointer as parameter. For return values you have to pass it as first argument. Here is another example from raylib-go:

C:
```c
RLAPI Texture2D LoadTextureCubemap(Image image, int layout);   
```
Golang:
```golang
var loadTextureCubemap func(texture *Texture2D, image *Image, layout int32)

purego.RegisterLibFunc(&loadTextureCubemap, raylibDll, "LoadTextureCubemap")

func LoadTextureCubemap(image Image, layout int32) Texture2D {
	var texture Texture2D
	loadTextureCubemap(&texture, &image, layout)
	return texture
}
```

judah-caruso:
Closes #237 

Feature

## What this PR does | solves
Adds struct parameter and return support for Windows (amd64)

### DIFF
--- a/testdata/structtest/struct_test.c
+++ b/testdata/structtest/struct_test.c
@@ -3,19 +3,19 @@
 
 #include "stdint.h"
 
-#if defined(__x86_64__) || defined(__aarch64__) || defined(_WIN64)
+#if defined(__x86_64__) || defined(__aarch64__) || defined(_WIN32)
 typedef int64_t GoInt;
 typedef uint64_t GoUint;
 #endif
 
 #ifdef _WIN32
 #define EXPORT __declspec(dllexport)
-#elif __APPLE__
+#else
 #define EXPORT
 #endif
 
 // Empty structs are not allowed when using MSVC
-#ifdef __APPLE__
+#ifndef _WIN32
 // Empty is empty
 struct Empty;
 
--- a/testdata/abitest/abi_test.c
+++ b/testdata/abitest/abi_test.c
@@ -8,7 +8,7 @@
 
 #ifdef _WIN32
 #define EXPORT __declspec(dllexport)
-#elif __APPLE__
+#else
 #define EXPORT
 #endif
 
--- a/testdata/libdlnested/nested_test.cpp
+++ b/testdata/libdlnested/nested_test.cpp
@@ -4,7 +4,7 @@
 #ifdef __APPLE__
 #include <dlfcn.h>
 #define DLSYM(F, N) dlsym(F, N)
-#elif _WIN32
+#elif defined(_WIN32)
 #define WIN32_LEAN_AND_MEAN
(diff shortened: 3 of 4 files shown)


func_test.go
```diff
@@ -4,36 +4,15 @@
...
 	"runtime"
-	"strings"
 	"testing"
 	"unsafe"
 
 	"github.com/ebitengine/purego"
 	"github.com/ebitengine/purego/internal/load"
 )
 
-func getSystemLibrary() (string, error) {
```
TotallyGamerJet, line 20: I think this can stay here instead of moving it to a separate file

func_test.go
```diff
@@ -186,41 +162,3 @@ func TestABI(t *testing.T) {
 		}
 	}
 }
-
```
TotallyGamerJet, line 189: I would also keep this here and just modify it to support windows. The reason is it is also used by Linux so if it's pulled out it'd create another file which isn't necessary

testdata/structtest/struct_test.c
```diff
@@ -3,43 +3,52 @@
 
 #include "stdint.h"
 
-#if defined(__x86_64__) || defined(__aarch64__)
+#if defined(__x86_64__) || defined(__aarch64__) || defined(_WIN32)
```
TotallyGamerJet, line 6: Why specifically add the windows check is it not x86 or arm?

testdata/structtest/struct_test.c
```diff
@@ -3,43 +3,52 @@
...
+#define EXPORT __declspec(dllexport)
+#else
+#define EXPORT
+#endif
+
+// Empty structs are not allowed when using MSVC
+#ifndef _WIN32
 // Empty is empty
-struct Empty {};
+struct Empty;
```
TotallyGamerJet, line 20: Why remove the braces?

TotallyGamerJet (commented):
I know you're still working on this but wanted to leave some comments to help

PR #335, opened 2025-09-09
purego: add support for loong64

limeidan:
Loong64 stands for LoongArch64, a 64-bit processor architecture launched by Loongson in 2020. It has gained widespread support in the open source community, and Golang officially supported the Loong64 architecture in 2022.


hajimehoshi:
I was wondering if it'd be possible to add loong64 tests by QEMU.

.github/workflows/test.yml
```diff
@@ -74,6 +75,13 @@ jobs:
           env GOOS=netbsd GOARCH=amd64 go build -v ./...
           env GOOS=netbsd GOARCH=arm64 go build -v ./...
 
+      - name: go build
```
hajimehoshi: `go build (Linux loong64)`?

.github/workflows/test.yml
```diff
@@ -123,6 +131,27 @@ jobs:
...
+          go env -w CC=loongarch64-linux-gnu-gcc-14
+          go env -w CXX=loongarch64-linux-gnu-g++-14
+          env GOOS=linux GOARCH=loong64 CGO_ENABLED=0 go test -c -o=purego-test-nocgo .
+          env QEMU_LD_PREFIX=/usr/loongarch64-linux-gnu qemu-loongarch64 ./purego-test-nocgo -test.shuffle=on -test.v -test.count=10
+          env GOOS=linux GOARCH=loong64 CGO_ENABLED=1 go test -c -o=purego-test-cgo .
+          env QEMU_LD_PREFIX=/usr/loongarch64-linux-gnu qemu-loongarch64 ./purego-test-cgo -test.shuffle=on -test.v -test.count=10
+          go env -u CC
+          go env -u CXX
+
+      - name: go build (plugin)
```
hajimehoshi: Do we need this? If so, why not adding this for other archs?

.github/workflows/test.yml
```diff
@@ -32,7 +32,8 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           sudo apt-get update
-          sudo apt-get install -y gcc-aarch64-linux-gnu g++-aarch64-linux-gnu qemu-user
+          sudo apt-get install -y gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-14-loongarch64-linux-gnu g++-14-loongarch64-linux-gnu libc6-loong64-cross qemu-user
+          sudo ln -s /usr/loongarch64-linux-gnu/lib/ld-linux-loongarch-lp64d.so.1 /usr/lib64/ld-linux-loongarch-lp64d.so.1
```
hajimehoshi: Why do we need this symbolic link?

.github/workflows/test.yml
limeidan: I added the same test to loong64 based on #255, but I wrote it in the wrong place. It can be merged into `go test (Linux loong64)`. This test seems to be testing #254 (a bug on amd64), so maybe I don't need to add it to loong64.

.github/workflows/test.yml
limeidan: This is redundant, I will remove it.

sys_unix_loong64.s
```diff
@@ -0,0 +1,75 @@
...
+	MOVD	F6, 48(R14)
+	MOVD	F7, 56(R14)
+	MOVV	R4, 64(R14)	
+	MOVV	R5, 72(R14)	
+	MOVV	R6, 80(R14)	
+	MOVV	R7, 88(R14)	
+	MOVV	R8, 96(R14)	
+	MOVV	R9, 104(R14)	
+	MOVV	R10, 112(R14)	
+	MOVV	R11, 120(R14)
```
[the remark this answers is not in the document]
hajimehoshi: Was this resolved?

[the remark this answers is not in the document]
limeidan: Yes, it's done.

.github/workflows/test.yml
limeidan: Done. Thanks.

.github/workflows/test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           sudo apt-get update
-          sudo apt-get install -y gcc-aarch64-linux-gnu g++-aarch64-linux-gnu qemu-user
+          sudo apt-get install -y gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-14-loongarch64-linux-gnu g++-14-loongarch64-linux-gnu qemu-user
```
hajimehoshi, line 35: Doesn't `gcc-loongarch64-linux-gnu g++-loongarch64-linux-gnu` work?

zcallback_loong64.s
```diff
@@ -0,0 +1,4014 @@
+// Code generated by wincallback.go using 'go generate'. DO NOT EDIT.
```
hajimehoshi, line 1: We have similar files like `zcallback_amd64.s`, and the original files were generated by `wincallback.go`, but the current version should not have this 'DO NOT EDIT' comment. @TotallyGamerJet what do you think?

hajimehoshi: Also, was this file really generated by `wincallback.go`?

.github/workflows/test.yml
limeidan: Yes, the ubuntu-latest(24.04) does not have packages gcc-loongarch64-linux-gnu and g++-loongarch64-linux-gnu, but it has gcc-14-loongarch64-linux-gnu and g++-14-loongarch64-linux-gnu.

zcallback_loong64.s
limeidan: Yes, I modified the wincallback.go file locally and used it to generate the zcallback_loong64.s file.

hajimehoshi: Hmm, this is a little tricky situation. In this case, we should have the modified `wincallback.go` in this PR, but I'd like to wait for @TotallyGamerJet's answer about how we generated our files.

limeidan: OK, I will make revisions once the questions are answered.

TotallyGamerJet: I didn't generate them. I copied them from the Go repo directly but the script is simple enough that it likely should just be generated from it. 

https://github.com/golang/go/blob/master/src/runtime/wincallback.go

.github/workflows/test.yml
```diff
@@ -123,6 +130,18 @@ jobs:
...
           go env -u CXX
 
+      - name: go test (Linux loong64)
+        if: ${{ runner.os == 'Linux' && !startsWith(matrix.go, '1.18.') && !startsWith(matrix.go, '1.19.') && !startsWith(matrix.go, '1.20.') && !startsWith(matrix.go, '1.21.') && !startsWith(matrix.go, '1.22.') && !startsWith(matrix.go, '1.23.') && !startsWith(matrix.go, '1.24.') }}
+        run: |
+          go env -w CC=loongarch64-linux-gnu-gcc-14
+          go env -w CXX=loongarch64-linux-gnu-g++-14
+          env GOOS=linux GOARCH=loong64 CGO_ENABLED=0 go test -c -o=purego-test-nocgo .
+          env QEMU_LD_PREFIX=/usr/loongarch64-linux-gnu qemu-loongarch64 ./purego-test-nocgo -test.shuffle=on -test.v -test.count=10
+          env GOOS=linux GOARCH=loong64 CGO_ENABLED=1 go test -c -o=purego-test-cgo .
```
TotallyGamerJet, line 140: Shouldn't we still be testing that CGO works on >1.19?

zcallback_loong64.s
hajimehoshi: Then, we should either 1) have wincallback.go in this repository or 2) remove the `DO NOT EDIT` comments. I prefer 1).

TotallyGamerJet: I prefer 1 too

hajimehoshi: Sure, I'll send a PR soon later

hajimehoshi: https://github.com/ebitengine/purego/pull/337

hajimehoshi: @limeidan Rebase your PR and modify `wincallback.go` to generate `zcallback_loong64.s`.

limeidan: Done. Please review, thanks.

.github/workflows/test.yml
limeidan: I add tests on versions > 1.19 with cgo enabled, but it reported an error on go1.22 (nosplit stack exceeds 800 byte limit). 
Maybe you can approve CI and see the results. (I don't think it's a code error, but I don't know how to do to save stack space so that the nosplit stack does not exceed its limit)

hajimehoshi: > I add tests on versions > 1.19 with cgo enabled, but it reported an error on go1.22 (nosplit stack exceeds 800 byte limit).

https://github.com/ebitengine/purego/actions/runs/17757918397/job/50468769359?pr=335

Is this issue only for Go 1.22? Do the other versions work?

limeidan: Yes, only on go1.22, the other versions are all work well.

hajimehoshi: If this is difficult to solve, I think testing only > 1.24 is fine, as `CGO_ENABLED=0` already does so. @TotallyGamerJet thoughts?

.github/workflows/test.yml
```diff
@@ -123,6 +130,26 @@ jobs:
           go env -u CC
           go env -u CXX
 
+      - name: go test (Linux loong64 noCgo)
```
hajimehoshi: s/noCgo/no Cgo/

limeidan: Done.

hajimehoshi:
Primary: @TotallyGamerJet and I can maintain
Secondary: maybe not

.github/workflows/test.yml
TotallyGamerJet: Yeah I think it's okay then. Solving it would be a challenge

TotallyGamerJet:
> Primary: @TotallyGamerJet and I can maintain. A release should not include bugs. If there is a ctirical bug before a new release, we have to fix them. Secondary: maybe not. A release can include bugs. Even if there is a critical bug, this is not a release blocker.

That makes sense. I think it's a good idea to be explicit in the README which platforms are primary and secondary and what that means

.github/workflows/test.yml
limeidan: Done.

hajimehoshi:
> That makes sense. I think it's a good idea to be explicit in the README which platforms are primary and secondary and what that means

That's a good idea. I'll try to clarify the policy later, so this PR doesn't have to include this.

.github/workflows/test.yml
```diff
@@ -74,6 +74,13 @@ jobs:
           env GOOS=netbsd GOARCH=amd64 go build -v ./...
           env GOOS=netbsd GOARCH=arm64 go build -v ./...
 
+      - name: go build (Linux loong64)
```
hajimehoshi, line 77: `go build (Linux loong64 w/o Cgo)`

.github/workflows/test.yml
```diff
@@ -123,6 +130,18 @@ jobs:
           go env -u CC
           go env -u CXX
 
+      - name: go test (Linux loong64)
```
hajimehoshi, line 133: `go build (Linux loong64 w/ Cgo)`?

hajimehoshi (approved):
LGTM % nits

.github/workflows/test.yml
limeidan: Sorry, I don't quite understand what you mean.

These two steps (this one and the other one you commented on) are one for cross-compilation testing and the other for functional testing.

Because we discussed only testing for versions > 1.24, I combined the cgo and no-cgo tests together. Perhaps you missed that?

hajimehoshi: Sorry but I thought "go build (Linux loong64)" and "go test (loong64)" seemed the same title. My point was to make different section names, but these are already different. Never mind.

hajimehoshi:
If @TotallyGamerJet is fine, I'll merge this.

TotallyGamerJet (approved):
LGTM

PR #341, opened 2025-09-23
more efficient type asserting

### LINKED ISSUE #340: Replace .Interface().(T) with reflect.TypeAssert
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

`reflect.TypeAssert` avoids unnecessary memory allocations see https://golang.org/doc/go1.25#reflectpkgreflect

The current locations are in the following files:
* func.go
* objc_runtime_darwin.go
* objc_block_darwin.go
Note this requires using Go 1.25 so we will wait on this.

### Why is this needed?

Improve memory usage

TotallyGamerJet:
Closes #340

### DIFF
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -372,7 +372,7 @@ func RegisterClass(name string, superClass Class, protocols []*Protocol, ivars [
 				}
 				id, ok := xreflect.TypeAssert[ID](args[0])
 				if !ok {
-					panic(fmt.Sprintf("objc: id argument is not a ID is %s", args[0].Type().String()))
+					panic(fmt.Sprintf("objc: id argument is not a ID but %s", args[0].Type().String()))
 				}
 				ptr := *(*unsafe.Pointer)(unsafe.Pointer(&id)) // circumvent go vet
 				// the variable is located at an offset from the id


internal/xreflect/reflect_go124.go
```diff
@@ -0,0 +1,12 @@
+//go:build !go1.25
```
hajimehoshi, line 4: Add the copyright comment

internal/xreflect/reflect_go125.go
```diff
@@ -0,0 +1,9 @@
+//go:build go1.25
```
hajimehoshi, line 4: ditto

objc/objc_block_darwin.go
```diff
@@ -167,7 +168,11 @@ func (b *blockCache) getLayout(typ reflect.Type) blockLayout {
 		reflect.MakeFunc(
 			typ,
 			func(args []reflect.Value) (results []reflect.Value) {
-				return b.Functions.Load(args[0].Interface().(Block)).Call(args)
+				block, ok := xreflect.TypeAssert[Block](args[0])
+				if !ok {
+					panic("objc: block argument is not a block")
```
hajimehoshi: What about 

```go
panic(fmt.Sprintf("objc: block argument is not a block but %T", arg[0]))
```

?

objc/objc_runtime_darwin.go
```diff
@@ -341,7 +342,10 @@ func RegisterClass(name string, superClass Class, protocols []*Protocol, ivars [
 				//	})(unsafe.Pointer(args[0].Interface().(ID)))).v = 123
 				//
 				// However, since the type of the variable is unknown reflection is used to actually assign the value
-				id := args[0].Interface().(ID)
+				id, ok := xreflect.TypeAssert[ID](args[0])
+				if !ok {
+					panic("objc: id argument is not a ID")
```
hajimehoshi: ditto

objc/objc_runtime_darwin.go
```diff
@@ -366,7 +370,10 @@ func RegisterClass(name string, superClass Class, protocols []*Protocol, ivars [
 				if len(args) != 2 {
 					panic(fmt.Sprintf("objc: incorrect number of args. expected 2 got %d", len(args)))
 				}
-				id := args[0].Interface().(ID)
+				id, ok := xreflect.TypeAssert[ID](args[0])
+				if !ok {
+					panic("objc: id argument is not a ID")
```
hajimehoshi: ditto

objc/objc_block_darwin.go
TotallyGamerJet: I think we need to do `args[0].Type().String()` since %T would print reflect.Value and not the underlying type

hajimehoshi: Yes, that's right

objc/objc_runtime_darwin.go
hajimehoshi: Remaining?

PR #342, opened 2025-09-29
internal/fakecgo: deduplicate arch-agnostic code

qmuntal:
None

Code clean-up

## What this PR does | solves

`fakecgo` contains multiple functions whose implementation is duplicated verbatim (excluding comments) for every supported GOOS/GOARCH pair. They can be deduplicated into arch-agnostic Go files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ This is a list of the copied files:
 * `internal/fakecgo/freebsd.go` from package `runtime/cgo`
 * `internal/fakecgo/netbsd.go` from package `runtime/cgo`
 
-The `internal/fakecgo/go_GOOS.go` files where modified from `runtime/cgo/gcc_GOOS_GOARCH.go`.
+The `internal/fakecgo/go_GOOS.go` files were modified from `runtime/cgo/gcc_GOOS_GOARCH.go`.
 
 The files `abi_*.h` and `internal/fakecgo/abi_*.h` are the same because Bazel does not support cross-package use of
 `#include` so we need each one once per package. (cf. [issue](https://github.com/bazelbuild/rules_go/issues/3636))


hajimehoshi:
@TotallyGamerJet

TotallyGamerJet:
I don't know if we want to do this. The way fakecgo is structured is to mirror how it's done in the stdlib. Deduplicating could make it more difficult to match what the runtime is doing 

https://github.com/golang/go/tree/master/src/runtime/cgo

internal/fakecgo/go_freebsd.go
TotallyGamerJet, line 1: Let's keep the following comment that's in amd64 and loong64

```
	// runtime/cgo uses __builtin_frame_address(0) instead of `uintptr(unsafe.Pointer(&size))`
	// but this should be OK since we are taking the address of the first variable in this function.
```

internal/fakecgo/go_linux.go
TotallyGamerJet, line 1: Same here

```
// runtime/cgo uses __builtin_frame_address(0) instead of `uintptr(unsafe.Pointer(&size))`
// but this should be OK since we are taking the address of the first variable in this function.
```

TotallyGamerJet (commented):
I've reviewed again and decided it's likely okay. The only one we'd maybe want to keep separated is Darwin/arm64 since there is iOS specific code not implemented. However, Idk if we'll ever implement CGO_ENABLED=0 for iOS so it's likely okay to keep it merged for now.

README.md
```diff
@@ -103,7 +103,7 @@ This is a list of the copied files:
 * `internal/fakecgo/abi_*.h` from package `runtime/cgo`
 * `internal/fakecgo/asm_GOARCH.s` from package `runtime/cgo`
 * `internal/fakecgo/callbacks.go` from package `runtime/cgo`
-* `internal/fakecgo/go_GOOS_GOARCH.go` from package `runtime/cgo`
+* `internal/fakecgo/go_GOOS.go` from package `runtime/cgo`
```
hajimehoshi: I'd like to separate this item from the list and put a sentence indicating that `go_GOOS.go` files were modified from `gcc_GOOS_GOARCH.go` files below the list.

qmuntal: Done.

qmuntal:
Note that I'm submitting a similar change to Go: https://go-review.googlesource.com/c/go/+/707955.

README.md
```diff
@@ -103,11 +103,12 @@ This is a list of the copied files:
 * `internal/fakecgo/abi_*.h` from package `runtime/cgo`
 * `internal/fakecgo/asm_GOARCH.s` from package `runtime/cgo`
 * `internal/fakecgo/callbacks.go` from package `runtime/cgo`
-* `internal/fakecgo/go_GOOS_GOARCH.go` from package `runtime/cgo`
 * `internal/fakecgo/iscgo.go` from package `runtime/cgo`
 * `internal/fakecgo/setenv.go` from package `runtime/cgo`
 * `internal/fakecgo/freebsd.go` from package `runtime/cgo`
 * `internal/fakecgo/netbsd.go` from package `runtime/cgo`
 
+The `internal/fakecgo/go_GOOS.go` files where modified from `runtime/cgo/gcc_GOOS_GOARCH.go`.
```
hajimehoshi: were?

qmuntal: Good catch.

PR #351, opened 2025-10-26
struct: fix bool field placement at non-zero offsets

### LINKED ISSUE #350: Bool fields in structs are placed at wrong bit offset on ARM64/AMD64/LoongArch64
### PureGo Version

v0.10.0-alpha.2

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

N/A

1. Create a struct with a bool field at a non-zero offset (e.g., after a float or int field)
2. Pass this struct by value through `RegisterLibFunc` to a C function
3. The C function will receive the bool field as 0 (false) even when Go passes true

Example code:

```go
package main

import (
)

type FloatAndBool struct {
    Value     float32  // offset 0-3 (4 bytes)
    Has_value bool     // offset 4 (1 byte)
}

func main() {
    // C function: int test_bool(struct FloatAndBool f) { return f.has_value ? 1 : 0; }

    var testFunc func(FloatAndBool) int32
    lib, _ := purego.Dlopen("./libtest.so", purego.RTLD_NOW)
    purego.RegisterLibFunc(&testFunc, lib, "test_bool")

    // Pass struct with Has_value=true
    result := testFunc(FloatAndBool{Value: 123.45, Has_value: true})
    // Result will be 0 instead of 1
}
```

The C function should receive `has_value=1` (true) and return `1`.

The bool field should be placed at its correct bit offset in the register. For a struct like:
- `float value` at bytes 0-3
- `bool has_value` at byte 4

The bool should be placed at bits 32-39 (byte 4), not at bit 0.

### What happens instead?

The C function receives `has_value=0` (false) and returns `0`, even though Go passed `true`.

n `struct_arm64.go`, `struct_amd64.go`, and `struct_loong64.go`, the bool case doesn't apply the bit shift:

```go
case reflect.Bool:
    if f.Bool() {
        val |= 1  // ❌ BUG: Always places bool at bit 0
    }
    shift += 8
```

Should be:
```go
case reflect.Bool:
    if f.Bool() {
        val |= 1 << shift  // ✓ Place bool at correct offset
    }
    shift += 8
```

Compare with the correct uint8 implementation:
```go
case reflect.Uint8:
    val |= f.Uint() << shift  // ✓ Correctly shifted
    shift += 8
```

tmc:
Bool fields were incorrectly placed at bit 0 regardless of their offset within the struct. This caused incorrect marshalling when bool fields appeared after other fields.

The fix properly shifts the bool value by the current bit offset, ensuring it's placed at the correct position in the packed struct data.

Adds regression test with FloatAndBool struct to verify the fix.

Closes #350 

bug

## What this PR does
Fixes offsets of bool fields.

### DIFF
--- a/struct_test.go
+++ b/struct_test.go
@@ -472,6 +472,20 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 			t.Fatalf("TakeGoUintAndReturn returned %+v wanted %+v", ret, expected)
 		}
 	}
+	{
+		type FloatAndBool struct {
+			x float32
+			y bool
+		}
+		var FloatAndBoolFn func(FloatAndBool) int32
+		purego.RegisterLibFunc(&FloatAndBoolFn, lib, "FloatAndBool")
+		if ret := FloatAndBoolFn(FloatAndBool{x: 12345.0, y: true}); ret != 1 {
+			t.Fatalf("FloatAndBool(y: true) = %d, want 1", ret)
+		}
+		if ret := FloatAndBoolFn(FloatAndBool{x: 12345.0, y: false}); ret != 0 {
+			t.Fatalf("FloatAndBool(y: false) = %d, want 0", ret)
+		}
+	}
 }
 
 func TestRegisterFunc_structReturns(t *testing.T) {
--- a/testdata/structtest/struct_test.c
+++ b/testdata/structtest/struct_test.c
@@ -352,3 +352,12 @@ GoUint GoUint4(struct GoUint4 g) {
 GoUint TakeGoUintAndReturn(GoUint a) {
     return a;
 }
+
+struct FloatAndBool {
+    float value;
+    _Bool has_value;
+};
+
+int FloatAndBool(struct FloatAndBool f) {
+    return f.has_value;
+}
--- a/struct_amd64.go
+++ b/struct_amd64.go
@@ -165,7 +165,7 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 				place(f)
 			case reflect.Bool:
 				if f.Bool() {
-					val |= 1
+					val |= 1 << shift
 				}
 				shift += 8
 				class |= _INTEGER
--- a/struct_arm64.go
+++ b/struct_arm64.go
@@ -123,7 +123,7 @@ func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr
 				place(f)
 			case reflect.Bool:
 				if f.Bool() {
-					val |= 1
+					val |= 1 << shift
 				}
 				shift += 8
 				class |= _INT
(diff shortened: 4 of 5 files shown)


struct_test.go
```diff
@@ -472,6 +472,24 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 			t.Fatalf("TakeGoUintAndReturn returned %+v wanted %+v", ret, expected)
 		}
 	}
+	{
+		// Regression test for bool field at non-zero offset
```
TotallyGamerJet: Unnecessary comment

struct_test.go
```diff
@@ -472,6 +472,24 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+		// Before fix: bool was always placed at bit 0 regardless of offset
+		// After fix: bool is correctly shifted to its field offset
+		type FloatAndBool struct {
+			Value     float32
+			Has_value bool
```
TotallyGamerJet: Uncommon to use underscores in Go fields. Call it `HasValue`? doesn't matter if it matches the C name exactly

struct_test.go
```diff
@@ -472,6 +472,24 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+		}
+		var FloatAndBoolFn func(FloatAndBool) int32
+		purego.RegisterLibFunc(&FloatAndBoolFn, lib, "FloatAndBool")
+		if ret := FloatAndBoolFn(FloatAndBool{Value: 12345.0, Has_value: true}); ret != 1 {
+			t.Fatalf("FloatAndBool returned %d wanted 1 (bool field not passed correctly)", ret)
+		}
+		// Also test with false
```
TotallyGamerJet: Unnecessary

testdata/structtest/struct_test.c
```diff
@@ -352,3 +352,20 @@ GoUint GoUint4(struct GoUint4 g) {
 GoUint TakeGoUintAndReturn(GoUint a) {
     return a;
 }
+
+// FloatAndBool tests that bool fields at non-zero offsets are handled correctly
+// This is a regression test for the bug where bool wasn't shifted by the current offset
```
TotallyGamerJet: Remove all the comments in this change

struct_test.go
```diff
@@ -472,6 +472,24 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+		// Before fix: bool was always placed at bit 0 regardless of offset
+		// After fix: bool is correctly shifted to its field offset
+		type FloatAndBool struct {
+			Value     float32
+			Has_value bool
+		}
+		var FloatAndBoolFn func(FloatAndBool) int32
+		purego.RegisterLibFunc(&FloatAndBoolFn, lib, "FloatAndBool")
+		if ret := FloatAndBoolFn(FloatAndBool{Value: 12345.0, Has_value: true}); ret != 1 {
+			t.Fatalf("FloatAndBool returned %d wanted 1 (bool field not passed correctly)", ret)
```
TotallyGamerJet: `FloatAndBool(true) returned %d wanted 1` don't include parentheses part since it's implicit in these tests

PR #353, opened 2025-10-28
Fix Darwin ARM64 calling convention for struct arguments, callbacks, and stack packing

### LINKED ISSUE #352: Support Apple ARM64 ABI for stack arguments
### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms describes how smaller-than-8byte arguments can be represented in darwin arm64 calling conventions.

### Why is this needed?

Supporting this better can allow additional arguments and would fix a current bug where forcing to 8byte alignment actually means arguments are not passed correctly.

tmc:
[Darwin ARM64](https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms) uses byte-aligned argument packing, not 8-byte slots.
The current code uses 8-byte slots for everything, causing two problems:

1. Functions with >8 parameters fail when using int32, float32, etc.
2. Struct arguments are packed incorrectly.

For example:

```go
// C: int32_t sum(int32_t a1, ..., int32_t a11)
var fn func(int32, int32, int32, int32, int32, int32, int32, int32, int32, int32, int32) int32
purego.RegisterLibFunc(&fn, lib, "sum_11_int32")
result := fn(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11)
// fails: stack arguments misaligned
```

The fix implements proper Darwin ARM64 calling conventions:

- Stack arguments use natural alignment (int32 = 4 bytes, not 8).
- Multiple small values pack into single 8-byte slots.
- Struct arguments use byte-level copying, not slot alignment.
- Callbacks unpack arguments using Darwin-specific rules.
- Register allocation tries registers before falling back to stack.

The code splits ARM64 struct handling into darwin and non-darwin versions
to preserve correct behavior on Linux and other ARM64 systems.

Fixes #352

### DIFF
--- a/syscall_other.go
+++ b/syscall_other.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2025 The Ebitengine Authors
 
-//go:build (android || linux) && (386 || arm)
+//go:build linux && !amd64 && !arm64 && !loong64
 
 package purego
 


func.go
```diff
@@ -113,6 +112,10 @@ func RegisterLibFunc(fptr any, handle uintptr, name string) {
 //
 // [Cgo rules]: https://pkg.go.dev/cmd/cgo#hdr-Go_references_to_C
 func RegisterFunc(fptr any, cfn uintptr) {
+	registerFunc(fptr, cfn, false) // false = might be a callback, don't use packing
```
TotallyGamerJet: Why is it that callback's don't get packing? They should match the C ABI of whatever platform they are on.

Also, remove this comment and the one above and place it on the `registerFunc` so it explains why we need packing and not packing for callbacks

stack_packing_darwin_arm64_test.go
```diff
@@ -0,0 +1,325 @@
+// Copyright 2025 The Ebitengine Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build darwin || freebsd || (linux && (amd64 || arm64 || loong64)) || netbsd
```
TotallyGamerJet: Is there a reason these tests are not ran on other platforms since the name excludes all these build tags

stack_packing_darwin_arm64_test.go
```diff
@@ -0,0 +1,325 @@
...
+
+// testFloatResult checks float results with tolerance
+func testFloatResult(t *testing.T, name string, got, want float64, expectedFailOnDarwinARM64 bool) {
+	t.Helper()
+
+	isDarwinARM64 := runtime.GOOS == "darwin" && runtime.GOARCH == "arm64"
+	matches := math.Abs(got-want) < 0.01
+	if !matches {
+		if expectedFailOnDarwinARM64 && isDarwinARM64 {
+			t.Logf("%s: got %v, want %v (expected failure on Darwin ARM64) ⚠️ KNOWN ISSUE", name, got, want)
```
TotallyGamerJet: I don't know if we want emojis in the logs. @hajimehoshi  what do you think?

testdata/stacktest/comprehensive_stack_test.c
```diff
@@ -0,0 +1,214 @@
+// SPDX-License-Identifier: Apache-2.0
+// Comprehensive Darwin ARM64 stack parameter tests
```
TotallyGamerJet: I think we still need to use the standard license header

```
// Copyright 2025 The Ebitengine Authors
```

stack_packing_darwin_arm64_test.go
hajimehoshi: No emoji please. Maybe AI generated them

stack_packing_darwin_arm64_test.go
hajimehoshi: > the name

You mean the file name (`*_darwin_arm64_*`)?

tmc: good point, can clean it up.

func.go
tmc: This caused a test failure and I was concerned about scope but let me think about it and perhaps this could remain more general.

stack_packing_darwin_arm64_test.go
```diff
@@ -0,0 +1,325 @@
...
+			t.Errorf("%s: got %v, want %v ❌ FAIL", name, got, want)
+		}
+	} else {
+		t.Logf("%s: got %v ✓ PASS", name, got)
+	}
+}
+
+// ============================================================================
+// UNIFORM TYPE TESTS - RegisterFunc
+// ============================================================================
```
hajimehoshi: Remove these comments.

testdata/stacktest/comprehensive_stack_test.c
```diff
@@ -0,0 +1,214 @@
...
+// Helper to print test results to stderr
+#define DEBUG_PRINT(fmt, ...) do { \
+    char buf[256]; \
+    int len = snprintf(buf, sizeof(buf), fmt "\n", ##__VA_ARGS__); \
+    write(2, buf, len); \
+} while(0)
+
+// ============================================================================
+// UNIFORM TYPE TESTS - All parameters same size
+// ============================================================================
```
hajimehoshi: Remove these comments.

func.go
```diff
@@ -29,7 +28,7 @@ func RegisterLibFunc(fptr any, handle uintptr, name string) {
 	if err != nil {
 		panic(err)
 	}
-	RegisterFunc(fptr, sym)
+	registerFunc(fptr, sym, true) // true = is a C library function, can use packing
```
hajimehoshi: Remove the comment.

func.go
hajimehoshi: Remove the comment.

testdata/stacktest/comprehensive_stack_test.c
hajimehoshi: The comment should be

```
```

func.go
tmc: On further reflection, supporting this sort of packing would overly complicate the trampoline back to go code which currently expects the 8 byte alignment.

struct_darwin_arm64.go
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
```
hajimehoshi: The filename should be `struct_darwin_arm64.go`

struct_notdarwin_arm64.go
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
```
hajimehoshi: The filename should be `struct_notdarwin_arm64.go`

struct_arm64_other.go
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
+
+//go:build arm64 && !darwin
```
hajimehoshi: `//go:build !darwin`

struct_darwin_arm64.go
```diff
@@ -0,0 +1,57 @@
...
+		var chunk uintptr
+		remaining := size - offset
+		if remaining >= 8 {
+			chunk = *(*uintptr)(unsafe.Add(ptr, offset))
+		} else {
+			// For the last partial chunk, read only the remaining bytes
+			bytes := (*[8]byte)(unsafe.Add(ptr, offset))
+			for i := uintptr(0); i < remaining; i++ {
+				chunk |= uintptr(bytes[i]) << (i * 8)
+			}
```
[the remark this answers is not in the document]
hajimehoshi: I don't think this is necessary

func_darwin_arm64.go
```diff
@@ -0,0 +1,70 @@
...
+
+//go:build darwin && arm64
+
+package purego
+
+import "reflect"
+
+// shouldBundleStackArgs determines if we need to start C-style packing for
+// Darwin ARM64 stack arguments. This happens when registers are exhausted.
+func shouldBundleStackArgs(v reflect.Value, numInts, numFloats int) bool {
```
hajimehoshi: What about moving `shouldBundleStackArgs` and `structFitsInRegisters` to `func.go` and remove `func_darwin_arm64.go` and `func_notdarwin.go`? Adding a check like 

```go
func shouldBundleStackArgs(v reflect.Value, numInts, numFloats int) bool {
    if runtime.GOOS != "darwin" || runtime.GOARCH != "arm64" {
        panic("purego: shouldBundleStackArgs must be called on darwin/arm64")
    }
    // ...
}
```

inside the functions would be nice.

tmc: yes, much nicer.

edit: the point of moving them was to isolate the calls to isHFA/isHFA -- let me see how to lay this out cleanly.

tmc: could do struct_nondarwin.go, struct_darwin_arm64.go, struct_darwin_amd64.go if we don't want to propagate func_* files.

or

struct_nondarwinarm64.go + struct_darwin_arm64.go ?

struct_arm64.go
```diff
@@ -86,7 +86,7 @@ func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFl
 	return keepAlive // the struct was allocated so don't panic
 }
 
-func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
+func placeRegistersAArch64(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
```
hajimehoshi: `placeRegistersArm64`?

func_darwin_arm64.go
hajimehoshi: > the point of moving them was to isolate the calls to isHFA/isHFA 

Confirmed.  What about `struct_darwin_arm64.go` and `struct_others.go`?

tmc:
> Awesome! Would it be possible to split the PR?

Open to that, can you suggest what you'd like to see separated?


edit: I can pull the callback fixes out to a separate PR, sound good?

func.go
```diff
@@ -281,28 +295,123 @@ func RegisterFunc(fptr any, cfn uintptr) {
...
+					}
+
+					structType := reflect.StructOf(fields)
+					structInstance := reflect.New(structType).Elem()
+
+					// Set values (skip padding fields)
+					argIndex := 0
+					for j := 0; j < structInstance.NumField(); j++ {
+						fieldName := structType.Field(j).Name
+						if len(fieldName) < 3 || fieldName[:3] != "Pad" {
```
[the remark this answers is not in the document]
hajimehoshi: I think this is a valid concern.

struct_darwin_arm64.go
```diff
@@ -0,0 +1,119 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
+
+//go:build darwin && arm64
```
hajimehoshi: `//go:build` is not needed for this file

struct_notdarwin_arm64.go
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
+
+//go:build !darwin && arm64
```
hajimehoshi: Just `//go:build !darwin` is fine

hajimehoshi:
> Open to that, can you suggest what you'd like to see separated?

I'll take a look tomorrow, but I'd like to know @TotallyGamerJet's opinion

tmc:
Separating the callback handling would require refactoring some tests to not do the somewhat artificial Go->Go vs the Go->c pathway with a change in how the values are arranged. I've tried to simplify and improve the organization of this so it's easier to review. If it still appears too large let me know and I can work on refactoring the callback tests.

IMO these changes go hand in hand and this rounds out the purego darwin amd64 ABI support.

func.go
```diff
@@ -488,3 +484,31 @@ func numOfIntegerRegisters() int {
...
+		// Check if this goes to register or stack
+		usesInt := arg.Kind() != reflect.Float32 && arg.Kind() != reflect.Float64
+		if usesInt && numInts < numOfIntegerRegisters() {
+			numInts++
+		} else if !usesInt && numFloats < numOfFloatRegisters {
+			numFloats++
+		} else {
+			// Goes to stack - accumulate total bytes
+			stackBytes += size
+		}
```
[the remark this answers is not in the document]
tmc: note: I was opting to simplify this to lower LOC, but agree that more exact checks here are merited.

func.go
```diff
@@ -488,3 +497,63 @@ func numOfIntegerRegisters() int {
...
+	}
+	// Round total to 8-byte boundary
+	if stackBytes > 0 && stackBytes%align8ByteSize != 0 {
+		stackBytes = (stackBytes + align8ByteMask) &^ align8ByteMask
```
TotallyGamerJet: why not use `roundUpTo8` here? Makes it clearer what's happening

func_test.go
```diff
@@ -199,6 +199,237 @@ func TestABI(t *testing.T) {
...
+	if runtime.GOOS == "windows" && runtime.GOARCH == "386" {
+		t.Skip("need a 32bit gcc to run this test") // TODO: find 32bit gcc for test
+	}
+	libFileName := filepath.Join(t.TempDir(), "abitest.so")
+	if err := buildSharedLib("CC", libFileName, filepath.Join("testdata", "abitest", "abi_test.c")); err != nil {
+		t.Fatal(err)
+	}
+	lib, err := load.OpenLibrary(libFileName)
+	if err != nil {
+		t.Fatalf("Dlopen(%q) failed: %v", libFileName, err)
```
TotallyGamerJet: this isn't calling dlopen change to `Failed to open library:`

func_test.go
```diff
@@ -199,6 +199,237 @@ func TestABI(t *testing.T) {
...
+	}
+	t.Cleanup(func() {
+		if err := load.CloseLibrary(lib); err != nil {
+			t.Errorf("Failed to close library: %v", err)
+		}
+	})
+
+	tests := []struct {
+		name string
+		fn   interface{}
```
TotallyGamerJet: change all `interface{}` -> `any`

func_test.go
```diff
@@ -199,6 +199,237 @@ func TestABI(t *testing.T) {
...
+		want string
+	}{
+		{
+			name: "10_int32_baseline",
+			fn:   new(func(*byte, uintptr, int32, int32, int32, int32, int32, int32, int32, int32, int32, int32)),
+			cFn:  "stack_10_int32",
+			call: func(f interface{}) string {
+				buf := make([]byte, 256)
+				(*f.(*func(*byte, uintptr, int32, int32, int32, int32, int32, int32, int32, int32, int32, int32)))(&buf[0], 256, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+				return string(buf[:strings.IndexByte(string(buf), 0)])
```
TotallyGamerJet: Why not use [bytes.IndexByte](https://pkg.go.dev/bytes#IndexByte) to avoid a conversion to string?

struct_darwin_arm64.go
```diff
@@ -0,0 +1,252 @@
...
+	for offset := uintptr(0); offset < size; offset += 8 {
+		// Read 8 bytes (or whatever remains) from the struct
+		var chunk uintptr
+		remaining := size - offset
+		if remaining >= 8 {
+			chunk = *(*uintptr)(unsafe.Add(ptr, offset))
+		} else {
+			// For the last partial chunk, read only the remaining bytes
+			bytes := (*[8]byte)(unsafe.Add(ptr, offset))
+			for i := uintptr(0); i < remaining; i++ {
```
TotallyGamerJet: Instead of looping why not mask off?

```go
mask := uintptr((1 << (remaining * 8)) - 1)
chunk = *(*uintptr)(unsafe.Add(ptr, offset)) & mask
```

struct_darwin_arm64.go
```diff
@@ -0,0 +1,252 @@
...
+				chunk |= uintptr(bytes[i]) << (i * 8)
+			}
+		}
+		addInt(chunk)
+	}
+}
+
+// shouldBundleStackArgs determines if we need to start C-style packing for
+// Darwin ARM64 stack arguments. This happens when registers are exhausted.
+func shouldBundleStackArgs(v reflect.Value, numInts, numFloats int) bool {
```
TotallyGamerJet: I think writing this to avoid multi-level if statements is more clear

```go
func shouldBundleStackArgs(v reflect.Value, numInts, numFloats int) bool {
	kind := v.Kind()
	isFloat := kind == reflect.Float32 || kind == reflect.Float64
	isInt := !isFloat && kind != reflect.Struct
	primitiveOnStack :=
		(isInt && numInts >= numOfIntegerRegisters()) ||
			(isFloat && numFloats >= numOfFloatRegisters)
	if primitiveOnStack {
		return true
	}
	if kind != reflect.Struct {
		return false
	}
	hfa := isHFA(v.Type())
	hva := isHVA(v.Type())
	size := v.Type().Size()
	eligible := hfa || hva || size <= 16
	if !eligible {
		return false
	}

	if hfa {
		need := v.NumField()
		return numFloats+need > numOfFloatRegisters
	}

	if hva {
		need := v.NumField()
		return numInts+need > numOfIntegerRegisters()
	}

	slotsNeeded := int((size + align8ByteMask) / align8ByteSize)
	return numInts+slotsNeeded > numOfIntegerRegisters()
}
```

struct_darwin_arm64.go
```diff
@@ -0,0 +1,252 @@
...
+	for j, val := range stackArgs {
+		valSize := val.Type().Size()
+		valAlign := val.Type().Align()
+
+		// ARM64 requires 8-byte alignment for 8-byte or larger structs
+		if val.Kind() == reflect.Struct && valSize >= 8 {
+			valAlign = 8
+		}
+
+		// Add padding field if needed for alignment
```
TotallyGamerJet: If we are now aligning struct arguments then we should update the [docs](https://pkg.go.dev/github.com/ebitengine/purego#hdr-Structs-RegisterFunc). I'm not sure if we want to be responsible for it though.

struct_darwin_arm64.go
```diff
@@ -0,0 +1,252 @@
...
+	ptr := unsafe.Pointer(structInstance.Addr().Pointer())
+	size := structType.Size()
+	for offset := uintptr(0); offset < size; offset += 8 {
+		var chunk uintptr
+		remaining := size - offset
+		if remaining >= 8 {
+			chunk = *(*uintptr)(unsafe.Add(ptr, offset))
+		} else {
+			// Handle partial chunk at the end
+			bytes := (*[8]byte)(unsafe.Add(ptr, offset))
```
TotallyGamerJet: Again, if doing the masking thing then do it here too.

struct_darwin_arm64.go
```diff
@@ -0,0 +1,252 @@
...
+			continue
+		}
+		structInstance.Field(j).Set(stackArgs[argIndex])
+		argIndex++
+	}
+
+	// Copy struct memory to stack in 8-byte chunks
+	ptr := unsafe.Pointer(structInstance.Addr().Pointer())
+	size := structType.Size()
+	for offset := uintptr(0); offset < size; offset += 8 {
```
TotallyGamerJet: Since we do this twice make it a function? `copyStruct8ByteChunks`

struct_others.go
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
+
+//go:build !darwin || !arm64
```
TotallyGamerJet: I find `struct_others.go` and `struct_other.go` kind of confusing. Can we merge these files and potentially rename it to `struct_notdarwin.go`?

struct_amd64.go
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2024 The Ebitengine Authors
 
+//go:build amd64
```
TotallyGamerJet: Build tag not needed because of file name

struct_arm64.go
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2024 The Ebitengine Authors
 
+//go:build arm64
```
TotallyGamerJet: Build tag not needed

func.go
TotallyGamerJet: @tmc you marked this resolved but I don't see why you didn't use it

func_test.go
TotallyGamerJet: @tmc You resolved this but didn't update the string

func_test.go
TotallyGamerJet: @tmc you didn't update this

struct_darwin_arm64.go
TotallyGamerJet: I don't see the comment updated

syscall_other.go
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
+
+//go:build !darwin && !freebsd && !(linux && (amd64 || arm64 || loong64)) && !netbsd && !windows
```
TotallyGamerJet: I would get rid of this file. I don't think it's needed since it lists all the platforms supported by purego which hints at it not even being used.

struct_darwin_arm64.go
```diff
@@ -0,0 +1,250 @@
...
+	// If the value is not addressable, create an addressable copy
+	if !v.CanAddr() {
+		addressable := reflect.New(v.Type()).Elem()
+		addressable.Set(v)
+		v = addressable
+	}
+	ptr := unsafe.Pointer(v.Addr().Pointer())
+	size := v.Type().Size()
+
+	// Copy the struct memory in 8-byte chunks
```
TotallyGamerJet: This comment seems unnecessary

struct_notdarwin_loong64.go
```diff
@@ -0,0 +1,36 @@
...
+
+package purego
+
+import (
+	"reflect"
+)
+
+// shouldBundleStackArgs always returns false on non-Darwin platforms
+// since C-style stack argument bundling is only needed on Darwin ARM64.
+func shouldBundleStackArgs(v reflect.Value, numInts, numFloats int) bool {
```
TotallyGamerJet: Delete this file and move all the methods into struct_loong64.go

struct_notdarwin.go
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2024 The Ebitengine Authors
+
+//go:build !darwin && !arm64 && !amd64 && !loong64
```
TotallyGamerJet: Can we just name this as the file `struct_arm.go` and a duplicate one `struct_386.go`. These files will make it obvious which platform it is for.

syscall_other.go
TotallyGamerJet: Ahh I see now. This is used on arm and 386. Can we do `//go:build (android || linux) && (386 || arm)`

hajimehoshi:
ping @tmc

func_test.go
```diff
@@ -191,14 +192,245 @@ func TestABI(t *testing.T) {
...
+			t.Errorf("Failed to close library: %v", err)
+		}
+	})
+
+	// Test that 25 int64 arguments (17 slots needed) exceeds the limit
+	t.Run("25_int64_exceeds_limit", func(t *testing.T) {
+		defer func() {
+			if r := recover(); r != nil {
+				msg := fmt.Sprint(r)
+				if !strings.Contains(msg, "too many stack arguments") {
```
hajimehoshi: I feel like testing an error message is fragile. Just checking an error existence should be enough.

struct_386.go
```diff
@@ -0,0 +1,41 @@
...
+func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFloat, addStack func(uintptr), keepAlive []any) []any {
+	panic("purego: struct arguments are not supported")
+}
+
+func getStruct(outType reflect.Type, syscall syscall15Args) (v reflect.Value) {
+	panic("purego: struct returns are not supported")
+}
+
+func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
+	panic("purego: not needed on 386")
```
hajimehoshi: What is not needed? The function itself?

struct_amd64.go
```diff
@@ -262,3 +262,29 @@ func placeStack(v reflect.Value, addStack func(uintptr)) {
 func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
 	panic("purego: not needed on amd64")
```
hajimehoshi: ditto

struct_amd64.go
```diff
@@ -262,3 +262,29 @@ func placeStack(v reflect.Value, addStack func(uintptr)) {
...
+// shouldBundleStackArgs always returns false on non-Darwin platforms
+// since C-style stack argument bundling is only needed on Darwin ARM64.
+func shouldBundleStackArgs(v reflect.Value, numInts, numFloats int) bool {
+	return false
+}
+
+// structFitsInRegisters is not used on non-Darwin platforms.
+// This stub exists for compilation but should never be called.
+func structFitsInRegisters(val reflect.Value, tempNumInts, tempNumFloats int) (bool, int, int) {
+	panic("purego: structFitsInRegisters should not be called on non-Darwin platforms")
```
hajimehoshi: The error message seems odd. Isn't this function defined for both Darwin and non-Darwin?

struct_arm.go
```diff
@@ -0,0 +1,41 @@
...
+func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFloat, addStack func(uintptr), keepAlive []any) []any {
+	panic("purego: struct arguments are not supported")
+}
+
+func getStruct(outType reflect.Type, syscall syscall15Args) (v reflect.Value) {
+	panic("purego: struct returns are not supported")
+}
+
+func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
+	panic("purego: not needed on arm")
```
hajimehoshi: ditto

struct_arm64.go
```diff
@@ -284,3 +297,253 @@ func isHVA(t reflect.Type) bool {
...
+// placeRegisters implements Darwin ARM64 calling convention for struct arguments.
+//
+// For HFA/HVA structs, each element must go in a separate register (or stack slot for elements
+// that don't fit in registers). We use placeRegistersArm64 for this.
+//
+// For non-HFA/HVA structs, Darwin uses byte-level packing. We copy the struct memory in
+// 8-byte chunks, which works correctly for both register and stack placement.
+func placeRegistersDarwin(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
+	if runtime.GOOS != "darwin" {
+		panic("purego: this should only be called on darwin")
```
hajimehoshi: s/this/placeRegistersDarwin/

struct_arm64.go
```diff
@@ -284,3 +297,253 @@ func isHVA(t reflect.Type) bool {
...
+	slotsNeeded := int((size + align8ByteMask) / align8ByteSize)
+	return numInts+slotsNeeded > numOfIntegerRegisters()
+}
+
+// structFitsInRegisters determines if a struct can still fit in remaining
+// registers, used during stack argument bundling to decide if a struct
+// should go through normal register allocation or be bundled with stack args.
+func structFitsInRegisters(val reflect.Value, tempNumInts, tempNumFloats int) (bool, int, int) {
+	if runtime.GOOS != "darwin" {
+		panic("purego: this should only be called on darwin")
```
hajimehoshi: s/this/structFitsInRegisters/

struct_arm64.go
```diff
@@ -284,3 +297,253 @@ func isHVA(t reflect.Type) bool {
...
+
+const (
+	paddingFieldPrefix = "Pad"
+)
+
+// bundleStackArgs bundles remaining arguments for Darwin ARM64 C-style stack packing.
+// It creates a packed struct with proper alignment and copies it to the stack in 8-byte chunks.
+func bundleStackArgs(stackArgs []reflect.Value, addStack func(uintptr)) {
+	if runtime.GOOS != "darwin" {
+		panic("purego: this should only be called on darwin")
```
hajimehoshi: s/this/bundleStackArgs/

struct_amd64.go
TotallyGamerJet: True, it's just not needed on amd64 at all

struct_386.go
TotallyGamerJet: I guess changing it to not implemented is good

syscall_other.go
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
+
+//go:build (android || linux) && (386 || arm)
```
hajimehoshi: * `android` is included by `linux`, so you don't need to specify android.
* This might be problematic when compiling an app with PureGo for minor architectures like Mips. PureGo doesn't have to work for such arches, but at least this should not prevent compilation.

Thus, what about

```go
//go:build linux && !amd64 && !arm64 && !loong64
```

hajimehoshi:
ping @tmc

Issue #356, opened 2025-10-28
darwin: string arguments are duplicated when mixed with int args


tmc:
### PureGo Version

HEAD

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

any

### What steps will reproduce the problem?

```
	var fn func(int32, int32, int32, int32, int32, int32, int32, int32, string, string, string) string
	purego.RegisterLibFunc(&fn, lib, "test_8i32_3strings")

	result := fn(1, 2, 3, 4, 5, 6, 7, 8, "grault", "garply", "waldo")
```

### What is the expected result?

each string is passed

### What happens instead?

string arguments are duplicated: "grault" is passed 3 times

PR #357, opened 2025-10-28
darwin: Fix darwin int string mixing

tmc:
#356 

bug

## What this PR does | solves
Fixes improper use of outer loop variable when handling strings.

Additionally, adds a tool to generate various stack tests.

### DIFF
--- a/func_test.go
+++ b/func_test.go
@@ -186,7 +186,7 @@ func TestABI(t *testing.T) {
 		}
 	}
 	{
-		const cName = "test_8i32_3strings"
+		const cName = "stack_8i32_3strings"
 		var fn func(*byte, uintptr, int32, int32, int32, int32, int32, int32, int32, int32, string, string, string)
 		purego.RegisterLibFunc(&fn, lib, cName)
 		buf := make([]byte, 256)
--- a/testdata/abitest/abi_test.c
+++ b/testdata/abitest/abi_test.c
@@ -26,7 +26,6 @@ uint32_t stack_string(uint32_t a, uint32_t b, uint32_t c, uint32_t d, uint32_t e
     return a | b | c | d | e | f | g | h;
 }
 
-// Test that multiple strings on stack get correct values (not duplicates)
-void test_8i32_3strings(char* result, size_t size, int32_t a1, int32_t a2, int32_t a3, int32_t a4, int32_t a5, int32_t a6, int32_t a7, int32_t a8, const char* s1, const char* s2, const char* s3) {
+void stack_8i32_3strings(char* result, size_t size, int32_t a1, int32_t a2, int32_t a3, int32_t a4, int32_t a5, int32_t a6, int32_t a7, int32_t a8, const char* s1, const char* s2, const char* s3) {
     snprintf(result, size, "%d:%d:%d:%d:%d:%d:%d:%d:%s:%s:%s", a1, a2, a3, a4, a5, a6, a7, a8, s1, s2, s3);
 }


internal/gentests/gen.go
```diff
@@ -0,0 +1,3 @@
+//go:generate go run generate_stacktest_tests.go
+
+package gentests
```
hajimehoshi: ?

internal/gentests/generate_stacktest_tests.go
```diff
@@ -0,0 +1,378 @@
+// Copyright 2025 The Ebitengine Authors
+// SPDX-License-Identifier: Apache-2.0
```
hajimehoshi:
```
```

testdata/abitest/abi_test.c
```diff
@@ -25,3 +25,13 @@ uint32_t stack_string(uint32_t a, uint32_t b, uint32_t c, uint32_t d, uint32_t e
     assert(strcmp(i, "test") == 0);
     return a | b | c | d | e | f | g | h;
 }
+
+// Test that multiple strings on stack get correct values (not duplicates)
+const char* test_8i32_3strings(int32_t a1, int32_t a2, int32_t a3, int32_t a4,
+                                int32_t a5, int32_t a6, int32_t a7, int32_t a8,
+                                const char* s1, const char* s2, const char* s3) {
+    static char result[256];
```
TotallyGamerJet: This doesn’t seem concurrently safe. The Go test is ran multiples times concurrently and this uses the same scratch buffer. Not necessarily an issue since this test currently works but may mask a bug if the result gets overriden

TotallyGamerJet: My suggestion is to pass a Go byte slice in that the C code then writes to

func_test.go
```diff
@@ -185,6 +185,18 @@ func TestABI(t *testing.T) {
 			t.Fatalf("%s: got %d, want %d", cName, res, expect)
 		}
 	}
+	{
+		const cName = "test_8i32_3strings"
+		var fn func(*byte, uintptr, int32, int32, int32, int32, int32, int32, int32, int32, string, string, string)
+		purego.RegisterLibFunc(&fn, lib, cName)
+		buf := make([]byte, 256)
+		fn(&buf[0], 256, 1, 2, 3, 4, 5, 6, 7, 8, "foo", "bar", "baz")
```
TotallyGamerJet: Instead of 256 len(buf)?

hajimehoshi:
> note: This change is included in https://github.com/ebitengine/purego/pull/353

Then I'll make #353 a draft until this PR is merged.

testdata/abitest/abi_test.c
```diff
@@ -25,3 +25,11 @@ uint32_t stack_string(uint32_t a, uint32_t b, uint32_t c, uint32_t d, uint32_t e
     assert(strcmp(i, "test") == 0);
     return a | b | c | d | e | f | g | h;
 }
+
+// Test that multiple strings on stack get correct values (not duplicates)
+void test_8i32_3strings(char* result, size_t size, int32_t a1, int32_t a2, int32_t a3, int32_t a4,
+                        int32_t a5, int32_t a6, int32_t a7, int32_t a8,
+                        const char* s1, const char* s2, const char* s3) {
```
hajimehoshi: The indentation seems unnatural. Let's not have line breaks for arguments.

hajimehoshi: This seems natural when the line `void test_8i32_3strings(...` is shown in one line. I prefer not having line breaks here (unless we have a clear rule) anyway...

func_test.go
hajimehoshi: ping @tmc

testdata/abitest/abi_test.c
```diff
@@ -25,3 +25,8 @@ uint32_t stack_string(uint32_t a, uint32_t b, uint32_t c, uint32_t d, uint32_t e
     assert(strcmp(i, "test") == 0);
     return a | b | c | d | e | f | g | h;
 }
+
+// Test that multiple strings on stack get correct values (not duplicates)
+void test_8i32_3strings(char* result, size_t size, int32_t a1, int32_t a2, int32_t a3, int32_t a4, int32_t a5, int32_t a6, int32_t a7, int32_t a8, const char* s1, const char* s2, const char* s3) {
```
hajimehoshi: Maybe `stack_8i32_3strings` would be better. Also, remove the comment

PR #360, opened 2025-10-30
Fix arm64 argument corruption on alignment flush

### LINKED ISSUE #359: darwin/arm64 struct corruption for homogeneous int32 structs
### PureGo Version

v0.10.0-alpha.2

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

1.25

```go
type FourInt32s struct {
    f0, f1, f2, f3 int32
}

// C function
int32_t FourInt32sSum(struct FourInt32s s) {
    return s.f0 + s.f1 + s.f2 + s.f3;
}

// Go call
var fn func(FourInt32s) int32
purego.RegisterLibFunc(&fn, lib, "FourInt32sSum")
sum := fn(FourInt32s{1, 2, 3, 4})
// Expected: 10 (1+2+3+4)
// Actual: 12 (1+2+3+6) - f3 corrupted from 4 to 6
```

// Expected: 10 (1+2+3+4)

### What happens instead?

// Actual: 12 (1+2+3+6) - f3 corrupted from 4 to 6

### The Bug Chain

1. **`isHVA()` returns `true` for `FourInt32s`**
   - Located in `struct_arm64.go:258-284`
   - Checks if struct has homogeneous integer fields (line 266: `case reflect.Int32`)
   - Returns `true` for any struct with all `int32` fields, size 8 or 16 bytes

2. **`addStruct()` calls `placeRegisters()` for HVAs**
   - Located in `struct_arm64.go:73`
   - Condition: `hva || hfa || size <= 16`
   - Since `isHVA()` returns `true`, it calls `placeRegisters()`

3. **`placeRegisters()` uses `placeRegistersAArch64()` logic**
   - Located in `struct_arm64.go:89-198` (on non-Darwin ARM64)
   - Uses field-by-field bit-shifting and OR operations
   - For `int32` fields (line 155-158):
     ```go
     case reflect.Int32:
         val |= uint64(f.Int()&0xFFFF_FFFF) << shift
         shift += 32
         class |= _INT
     ```

4. **Bit-shifting overflow corrupts the 4th field**
   - After field 0 (f0=1): `val = 0x0000000000000001`, `shift = 32`
   - After field 1 (f1=2): `val = 0x0000000200000001`, `shift = 64`
   - After field 2 (f2=3): `shift = 64` causes flush, new `val = 0x0000000000000003`, `shift = 32`
   - After field 3 (f3=4): `shift = 32`, so `val |= 4 << 32 = 0x0000000400000003`
   - But this is wrong! The flush happened too late, causing f3 to OR with f1's position
   - Result: Second register contains `0x0000000600000003` instead of `0x0000000400000003`

### Memory Layout Evidence

**Expected (byte-level packing):**
```
Chunk 0 (X0): 0x0000000200000001  (f1:f0 = 2:1)
Chunk 1 (X1): 0x0000000400000003  (f3:f2 = 4:3)
```

**Actual (with bug):**
```
Chunk 0 (X0): 0x0000000200000001  (f1:f0 = 2:1)
Chunk 1 (X1): 0x0000000600000003  (f3:f2 = 6:3) <- f3 corrupted!
```

**C side receives:**
```c
byte[12] = 0x06  // Should be 0x04
// Interpreted as f3 = 6 instead of 4
```

tmc:
Fixes ARM64 struct argument corruption when field alignment causes register boundary crossing.

## What type of issue is this addressing?
bug

## What this PR does | solves
Fixes #359 

This fixes a critical bug in ARM64 struct argument passing where field values were being corrupted due to improper state management during register flushes.

### Problem
When passing structs with multiple small integer fields (e.g., `FourInt32s{1, 2, 3, 4}`), field values could be corrupted when struct packing crossed register boundaries. For example, the value `4` was being corrupted to `6`.

### Root Cause
In `struct_arm64.go`, when field alignment caused `shift >= 64`, the code flushed the current register but failed to reset the accumulator (`val`) and classification (`class`) state. This caused subsequent fields to OR with stale data from previous fields.

### Solution
Added two lines in `struct_arm64.go:120-121` to reset state after register flush:
```go
val = 0           // Reset register value
class = _NO_CLASS // Reset classification
```

### Testing
- Added `FourInt32s` test case that reproduces the bug
- All tests pass on Darwin ARM64, Darwin AMD64, Linux ARM64, and Linux AMD64

### DIFF
--- a/struct_test.go
+++ b/struct_test.go
@@ -493,8 +493,8 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 		}
 		var FourInt32sFn func(FourInt32s) int32
 		purego.RegisterLibFunc(&FourInt32sFn, lib, "FourInt32s")
-		result := FourInt32sFn(FourInt32s{100, -50, 25, -75})
-		const want = 100 - 50 + 25 - 75
+		result := FourInt32sFn(FourInt32s{100, -127, 4, -100})
+		const want = 100 - 127 + 4 - 100
 		if result != want {
 			t.Fatalf("FourInt32s returned %d wanted %d", result, want)
 		}
--- a/testdata/structtest/struct_test.c
+++ b/testdata/structtest/struct_test.c
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2024 The Ebitengine Authors
 
-#include "stdint.h"
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 


tmc:
> Let’s separate the fix from the adding Linux struct support. That way it’s better for cherry-picking

done, opened #361 separately.

testdata/structtest/struct_test.c
```diff
@@ -361,3 +361,19 @@ struct FloatAndBool {
...
+    int32_t f1;
+    int32_t f2;
+    int32_t f3;
+};
+
+#include <stdio.h>
+#include <stdlib.h>
+
+char* FourInt32s(struct FourInt32s s) {
+    char* result = malloc(64);
```
TotallyGamerJet: Do we care to free this on the go side?

tmc: I feel like for tests this isn't really needed (and keeps the go side simple).

testdata/structtest/struct_test.c
```diff
@@ -361,3 +361,19 @@ struct FloatAndBool {
...
+
+struct FourInt32s {
+    int32_t f0;
+    int32_t f1;
+    int32_t f2;
+    int32_t f3;
+};
+
+#include <stdio.h>
+#include <stdlib.h>
```
hajimehoshi: #include should be at the top

testdata/structtest/struct_test.c
hajimehoshi: I tend to be OK to not care about memory leaks in the tests, but as we have already done by passing buffers from the Go side in other tests, what about following the same way for consistency?

TotallyGamerJet: If you don't care then it doesn't matter to me

TotallyGamerJet: Also we could just or all the numbers instead and use arguments that all have a different bit so 1,2,4,8 then no string needed

hajimehoshi: Yeah, getting results by integer pointers is better

hajimehoshi:
> could we go the other way and use this (IMO) simpler pattern in the other places?

Are you talking about memory allocation? Please reply this to the comment...

(IMO we should use just pointer values for integers btw. If strings were involed, we could consider a simpler way)

testdata/structtest/struct_test.c
tmc: i did it this way because it's easier to extend to a variety of types across the call boundary using the same pattern.

tmc:
> > could we go the other way and use this (IMO) simpler pattern in the other places?
> 
> Are you talking about memory allocation? Please reply this to the comment...

github ui isn't/wasn't allowing an in-line response for some reason, this was referring to the buffer return suggestion.

> (IMO we should use just pointer values for integers btw. 

I'm open to whatever, was just thinking ahead about how to have common test approaches for a variety of argument structures passed across the boundary, I figured comparing string reprs between the two would provide an easy to implement and easy to compare pattern.

func.go
```diff
@@ -60,7 +60,7 @@ func RegisterLibFunc(fptr any, handle uintptr, name string) {
 //	int64 <=> int64_t
 //	float32 <=> float
 //	float64 <=> double
-//	struct <=> struct (WIP - darwin only)
+//	struct <=> struct (darwin amd64/arm64, linux amd64/arm64)
```
TotallyGamerJet: Did you mean to push the Linux changes? I thought we were keeping them separate

tmc: no thanks for catching, fixed in new push.

tmc: sorry missed a commit, fixing.

struct_test.go
```diff
@@ -373,8 +373,9 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 		}
 		var Array4CharsFn func(chars Array4Chars) int32
 		purego.RegisterLibFunc(&Array4CharsFn, lib, "Array4Chars")
-		if ret := Array4CharsFn(Array4Chars{a: [...]int8{100, -127, 4, -100}}); ret != expectedSigned {
-			t.Fatalf("Array4CharsFn returned %#x wanted %#x", ret, expectedSigned)
+		const expectedSum = 1 + 2 + 4 + 8
+		if ret := Array4CharsFn(Array4Chars{a: [...]int8{1, 2, 4, 8}}); ret != expectedSum {
```
TotallyGamerJet, line 377: Why is this changing? This purposefully uses negative numbers to make sure the signing is done correctly

tmc: see the previous discussion regarding signed-ness of 'char' differing between platforms.

tmc: i'll change the 4 ints to use negatives

hajimehoshi: > see the previous discussion regarding signed-ness of 'char' differing between platforms.

This one? https://github.com/ebitengine/purego/pull/361#discussion_r2481435073

hajimehoshi: IIUC char type had some confusion regarding signs, but with integer types, shouldn't we be able to treat negative values?

tmc: yes, updated/fixed in 711192299083fb6d239b808f25926e3f0a857b20

hajimehoshi: Ah OK thanks,

testdata/structtest/struct_test.c
```diff
@@ -2,6 +2,8 @@
 // SPDX-FileCopyrightText: 2024 The Ebitengine Authors
 
 #include "stdint.h"
```
hajimehoshi: Optional: Use `<>` for consistency

hajimehoshi (approved):
LGTM

struct_test.go
```diff
@@ -486,6 +487,18 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 			t.Fatalf("FloatAndBool(y: false) = %d, want 0", ret)
 		}
 	}
+	{
+		type FourInt32s struct {
+			f0, f1, f2, f3 int32
+		}
+		var FourInt32sFn func(FourInt32s) int32
+		purego.RegisterLibFunc(&FourInt32sFn, lib, "FourInt32s")
+		result := FourInt32sFn(FourInt32s{100, -50, 25, -75})
```
hajimehoshi: Optional: using `100, -127, 4, -100` sounds more random-ish and might be a better test

PR #367, opened 2025-11-11
purego: add dlfcn simulation on windows (#366)

### LINKED ISSUE #366: Add dlfcn simulation on windows
### Operating System

- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

The dlfcn functions(Dlopen, Dlsym and Dlclose) simulation on windows.

### Why is this needed?

I use purego on Linux, MacOS and Windows. However, I have to handle dlopen specially for Windows (need load.go and load_windows.go). Why not add a dlcfcn simulation implementation for Windows? MinGW has already done this, so that the loading operation of the dynamic library can be accomplished with just one function.

I have implemented this feature and am ready to submit the pull request.

In this way, the dynamic library loading can be accomplished as follows:

```go
	var libPath string
	switch runtime.GOOS {
	case "windows":
		libPath = "path/to/library.dll"
	case "linux":
		libPath = "path/to/library.so"
	case "darwin":
		libPath = "path/to/library.dylib"
	}
	library, err := purego.Dlopen(libPath, purego.RTLD_NOW)
	if err != nil {
		panic(err)
	}
	defer purego.Dlclose(library)

	sym, err := purego.Dlsym(library, "myTestFunc")
	if err != nil {
		panic(err)
	}

	var myTestFunc func()
	purego.RegisterFunc(&myTestFunc, sym)

	myTestFunc()
```

xuges:
Closes #366 

feature

## What this PR does | solves
add dlfcn functions(Dlopen, Dlsym and Dlclose) simulation on windows.

### DIFF
--- a/dlfcn_windows.go
+++ b/dlfcn_windows.go
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+package purego
+
+import (
+	"golang.org/x/sys/windows"
+	"path/filepath"
+)
+
+const (
+	RTLD_DEFAULT = 0x00000 // Pseudo-handle for dlsym so search for any loaded symbol
+	RTLD_LAZY    = 0x00001 // Relocations are performed at an implementation-dependent time.
+	RTLD_NOW     = 0x00002 // Relocations are performed when the object is loaded.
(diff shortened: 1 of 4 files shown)


hajimehoshi:
As #366 is marked wontfix, I close this

PR #377, opened 2025-12-16
fakecgo: prefix autogenerate files with z

qmuntal:
It is common an idiomatic for autogenerated file names to start with `z`. This way it is easier to identify them by muscle memory.


TotallyGamerJet (approved):
LGTM

Issue #343, opened 2025-09-29
`syscall.AllThreadsSyscall` panics when using fakecgo


qmuntal:
### PureGo Version

v0.9.0

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go1.25.1

### What steps will reproduce the problem?

Run the following code:

```go
package main

import (
	"fmt"
	"syscall"
)

func main {
	_, _, err := syscall.AllThreadsSyscall(syscall.SYS_FCNTL, 0, 0, 0)
	fmt.Println("error:", err)
}
```

### What is the expected result?

`error: operation not supported`

### What happens instead?

`panic: doAllThreadsSyscall not supported with cgo enabled `

### Anything else you feel useful to add?

[syscall.AllThreadsSyscall](https://pkg.go.dev/syscall#AllThreadsSyscall) returns ``syscall.ENOTSUP` is `cgo_libc_setegid` is defined in `runtime/cgo`. If it is not defined, it goes ahead with the operation but panicking if the `runtime/cgo` defines `iscgo`.

fakecgo does not define `cgo_libc_setegid` but defines `iscgo`, so `syscall.AllThreadsSyscall` panics.

Note that a similar failure mode is reproduced with [syscall.Setegid](https://pkg.go.dev/syscall#Setegid), and friends.

hajimehoshi:
I'm not sure how serious this issue is. Should we cherry-pick the fix to the stable branch?

hajimehoshi:
> Do we want to put this as part of v0.10.1 instead of waiting until v0.11.0?

I tend to be against v0.10.1, as 1. this issue is not a regression, 2. the issue seems niche, and 3. the fix is pretty big.

PR #378, opened 2025-12-17
fakecgo: move setg and call5 out of autogenerated files

qmuntal:
Updates #343

Refactor

## What this PR does | solves
`setg_trampoline` and `call5` function signatures don't need to live in autogenerated files, as they don't depend on any input from `gen.go`.

This will make future enchancements to autogenerated code a bit easier.


hajimehoshi (approved):
LGTM

@TotallyGamerJet

qmuntal:
I can put them in a separate file which would contain functions that don't appear upstream, if you wish.

PR #379, opened 2025-12-17
fakecgo: don't overwrite the caller stack on arm64 trampolines

qmuntal:
Updates https://github.com/ebitengine/purego/issues/343

bug

## What this PR does | solves
amd64 and arm64 assembly trampolines moved from stack arguments to register arguments in #73. This means that we don't need to store the registers on the stack anymore. Note that doing so might even lead to some nasty bugs, as we are writing on stack slots that haven't been allocated in the function frame, potentially corrupting the caller stack.

### DIFF
--- a/internal/fakecgo/trampolines_arm64.s
+++ b/internal/fakecgo/trampolines_arm64.s
@@ -7,7 +7,8 @@
 #include "go_asm.h"
 #include "abi_arm64.h"
 
-// these trampolines map the gcc ABI to Go ABI and then calls into the Go equivalent functions.
+// These trampolines map the gcc ABI to Go ABIInternal and then calls into the Go equivalent functions.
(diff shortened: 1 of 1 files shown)


internal/fakecgo/trampolines_arm64.s
```diff
@@ -10,29 +10,24 @@
 // these trampolines map the gcc ABI to Go ABI and then calls into the Go equivalent functions.
```
TotallyGamerJet: Like this comment says this is technically converting from C to Go ABI. Since in this case the arguments get placed in the same spot is it a good idea to leave a comment say R0, R1 are already in the right spot for Go's ABI?

qmuntal: Good call, done!

PR #384, opened 2026-01-07
fakecgo: improve generation of OS-specific symbols

qmuntal:
Updates #343

feature 

## What this PR does | solves
The `fakecgo` package autogenerates some symbols that are not used on all supported OSes. For example, `pthread_self` is only used on Darwin.

Those symbols shouldn't "leak" into OSes not using them, but they currently do. The Go linker is smart enough to avoid adding them in the final binary, although this is a bit brittle. Regardless, it is confusing to have a symbol defined on a platform that doesn't support it, like `purego_pthread_get_stacksize_np` which is only supported on Darwin.

Note that I plan to grow the list of OS-specific symbols as part of #343, so this problem is only going to get worse.


internal/fakecgo/ztrampolines_darwin.s
```diff
@@ -0,0 +1,19 @@
+// Code generated by 'go generate' with gen.go. DO NOT EDIT.
+
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2022 The Ebitengine Authors
+
+//go:build !cgo && darwin
```
hajimehoshi: As the file name already includes GOOS, we don't have to specify GOOS at `//go:build`.

qmuntal: Good catch. Done.

hajimehoshi:
@TotallyGamerJet

internal/fakecgo/gen.go
```diff
@@ -187,14 +243,15 @@ func run() error {
...
-	t, err = template.New("zsymbols_goos.go").Parse(templateSymbolsGoos)
+	t, err = template.New("zsymbols_goos.go").Funcs(funcs).Parse(templateSymbols)
 	if err != nil {
 		return err
 	}
-	for _, goos := range []string{"darwin", "freebsd", "linux", "netbsd"} {
+	for _, goos := range allOSes {
+		goosSymbols := filterSymbols(allSymbols, goos)
 		f, err = os.Create(fmt.Sprintf("zsymbols_%s.go", goos))
 		defer f.Close()
```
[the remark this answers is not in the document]
hajimehoshi: Another point is, this should have been

```go
f, err = os.Create(...)
if err != nil {
    return err
}
defer f.Close()
```

as `f` is often nil when an error is returned.

On the other hand,

> This could lead to "too many open files" errors if the loop has many iterations.

I don't think we have to care this as `allOSes` is not too many.

internal/fakecgo/gen.go
```diff
@@ -233,6 +294,24 @@ func run() error {
 		if _, err = f.Write(src); err != nil {
 			return err
 		}
+		if len(goosSymbols) == 0 {
+			continue
+		}
+		t, err := template.New("ztrampolines_goos.s").Funcs(funcs).Parse(templateTrampolinesStubs)
```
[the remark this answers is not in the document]
hajimehoshi: IMO it's OK to use `t`

internal/fakecgo/gen.go
```diff
@@ -117,45 +105,109 @@ type Symbol struct {
...
+func contains(slice []string, item string) bool {
+	for _, s := range slice {
+		if s == item {
+			return true
+		}
+	}
+	return false
+}
+
+var allOSes = []string{"darwin", "freebsd", "linux", "netbsd"}
```
TotallyGamerJet: Put this by `funcs` global to keep variables grouped together

internal/fakecgo/gen.go
```diff
@@ -117,45 +105,109 @@ type Symbol struct {
...
+
+type SharedObject struct {
+	Name    string
+	Symbols []Symbol
 }
 
 type LocatedSymbols struct {
-	SharedObject string
-	Symbols      []Symbol
+	Tags          string
```
TotallyGamerJet: I think this should be singular "Tag" according to the docs

https://pkg.go.dev/go/build#hdr-Build_Constraints

internal/fakecgo/gen.go
```diff
@@ -117,45 +105,109 @@ type Symbol struct {
 	Name   string
 	Args   [5]Arg
 	Return string
+	Goos   []string // empty means all GOOSes
```
hajimehoshi: `GOOSes` rather than `Goos`?

internal/fakecgo/gen.go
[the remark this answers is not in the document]
qmuntal: Will improve this code anyway to avoid shadowing.

internal/fakecgo/gen.go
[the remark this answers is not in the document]
qmuntal: Will fix that.

internal/fakecgo/gen.go
hajimehoshi: This is not resolved yet.

qmuntal: It is now 😸

internal/fakecgo/gen.go
```diff
@@ -69,44 +75,29 @@ call5({{.Name}}ABI0,
...
+{{- range .Symbols }}
 //go:linkname _{{.Name}} _{{.Name}}
 var _{{.Name}} uint8
 var {{.Name}}ABI0 = uintptr(unsafe.Pointer(&_{{.Name}}))
 {{ end }}
-`
+`))
 
-const templateTrampolinesStubs = `// Code generated by 'go generate' with gen.go. DO NOT EDIT.
+var templateTrampolinesStubs = template.Must(template.New("trampolines").Funcs(funcs).Parse(
```
TotallyGamerJet: No reason to pass `Funcs(funcs)` here since this template doesn't use any of them

internal/fakecgo/gen.go
```diff
@@ -11,25 +11,31 @@ import (
...
 	"strings"
 	"text/template"
 )
 
-const templateSymbols = `// Code generated by 'go generate' with gen.go. DO NOT EDIT.
+var templateSymbols = template.Must(template.New("symbols").Funcs(funcs).Parse(
+	`// Code generated by 'go generate' with gen.go. DO NOT EDIT.
 
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
```
TotallyGamerJet, line 24: @hajimehoshi Do we need to update this year now that we are creating files that didn't exist before?

hajimehoshi: I don't think we need to update the year. The copyright year doesn't have to be updated. (This is based on Google's rule)

internal/fakecgo/gen.go
qmuntal: Done.

Issue #382, opened 2025-12-26
Support RISCV


nikita322:
### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

### What feature would you like to be added?

Please add support for the RISC architecture

### Why is this needed?

It's an architecture that's gaining popularity, and it would be great if your library supported riscv.

TotallyGamerJet:
I believe Purego works with CGO_ENABLED=1. Although asking no cgo support would be nice

PR #385, opened 2026-01-15
purego: add support for riscv64

gdams:
closes: #382 

feature

## What this PR does | solves
This PR adds the neccessary code logic to support RISCV64 with both CGO enabled/disabled

### DIFF
--- a/sys_unix_riscv64.s
+++ b/sys_unix_riscv64.s
@@ -11,7 +11,7 @@ TEXT callbackasm1(SB), NOFRAME, $0
 	NO_LOCAL_POINTERS
 
 	// On entry, the trampoline in zcallback_riscv64.s left
-	// the callback index in X5.
+	// the callback index in X7.
 
 	// Save callback register arguments X10-X17 and F10-F17.
 	// Stack args (if any) are at 0(SP), 8(SP), etc.
@@ -53,7 +53,7 @@ TEXT callbackasm1(SB), NOFRAME, $0
 	// callbackArgs struct: index(0), args(8), result(16)
 	// Place it at 16(SP) to avoid overlap
 	ADD	$16, SP, X9
-	MOV	X5, 0(X9)     // callback index
+	MOV	X7, 0(X9)     // callback index
 	MOV	X6, 8(X9)     // address of args vector
 	MOV	X0, 16(X9)    // result = 0
 
--- a/syscall_sysv.go
+++ b/syscall_sysv.go
@@ -145,7 +145,6 @@ func callbackWrap(a *callbackArgs) {
 	fnType := fn.Type()
 	args := make([]reflect.Value, fnType.NumIn())
 	frame := (*[callbackMaxFrame]uintptr)(a.args)
-
 	var floatsN int // floatsN represents the number of float arguments processed
 	var intsN int   // intsN represents the number of integer arguments processed
 	// stack points to the index into frame of the current stack element.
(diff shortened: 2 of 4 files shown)


struct_riscv64.go
```diff
@@ -0,0 +1,141 @@
...
+		addInt(v.Addr().Pointer())
+		return keepAlive
+	}
+	ptr := reflect.New(v.Type())
+	ptr.Elem().Set(v)
+	addInt(ptr.Pointer())
+	return append(keepAlive, ptr.Interface())
+}
+
+// RISCV64 does not bundle stack args
```
hajimehoshi: This comment should be inside the function

hajimehoshi (approved):
LGTM % comments, thanks!

struct_riscv64.go
gdams: updated PTAL

.github/workflows/test.yml
```diff
@@ -162,6 +162,38 @@ jobs:
           go env -u CC
           go env -u CXX
 
+  riscv64:
+    strategy:
+      matrix:
+        go: ['1.25.x']
```
TotallyGamerJet: why only 1.25?

internal/fakecgo/trampolines_riscv64.s
```diff
@@ -0,0 +1,71 @@
...
+
+#include "textflag.h"
+#include "go_asm.h"
+
+// these trampolines map the gcc ABI to Go ABI and then calls into the Go equivalent functions.
+
+TEXT x_cgo_init_trampoline(SB), NOSPLIT, $16
+	MOV	X10, 8(SP)
+	MOV	X11, 16(SP)
+	MOV	·x_cgo_init_call(SB), X6
```
TotallyGamerJet: This one uses X6-X7 but other one's use X5/X6. Let's be consistent

struct_riscv64.go
```diff
@@ -0,0 +1,141 @@
...
+func placeSmallAggregateRISCV64(
+	v reflect.Value,
+	addFloat, addInt func(uintptr),
+	keepAlive []any,
+) []any {
+	size := v.Type().Size()
+
+	var ptr unsafe.Pointer
+	if v.CanAddr() {
+		ptr = unsafe.Pointer(v.UnsafeAddr())
```
TotallyGamerJet: Why not [UnsafePointer](https://cs.opensource.google/go/go/+/go1.25.6:src/reflect/value.go;l=2570) [¶](https://pkg.go.dev/reflect#Value.UnsafePointer) to avoid the cast?

struct_riscv64.go
```diff
@@ -0,0 +1,141 @@
...
+	args []reflect.Value,
+	i, numInts, numFloats int,
+	keepAlive []any,
+	addInt, addFloat, addStack func(uintptr),
+	numIntsPtr, numFloatsPtr, numStackPtr *int,
+) ([]reflect.Value, []any) {
+	return nil, keepAlive
+}
+
+func bundleStackArgs(stackArgs []reflect.Value, addStack func(uintptr)) {}
```
TotallyGamerJet: panic here in case this does get called?

.github/workflows/test.yml
gdams: good point, I'll restructure this a little to test on all go versions.

gdams: Noting that it only seems to work for go 1.21 onwards. 1.20 and below is missing several riscv specific patches

gdams:
@hajimehoshi I've pushed a fix. This should be good to go now

PR #387, opened 2026-01-15
purego: bug fix: the register number in a generator file was wrong

hajimehoshi:
n/a

bug

## What this PR does | solves
The register number in a generator file wincallback.go and a generated file zcallback_loong64.s were not synchronized correctly. This was introduced in the original PR for loong64: 045732660bc4b298c428c8d98716c8eec4266289

The generated file by `go generate` and the current files are correctly synchronized after this change.


TotallyGamerJet (approved):
LGTM

Issue #336, opened 2025-09-09
Add support for arm7 platform


cppcoffee:
### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

### What feature would you like to be added?

Add support for arm7 platform

Issue #369, opened 2025-12-02
cross compilation for 32bit Linux


hajimehoshi:
### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

### What feature would you like to be added?

Now compiling PureGo for 32bit Linux without Cgo (i.e. `CGO_ENABLED=0`) fails:

```
# github.com/ebitengine/purego
Error: ./func.go:323:20: undefined: syscall15XABI0
Error: ./func.go:334:20: undefined: syscall15XABI0
Error: ./func.go:338:32: undefined: syscall_syscall15X
Error: ./func.go:404:10: undefined: NewCallback
Error: ./syscall.go:55:9: undefined: syscall_syscall15X
Error: Process completed with exit code 1.
```

PR #391, opened 2026-01-21
purego: add support for linux/arm32

gdams:
fixes: https://github.com/ebitengine/purego/issues/336
updates: https://github.com/ebitengine/purego/issues/369

feature

## What this PR does | solves
This PR adds the neccessary code logic to support linux/arm32 with both CGO enabled/disabled

### DIFF
--- a/internal/fakecgo/asm_arm.s
+++ b/internal/fakecgo/asm_arm.s
@@ -8,42 +8,45 @@
 // func crosscall2(fn, a unsafe.Pointer, n int32, ctxt uintptr)
 // Saves C callee-saved registers and calls cgocallback with three arguments.
 // fn is the PC of a func(a unsafe.Pointer) function.
-TEXT crosscall2(SB),NOSPLIT|NOFRAME,$0
-	SUB	$(8*9), R13 // Reserve space for the floating point registers.
+TEXT crosscall2(SB), NOSPLIT|NOFRAME, $0
+	SUB $(8*9), R13 // Reserve space for the floating point registers.
+
 	// The C arguments arrive in R0, R1, R2, and R3. We want to
 	// pass R0, R1, and R3 to Go, so we push those on the stack.
 	// Also, save C callee-save registers R4-R12.
-	MOVM.WP	[R0, R1, R3, R4, R5, R6, R7, R8, R9, g, R11, R12], (R13)
+	MOVM.WP [R0, R1, R3, R4, R5, R6, R7, R8, R9, g, R11, R12], (R13)
+
 	// Finally, save the link register R14. This also puts the
 	// arguments we pushed for cgocallback where they need to be,
 	// starting at 4(R13).
-	MOVW.W	R14, -4(R13)
+	MOVW.W R14, -4(R13)
 
 	// Save VFP callee-saved registers D8-D15 (same as S16-S31).
 	// Note: We always save these since we target hard-float ABI.
-	MOVD	F8, (13*4+8*1)(R13)
-	MOVD	F9, (13*4+8*2)(R13)
-	MOVD	F10, (13*4+8*3)(R13)
-	MOVD	F11, (13*4+8*4)(R13)
-	MOVD	F12, (13*4+8*5)(R13)
-	MOVD	F13, (13*4+8*6)(R13)
-	MOVD	F14, (13*4+8*7)(R13)
-	MOVD	F15, (13*4+8*8)(R13)
+	MOVD F8, (13*4+8*1)(R13)
+	MOVD F9, (13*4+8*2)(R13)
+	MOVD F10, (13*4+8*3)(R13)
+	MOVD F11, (13*4+8*4)(R13)
+	MOVD F12, (13*4+8*5)(R13)
+	MOVD F13, (13*4+8*6)(R13)
+	MOVD F14, (13*4+8*7)(R13)
+	MOVD F15, (13*4+8*8)(R13)
 
-	BL	runtime·load_g(SB)
+	BL runtime·load_g(SB)
+
 	// We set up the arguments to cgocallback when saving registers above.
-	BL	runtime·cgocallback(SB)
+	BL runtime·cgocallback(SB)
 
-	MOVD	(13*4+8*1)(R13), F8
-	MOVD	(13*4+8*2)(R13), F9
-	MOVD	(13*4+8*3)(R13), F10
-	MOVD	(13*4+8*4)(R13), F11
-	MOVD	(13*4+8*5)(R13), F12
-	MOVD	(13*4+8*6)(R13), F13
-	MOVD	(13*4+8*7)(R13), F14
-	MOVD	(13*4+8*8)(R13), F15
+	MOVD (13*4+8*1)(R13), F8
+	MOVD (13*4+8*2)(R13), F9
+	MOVD (13*4+8*3)(R13), F10
+	MOVD (13*4+8*4)(R13), F11
+	MOVD (13*4+8*5)(R13), F12
+	MOVD (13*4+8*6)(R13), F13
+	MOVD (13*4+8*7)(R13), F14
+	MOVD (13*4+8*8)(R13), F15
 
-	MOVW.P	4(R13), R14
-	MOVM.IAW	(R13), [R0, R1, R3, R4, R5, R6, R7, R8, R9, g, R11, R12]
-	ADD	$(8*9), R13
-	MOVW	R14, R15
+	MOVW.P   4(R13), R14
+	MOVM.IAW (R13), [R0, R1, R3, R4, R5, R6, R7, R8, R9, g, R11, R12]
+	ADD      $(8*9), R13
+	MOVW     R14, R15
--- a/internal/fakecgo/trampolines_arm.s
+++ b/internal/fakecgo/trampolines_arm.s
@@ -13,30 +13,30 @@
 TEXT x_cgo_init_trampoline(SB), NOSPLIT, $8-0
 	MOVW R0, 4(R13)
 	MOVW R1, 8(R13)
-	MOVW    ·x_cgo_init_call(SB), R12
-	MOVW    (R12), R12
-	CALL    (R12)
+	MOVW ·x_cgo_init_call(SB), R12
+	MOVW (R12), R12
+	CALL (R12)
 	RET
 
 TEXT x_cgo_thread_start_trampoline(SB), NOSPLIT, $8-0
 	MOVW R0, 4(R13)
-	MOVW    ·x_cgo_thread_start_call(SB), R12
-	MOVW    (R12), R12
-	CALL    (R12)
+	MOVW ·x_cgo_thread_start_call(SB), R12
+	MOVW (R12), R12
+	CALL (R12)
 	RET
 
 TEXT x_cgo_setenv_trampoline(SB), NOSPLIT, $8-0
 	MOVW R0, 4(R13)
-	MOVW    ·x_cgo_setenv_call(SB), R12
-	MOVW    (R12), R12
-	CALL    (R12)
+	MOVW ·x_cgo_setenv_call(SB), R12
+	MOVW (R12), R12
+	CALL (R12)
 	RET
 
 TEXT x_cgo_unsetenv_trampoline(SB), NOSPLIT, $8-0
 	MOVW R0, 4(R13)
-	MOVW    ·x_cgo_unsetenv_call(SB), R12
-	MOVW    (R12), R12
-	CALL    (R12)
+	MOVW ·x_cgo_unsetenv_call(SB), R12
+	MOVW (R12), R12
+	CALL (R12)
 	RET
 
 TEXT x_cgo_notify_runtime_init_done_trampoline(SB), NOSPLIT, $0-0
@@ -51,29 +51,31 @@ TEXT x_cgo_bindm_trampoline(SB), NOSPLIT, $0
 TEXT ·setg_trampoline(SB), NOSPLIT, $0-8
 	MOVW G+4(FP), R0
 	MOVW setg+0(FP), R12
-	BL (R12)
+	BL   (R12)
 	RET
 
 TEXT threadentry_trampoline(SB), NOSPLIT, $8-0
 	// See crosscall2.
 	MOVW R0, 4(R13)
-	MOVW    ·threadentry_call(SB), R12
-	MOVW    (R12), R12
-	CALL    (R12)
+	MOVW ·threadentry_call(SB), R12
+	MOVW (R12), R12
+	CALL (R12)
 	RET
 
 TEXT ·call5(SB), NOSPLIT, $8-28
-    MOVW    fn+0(FP), R12
-    MOVW    a1+4(FP), R0
-    MOVW    a2+8(FP), R1
-    MOVW    a3+12(FP), R2
-    MOVW    a4+16(FP), R3
-    MOVW    a5+20(FP), R4
-    // Store 5th arg below SP (in local frame area)
-    MOVW    R4, arg5-8(SP)
-    // Align SP to 8 bytes for call (required by ARM AAPCS)
-    SUB     $8, R13
-	CALL   	(R12)
-    ADD     $8, R13
-    MOVW    R0, r1+24(FP)
-    RET
+	MOVW fn+0(FP), R12
+	MOVW a1+4(FP), R0
+	MOVW a2+8(FP), R1
+	MOVW a3+12(FP), R2
+	MOVW a4+16(FP), R3
+	MOVW a5+20(FP), R4
+
+	// Store 5th arg below SP (in local frame area)
+	MOVW R4, arg5-8(SP)
+
+	// Align SP to 8 bytes for call (required by ARM AAPCS)
+	SUB  $8, R13
+	CALL (R12)
+	ADD  $8, R13
+	MOVW R0, r1+24(FP)
+	RET
--- a/sys_arm.s
+++ b/sys_arm.s
@@ -30,112 +30,113 @@ TEXT syscall15X(SB), NOSPLIT|NOFRAME, $0-0
 	// Called via C calling convention: R0 = pointer to syscall15Args
 	// NOT via Go calling convention
 	// Save link register and callee-saved registers first
-	MOVW.W	R14, -4(R13)	// save LR (decrement and store)
-	MOVM.DB.W [R4, R5, R6, R7, R8, R9, R11], (R13)  // save callee-saved regs
+	MOVW.W    R14, -4(R13)                         // save LR (decrement and store)
+	MOVM.DB.W [R4, R5, R6, R7, R8, R9, R11], (R13) // save callee-saved regs
 
-	MOVW	R0, R8
-	SUB	$STACK_SIZE, R13
-	MOVW	R8, PTR_ADDRESS(R13)
+	MOVW R0, R8
+	SUB  $STACK_SIZE, R13
+	MOVW R8, PTR_ADDRESS(R13)
 
 	// Load function pointer first (before anything can corrupt R8)
-	MOVW	syscall15Args_fn(R8), R5
-	MOVW	R5, (PTR_ADDRESS-4)(R13)	// save fn at offset 56
+	MOVW syscall15Args_fn(R8), R5
+	MOVW R5, (PTR_ADDRESS-4)(R13) // save fn at offset 56
 
 	// Load floating point arguments
 	// Each float64 spans 2 uintptr slots (8 bytes) on ARM32, so we skip by 2
-	MOVD	syscall15Args_f1(R8), F0	// f1+f2 -> D0
-	MOVD	syscall15Args_f3(R8), F1	// f3+f4 -> D1
-	MOVD	syscall15Args_f5(R8), F2	// f5+f6 -> D2
-	MOVD	syscall15Args_f7(R8), F3	// f7+f8 -> D3
-	MOVD	syscall15Args_f9(R8), F4	// f9+f10 -> D4
-	MOVD	syscall15Args_f11(R8), F5	// f11+f12 -> D5
-	MOVD	syscall15Args_f13(R8), F6	// f13+f14 -> D6
-	MOVD	syscall15Args_f15(R8), F7	// f15+f16 -> D7
+	MOVD syscall15Args_f1(R8), F0  // f1+f2 -> D0
+	MOVD syscall15Args_f3(R8), F1  // f3+f4 -> D1
+	MOVD syscall15Args_f5(R8), F2  // f5+f6 -> D2
+	MOVD syscall15Args_f7(R8), F3  // f7+f8 -> D3
+	MOVD syscall15Args_f9(R8), F4  // f9+f10 -> D4
+	MOVD syscall15Args_f11(R8), F5 // f11+f12 -> D5
+	MOVD syscall15Args_f13(R8), F6 // f13+f14 -> D6
+	MOVD syscall15Args_f15(R8), F7 // f15+f16 -> D7
 
 	// Load integer arguments into registers (R0-R3 for ARM EABI)
-	MOVW	syscall15Args_a1(R8), R0	// a1
-	MOVW	syscall15Args_a2(R8), R1	// a2
-	MOVW	syscall15Args_a3(R8), R2	// a3
-	MOVW	syscall15Args_a4(R8), R3	// a4
+	MOVW syscall15Args_a1(R8), R0 // a1
+	MOVW syscall15Args_a2(R8), R1 // a2
+	MOVW syscall15Args_a3(R8), R2 // a3
+	MOVW syscall15Args_a4(R8), R3 // a4
 
 	// push a5-a32 onto stack
-	MOVW	syscall15Args_a5(R8), R4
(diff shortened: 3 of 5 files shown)


wincallback.go
```diff
@@ -100,8 +100,37 @@ TEXT callbackasm(SB),NOSPLIT|NOFRAME,$0
 	}
 }
 
+func genasmArm() {
```
hajimehoshi: Move this above genasmArm64

hajimehoshi: This is not fixed yet.

gdams: I pushed a fix with it?

hajimehoshi: This function definition has not been moved yet, right?

gdams: oh sorry I understand what you mean now, I thought you meant the calling order

gdams: updated

gdams:
> This also needs to update README, which was updated at [828401e](https://github.com/ebitengine/purego/commit/828401e0a508c122614e9109dece0182b6541397).

updated

internal/fakecgo/go_util.go
```diff
@@ -28,8 +28,10 @@ func x_cgo_thread_start(arg *ThreadStart) {
 		abort()
 	}
 	// *ts = *arg would cause a writebarrier so copy using slices
-	s1 := unsafe.Slice((*uintptr)(unsafe.Pointer(ts)), unsafe.Sizeof(*ts)/8)
-	s2 := unsafe.Slice((*uintptr)(unsafe.Pointer(arg)), unsafe.Sizeof(*arg)/8)
+	// Use unsafe.Sizeof(uintptr(0)) to work on both 32-bit and 64-bit
```
hajimehoshi: I don't think this comment is needed.

callback_test.go
```diff
@@ -1,14 +1,15 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023 The Ebitengine Authors
 
-//go:build darwin || (linux && (amd64 || arm64 || loong64 || riscv64))
+//go:build darwin || (linux && (amd64 || arm64 || arm || loong64 || riscv64))
```
hajimehoshi: Order the build tags like this:

```go
//go:build darwin || (linux && (amd64 || arm || arm64 || loong64 || riscv64))
```

internal/cgo/syscall_cgo_unix.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-//go:build freebsd || (linux && !(arm64 || amd64 || loong64 || riscv64)) || netbsd
+//go:build freebsd || (linux && !(arm64 || amd64 || arm || loong64 || riscv64)) || netbsd
```
hajimehoshi: Reorder the build tags like this

```go
//go:build freebsd || (linux && !(arm64 || amd || arm64 || loong64 || riscv64)) || netbsd
```

func.go
```diff
@@ -537,9 +550,9 @@ func estimateStackBytes(ty reflect.Type) int {
 // tight packing as normal C function calls.
 func isCallbackFunction(cfn uintptr) bool {
 	// Only platforms with syscall_sysv.go have callback detection.
-	// Match the build constraint: darwin || freebsd || (linux && (amd64 || arm64 || loong64 || riscv64)) || netbsd
+	// Match the build constraint: darwin || freebsd || (linux && (amd64 || arm64 || arm || loong64 || riscv64)) || netbsd
```
hajimehoshi: Reorder the build tags like this

```go
// Match the build constraint: darwin || freebsd || (linux && (amd64 || arm || arm64 || loong64 || riscv64)) || netbsd
```

func.go
```diff
@@ -537,9 +550,9 @@ func estimateStackBytes(ty reflect.Type) int {
...
 	hasSyscallSysv := runtime.GOOS == "darwin" || runtime.GOOS == "freebsd" || runtime.GOOS == "netbsd" ||
-		(runtime.GOOS == "linux" && (runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64" || runtime.GOARCH == "loong64" || runtime.GOARCH == "riscv64"))
+		(runtime.GOOS == "linux" && (runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64" || runtime.GOARCH == "arm" || runtime.GOARCH == "loong64" || runtime.GOARCH == "riscv64"))
```
hajimehoshi: ditto

syscall_cgo_linux.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-//go:build cgo && !(amd64 || arm64 || loong64 || riscv64)
+//go:build cgo && !(amd64 || arm64 || arm || loong64 || riscv64)
```
hajimehoshi: ditto

syscall_sysv.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2022 The Ebitengine Authors
 
-//go:build darwin || freebsd || (linux && (amd64 || arm64 || loong64 || riscv64)) || netbsd
+//go:build darwin || freebsd || (linux && (amd64 || arm64 || arm || loong64 || riscv64)) || netbsd
```
hajimehoshi: ditto

callback_test.go
```diff
@@ -51,6 +52,9 @@ func TestCallGoFromSharedLib(t *testing.T) {
 }
 
 func TestNewCallbackFloat64(t *testing.T) {
+	if runtime.GOARCH == "arm" || runtime.GOARCH == "386" {
+		t.Skip("float callbacks not supported on 32-bit platforms")
```
hajimehoshi: Doesn't this work before this PR? (If so, we would have to update README. @TotallyGamerJet )

gdams: no floats were never supported, we actually throw a panic if the tests are called on 32 bit platforms: https://github.com/ebitengine/purego/blob/828401e0a508c122614e9109dece0182b6541397/func.go#L171

Because I added the arm constraint to the top of the `callback_test.go` we are now running some of the tests. For what it's worth I can probably get floats to work for linux arm32 now we have the assembly but figured it would make sense to do this in a follow-up PR.

gdams: @hajimehoshi ignore this, I've managed to add support via some pretty minimal changes.

callback_test.go
```diff
@@ -125,6 +126,9 @@ func TestNewCallbackFloat32(t *testing.T) {
 }
 
 func TestNewCallbackFloat32AndFloat64(t *testing.T) {
+	if runtime.GOARCH == "arm" {
+		t.Skip("too many arguments for 32-bit purego (max 16 slots)")
```
hajimehoshi: Is this panic message valid?

gdams: I actually managed to fix this so I'll unskip the test

syscall_sysv.go
```diff
@@ -149,32 +151,35 @@ func callbackWrap(a *callbackArgs) {
...
+			slots = int((fnType.In(i).Size() + ptrSize - 1) / ptrSize)
+			if floatsN+slots > numOfFloatRegisters() {
 				pos = stack
-				stack++
+				stack += slots
 			} else {
 				pos = floatsN
 			}
-			floatsN++
+			floatsN += slots
```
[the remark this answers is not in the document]
gdams: added some comments to clarify

sys_arm.s
```diff
@@ -0,0 +1,140 @@
...
+
+	// Load floating point arguments
+	MOVD	syscall15Args_f1(R8), F0	// f1
+	MOVD	syscall15Args_f3(R8), F1	// f2
+	MOVD	syscall15Args_f5(R8), F2	// f3
+	MOVD	syscall15Args_f7(R8), F3	// f4
+	MOVD	syscall15Args_f9(R8), F4	// f5
+	MOVD	syscall15Args_f11(R8), F5	// f6
+	MOVD	syscall15Args_f13(R8), F6	// f7
+	MOVD	syscall15Args_f15(R8), F7	// f8
```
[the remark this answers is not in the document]
gdams: Not correct this time Copilot:

On ARM32, `uintptr` is 4 bytes, but MOVD loads/stores 8 bytes (64-bit double precision float). The syscall15Args struct has f1 through f16 as sequential `uintptr` fields, so each float64 value spans two consecutive fields:

f1 + f2 (8 bytes total) → holds one float64 → loaded into D0/F0
f3 + f4 (8 bytes total) → holds one float64 → loaded into D1/F1
f5 + f6 (8 bytes total) → holds one float64 → loaded into D2/F2
... and so on
Loading from syscall15Args_f1, syscall15Args_f3, syscall15Args_f5, etc. correctly advances by 8 bytes each time (skipping 2 slots), which is exactly what we need for 64-bit floats.

I tried the suggested change and it caused test failures (`TestNewCallbackFloat64`, `TestNewCallbackFloat32`, etc.) because it was only advancing by 4 bytes per load, causing overlapping/incorrect float values.

I've added comments in the code to clarify this behavior.

internal/fakecgo/trampolines_arm.s
```diff
@@ -0,0 +1,79 @@
...
+// These trampolines map the gcc ABI to Go ABI0 and then call into the Go equivalent functions.
+// On ARM32, Go ABI0 uses stack-based calling convention.
+// Arguments are placed on the stack starting at 4(SP) after the prologue.
+
+TEXT x_cgo_init_trampoline(SB), NOSPLIT, $8-0
+	MOVW R0, 4(R13)
+	MOVW R1, 8(R13)
+	MOVW    ·x_cgo_init_call(SB), R12
+	MOVW    (R12), R2
+	CALL    (R2)
```
qmuntal: You can reuse R12 in both instructions. The same applied to other trampolines.

```suggestion
	MOVW    (R12), R12
	CALL    (R12)
```

qmuntal (changes requested):

internal/fakecgo/trampolines_arm.s
gdams: updated

internal/fakecgo/trampolines_arm.s
```diff
@@ -0,0 +1,79 @@
...
+
+TEXT x_cgo_bindm_trampoline(SB), NOSPLIT, $0
+	CALL ·x_cgo_bindm(SB)
+	RET
+
+// func setg_trampoline(setg uintptr, g uintptr)
+TEXT ·setg_trampoline(SB), NOSPLIT, $0-8
+	MOVW G+4(FP), R0
+	MOVW setg+0(FP), R1
+	BL (R1)
```
qmuntal: For consistency:

```suggestion
	MOVW setg+0(FP), R12
	BL (R12)
```

internal/fakecgo/trampolines_arm.s
```diff
@@ -0,0 +1,79 @@
...
+    MOVW    a1+4(FP), R0
+    MOVW    a2+8(FP), R1
+    MOVW    a3+12(FP), R2
+    MOVW    a4+16(FP), R3
+    MOVW    a5+20(FP), R4
+    // Store 5th arg below SP (in local frame area)
+    MOVW    R4, arg5-8(SP)
+    // Align SP to 8 bytes for call (required by ARM AAPCS)
+    SUB     $8, R13
+	CALL   	(R5)
```
qmuntal: For consistency:

```suggestion
    MOVW    fn+0(FP), R12
    MOVW    a1+4(FP), R0
    MOVW    a2+8(FP), R1
    MOVW    a3+12(FP), R2
    MOVW    a4+16(FP), R3
    MOVW    a5+20(FP), R4
    // Store 5th arg below SP (in local frame area)
    MOVW    R4, arg5-8(SP)
    // Align SP to 8 bytes for call (required by ARM AAPCS)
    SUB     $8, R13
	CALL   	(R12)
```

qmuntal (commented):
Missed this in my previous review 😓

internal/fakecgo/asm_arm.s
```diff
@@ -0,0 +1,49 @@
...
+	MOVD	F9, (13*4+8*2)(R13)
+	MOVD	F10, (13*4+8*3)(R13)
+	MOVD	F11, (13*4+8*4)(R13)
+	MOVD	F12, (13*4+8*5)(R13)
+	MOVD	F13, (13*4+8*6)(R13)
+	MOVD	F14, (13*4+8*7)(R13)
+	MOVD	F15, (13*4+8*8)(R13)
+
+	BL	runtime·load_g(SB)
+	// We set up the arguments to cgocallback when saving registers above.
```
hajimehoshi, line 37: Let's avoid `we`. Just `Set up...` is fine.

func.go
```diff
@@ -231,7 +231,7 @@ func RegisterFunc(fptr any, cfn uintptr) {
 
 	v := reflect.MakeFunc(ty, func(args []reflect.Value) (results []reflect.Value) {
 		var sysargs [maxArgs]uintptr
-		var floats [numOfFloatRegisters]uintptr
+		var floats [maxArgs]uintptr
```
hajimehoshi: Why is this changed?

sys_arm.s
```diff
@@ -0,0 +1,141 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
```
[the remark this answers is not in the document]
gdams: Copilot fail 🤦🏼

internal/fakecgo/asm_arm.s
gdams: This is a direct lift and shift of the code from the stdlib: https://github.com/golang/go/blob/2d37c2077854e225204a7c4c26c10453a20de6d1/src/runtime/cgo/asm_arm.s#L50

It feels odd to deviate from it unless necessary?

func.go
gdams: `numOfFloatRegisters` is no longer a constant, and we want to keep that code path allocation-free. The only way to achieve that is using the `maxArgs` constant, which is safe to use because functions can't take more float parameters than allowed by `maxArgs`.

internal/fakecgo/asm_arm.s
hajimehoshi: Ah OK, then let's keep it, sorry

func.go
hajimehoshi: Then let's leave a comment explaining why `maxArgs` is used instead of `numOfFloatRegisters`

gdams: added a comment

hajimehoshi: We need to explain that maxArgs is always >= numOfFloatRegisters() and that's why maxArgs always works

hajimehoshi: Another idea is to make numOfFloatRegisters constant by adding `func_*.go` files for each architecture.

gdams: > We need to explain that maxArgs is always >= numOfFloatRegisters() and that's why maxArgs always works

I'll add a clarifying comment

> Another idea is to make numOfFloatRegisters constant by adding func_*.go files for each architecture.

I did try this quite early on my implementation but it felt like overkill to me? Happy to do this though if you'd prefer this approach.

hajimehoshi: I'm fine with both. Let's keep the current way then. Thanks!

func.go
```diff
@@ -382,7 +389,11 @@ func RegisterFunc(fptr any, cfn uintptr) {
 		case reflect.Float64:
 			// NOTE: syscall.r2 is only the floating return value on 64bit platforms.
 			// On 32bit platforms syscall.r2 is the upper part of a 64bit return.
-			v.SetFloat(math.Float64frombits(uint64(syscall.f1)))
+			if unsafe.Sizeof(uintptr(0)) == 4 {
```
hajimehoshi: Put `const is32bit = unsafe.Sizeof(uintptr(0)) == 4` in somewhere in this file and use it. This definition already exists in a function local, so we can also replace it. https://github.com/ebitengine/purego/blob/a280c0d3e73538f6cb53273dd8c343d9da954463/func.go#L169

EDIT: This would conflict with the definition at [dlfcn_android.go](https://github.com/ebitengine/purego/blob/a280c0d3e73538f6cb53273dd8c343d9da954463/dlfcn_android.go#L12). So, let's define the const as a local at this function instead.

dlfcn_android.go
```diff
@@ -8,13 +8,13 @@ import "github.com/ebitengine/purego/internal/cgo"
...
-	RTLD_DEFAULT = is32bit * 0xffffffff
+	is64bits     = 1 << (^uintptr(0) >> 63) / 2
+	is32bits     = 1 - is64bits
+	RTLD_DEFAULT = is32bits * 0xffffffff
 	RTLD_LAZY    = 0x00000001
-	RTLD_NOW     = is64bit * 0x00000002
+	RTLD_NOW     = is64bits * 0x00000002
 	RTLD_LOCAL   = 0x00000000
-	RTLD_GLOBAL  = is64bit*0x00100 | is32bit*0x00000002
+	RTLD_GLOBAL  = is64bits*0x00100 | is32bits*0x00000002
```
hajimehoshi: Renaming this to avoid confliction doesn't make sense.

func.go
```diff
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	align8ByteMask = 7 // Mask for 8-byte alignment: (val + 7) &^ 7
-	align8ByteSize = 8 // 8-byte alignment boundary
+	align8ByteMask = 7                              // Mask for 8-byte alignment: (val + 7) &^ 7
+	align8ByteSize = 8                              // 8-byte alignment boundary
+	is32bit        = unsafe.Sizeof(uintptr(0)) == 4 // Is this a 32-bit architecture
```
hajimehoshi: Put this at the function local to avoid confliction

gdams: agreed, I've moved it

syscall_other.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2025 The Ebitengine Authors
 
-//go:build linux && !amd64 && !arm64 && !loong64 && !riscv64
+//go:build linux && !amd64 && !arm64 && !arm && !loong64 && !riscv64
```
hajimehoshi: Reorder this

```go
//go:build linux && !amd64 && !arm && !arm64 && !loong64 && !riscv64
```

qmuntal (approved):

internal/fakecgo/go_util.go
```diff
@@ -28,8 +28,9 @@ func x_cgo_thread_start(arg *ThreadStart) {
 		abort()
 	}
 	// *ts = *arg would cause a writebarrier so copy using slices
-	s1 := unsafe.Slice((*uintptr)(unsafe.Pointer(ts)), unsafe.Sizeof(*ts)/8)
-	s2 := unsafe.Slice((*uintptr)(unsafe.Pointer(arg)), unsafe.Sizeof(*arg)/8)
+	ptrSize := unsafe.Sizeof(uintptr(0))
```
TotallyGamerJet:
```suggestion
	const ptrSize = unsafe.Sizeof(uintptr(0))
```

testdata/abitest/abi_test.c
```diff
@@ -125,7 +125,7 @@ void stack_20_int32(char *buf, size_t bufsize, int32_t a1, int32_t a2, int32_t a
...
-    snprintf(buf, bufsize, "%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64 ":%" PRId64,
-             a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20, a21, a22, a23, a24, a25);
+void stack_35_int64_exceeds(char *buf, size_t bufsize, int64_t a1, int64_t a2, int64_t a3, int64_t a4, int64_t a5, int64_t a6, int64_t a7, int64_t a8, int64_t a9, int64_t a10, int64_t a11, int64_t a12, int64_t a13, int64_t a14, int64_t a15, int64_t a16, int64_t a17, int64_t a18, int64_t a19, int64_t a20, int64_t a21, int64_t a22, int64_t a23, int64_t a24, int64_t a25, int64_t a26, int64_t a27, int64_t a28, int64_t a29, int64_t a30, int64_t a31, int64_t a32, int64_t a33, int64_t a34, int64_t a35) {
```
TotallyGamerJet: Just curious why this is being changed?

struct_arm.go
```diff
@@ -3,18 +3,62 @@
...
 
 func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFloat, addStack func(uintptr), keepAlive []any) []any {
-	panic("purego: struct arguments are not supported")
+	size := v.Type().Size()
+	if size == 0 {
+		return keepAlive
+	}
+
+	// ARM EABI: small structs are passed in registers or on stack
+	// For simplicity, pass by pointer for now
```
TotallyGamerJet, line 18: This should likely be a TODO since it's purposefully not implementing the entirety of the ARM calling convention

struct_arm.go
```diff
@@ -3,18 +3,62 @@
...
+		return reflect.NewAt(outType, unsafe.Pointer(&struct{ a, b uintptr }{syscall.a1, syscall.a2})).Elem()
+	}
+	// Larger structs returned via pointer in a1
+	return reflect.NewAt(outType, *(*unsafe.Pointer)(unsafe.Pointer(&syscall.a1))).Elem()
 }
 
 func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
-	panic("purego: placeRegisters not implemented on arm")
+	// For ARM32, just pass the struct data directly
+	// This is a simplified implementation
```
TotallyGamerJet, line 50: ditto on the TODO comment

testdata/abitest/abi_test.c
TotallyGamerJet: Oh I see now, it's because the syscall15Args struct was expanded but this shouldn't affect the number of args right since each uintptr is only 32bits and each of these args are 64bit? There wasn't a change to any of the other architectures they are still limited to 16 args but the maxArgs has increased.

syscall.go
```diff
@@ -13,14 +13,14 @@ package purego
 type CDecl struct{}
 
 const (
-	maxArgs             = 15
-	numOfFloatRegisters = 8 // arm64 and amd64 both have 8 float registers
+	maxArgs = 32
```
TotallyGamerJet: Like my comment above stated this seems wrong since right now the 64bit architectures only support 15 in the assembly. Perhaps move this to a `syscall_32bit.go` file that way the numOfFloatRegisters can stay const and there is no mixup between the 32bit and 64bit syscall15Args

testdata/abitest/abi_test.c
gdams: Correct

syscall.go
gdams: agreed, I've pushed a refactor. Let me know what you think.

testdata/abitest/abi_test.c
gdams: I've reverted this now I've refactored the code

internal/fakecgo/trampolines_arm.s
```diff
@@ -0,0 +1,79 @@
...
+// These trampolines map the gcc ABI to Go ABI0 and then call into the Go equivalent functions.
+// On ARM32, Go ABI0 uses stack-based calling convention.
+// Arguments are placed on the stack starting at 4(SP) after the prologue.
+
+TEXT x_cgo_init_trampoline(SB), NOSPLIT, $8-0
+	MOVW R0, 4(R13)
+	MOVW R1, 8(R13)
+	MOVW    ·x_cgo_init_call(SB), R12
+	MOVW    (R12), R12
+	CALL    (R12)
```
TotallyGamerJet: Looks like some of these lines in this file aren't formatted the same. You can run `https://github.com/klauspost/asmfmt` to keep it consistent

gdams: Turns out most of the files needed formatting. I've pushed the results

hajimehoshi: Hmm? We have not run https://github.com/klauspost/asmfmt so now this PR modified unrelated .s files. If we do, we should have a separate PR to just format .s files to minimize the changes... But did we agree with using asmfmt?

gdams: let me back this out and only format the arm32 files. I'll then raise a separate PR

gdams: okay updated PTAL

hajimehoshi: Thanks. I'm looking forward to the following up PR to format asm files, thanks!

PR #393, opened 2026-01-22
Simplify and better document callbackasm for riscv64

qmuntal:
Since Go 1.26, MOV instructions with immediate values lower than or equal to 32 are encoded in 2 bytes rather than 4 bytes (see https://github.com/golang/go/issues/47560), which breaks the assumption that each callback entry in `callbackasm` is 8 bytes long. Therefore, for callback indices less than or equal to 32, we add a PCALIGN directive to align the next instruction to an 8-byte boundary.

This was already done in #385. This PR documents why PCALIGN are necessary in `callbackasm` and also remove all unnecessary PCALIGN instructions.

### DIFF
--- a/wincallback.go
+++ b/wincallback.go
@@ -112,7 +112,7 @@ func genasmRiscv64() {
 // Since Go 1.26, MOV instructions with immediate values lower than or equal to 32
 // are encoded in 2 bytes rather than 4 bytes, which breaks the assumption that each
 // callback entry is 8 bytes long. Therefore, for callback indices less than or equal to 32,
-// we add a PCALIGN directive to align the next instruction to an 8-byte boundary.
+// add a PCALIGN directive to align the next instruction to an 8-byte boundary.
 // The MOV instruction loads X7 with the callback index, and the
 // JMP instruction branches to callbackasm1.
 // callbackasm1 takes the callback index from X7 and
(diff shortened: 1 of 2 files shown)


gdams:
Much better! thank you

wincallback.go
```diff
@@ -109,6 +109,10 @@ func genasmRiscv64() {
 
 // External code calls into callbackasm at an offset corresponding
 // to the callback index. Callbackasm is a table of MOV and JMP instructions.
+// Since Go 1.26, MOV instructions with immediate values lower than or equal to 32
+// are encoded in 2 bytes rather than 4 bytes, which breaks the assumption that each
+// callback entry is 8 bytes long. Therefore, for callback indices less than or equal to 32,
+// we add a PCALIGN directive to align the next instruction to an 8-byte boundary.
```
hajimehoshi: Who are we? (The Go team?)

zcallback_riscv64.s
```diff
@@ -4,6 +4,10 @@
 
 // External code calls into callbackasm at an offset corresponding
 // to the callback index. Callbackasm is a table of MOV and JMP instructions.
+// Since Go 1.26, MOV instructions with immediate values lower than or equal to 32
+// are encoded in 2 bytes rather than 4 bytes, which breaks the assumption that each
+// callback entry is 8 bytes long. Therefore, for callback indices less than or equal to 32,
+// we add a PCALIGN directive to align the next instruction to an 8-byte boundary.
```
hajimehoshi: ditto

hajimehoshi (approved):
LGTM, thanks!

PR #395, opened 2026-01-22
fakecgo: use C ABI in arm64/loong64/riscv64 trampolines

qmuntal:
Same as #394 but for arm64/loong64/riscv64.


TotallyGamerJet (approved):
LGTM

Issue #399, opened 2026-01-28
reduce per-call overhead in RegisterFunc and callback dispatches


tmc:
### Operating System

- [ ] Windows
- [x] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

### What feature would you like to be added?

A reduction in overhad for purego calls

### Why is this needed?

RegisterFunc/CFunc/15args costs ~1845 ns with 44 allocs. Profiling shows the hot spots are repeated reflect.StructOf in Darwin ARM64 stack bundling (rebuilds struct type every call), heap-allocated []reflect.Value in callbackWrap (38.5% of total allocs), deferred cleanup closures, and per-call reflect.Kind dispatch. All of this is internal — no API changes needed.

PR #441, opened 2026-03-18
purego: avoid reflect.MakeFunc in RegisterFunc fast path

tmc:
#399 

feature

## What this PR does | solves

`RegisterFunc` currently uses `reflect.MakeFunc` to build closures, which allocates per-call for argument packing. This PR detects common signatures (up to 15 integer/pointer/bool args, with optional float32/float64 args) and installs concrete typed closures that call the assembly trampoline directly — zero allocations on the hot path.

**Scope:**
- Pure integer/pointer/bool signatures (1-15 args): zero-alloc concrete closures
- Trailing float32 or float64 (e.g. `func(int64, int64, float32)`): zero-alloc concrete closures
- Interleaved floats (e.g. `func(int64, float32, int64)`): `reflect.MakeFunc` wrapper that still avoids per-arg boxing
- Non-Windows, amd64/arm64 only (Windows uses `syscall.Syscall15` which takes a different path)
- SyscallN and RoundTrip paths are unchanged

## Results

Apple M4 Max, arm64, Go 1.24, benchstat -count=8 -benchtime=5s:

### RegisterFunc calling C functions (direct call, no callback)

```
                   │  main ns/op  │  fast ns/op  │   delta    │ allocs old → new │
1 arg                   145          26            -82%         3 → 0
2 args                  167          26            -84%         4 → 0
3 args                  184          27            -86%         5 → 0
5 args                  226          29            -87%         7 → 0
10 args                 828          30            -96%        22 → 0
14 args                1735          53            -97%        40 → 0
15 args                1904          72            -96%        43 → 0
```

### RegisterFunc calling C functions (with Go callback round-trip)

```
                   │  main ns/op  │  fast ns/op  │   delta    │ allocs old → new │
1 arg                   347         165            -52%         5 → 2
2 args                  527         206            -61%         7 → 3
3 args                  835         232            -72%         8 → 3
5 args                  707         270            -62%        10 → 3
10 args                1290         400            -69%        25 → 3
14 args                2215         508            -77%        43 → 3
15 args                2358         592            -75%        46 → 3
```

### SyscallN / RoundTrip

No statistically significant change (expected — these paths are not modified).


func.go
```diff
@@ -218,6 +218,73 @@ func RegisterFunc(fptr any, cfn uintptr) {
...
+		numFloats := 0
+		numInts := 0
+		trailingFloats := true // whether all floats are trailing (no ints after floats)
+		seenFloat := false
+		intRegCount := numOfIntegerRegisters()
+		for i := 0; i < numIn; i++ {
+			switch ty.In(i).Kind() {
+			case reflect.Uintptr, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
+				reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+				reflect.Bool, reflect.Ptr, reflect.UnsafePointer:
```
hajimehoshi: Use `reflect.Pointer`

fast_func.go
```diff
@@ -0,0 +1,1103 @@
...
+func registerFastFunc(fn reflect.Value, cfn uintptr, numArgs int, hasReturn bool) {
+	switch numArgs {
+	case 0:
+		if hasReturn {
+			impl := func() uintptr { return fastCall0(cfn) }
+			forceFuncPtr(fn, &impl)
+		} else {
+			impl := func() { fastCall0(cfn) }
+			forceFuncPtr(fn, &impl)
+		}
```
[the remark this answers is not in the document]
tmc: The GC concern does not apply here. Unlike SyscallN where the user explicitly passes uintptr (requiring go:uintptrescapes), in the fast path the user calls through the original typed signature -- e.g. myFunc(somePtr). The caller stack frame holds the typed *T / unsafe.Pointer and keeps it live across the call. forceFuncPtr swaps the function pointer so the closure receives the bits as uintptr, but runtime_cgocall runs synchronously on the same goroutine, so the caller typed pointer cannot be collected or moved during the C call.

Adding go:uintptrescapes to the fastCallN functions would not help anyway -- by the time they are called, the value is already a uintptr. The annotation only matters at the boundary where user code converts a pointer to uintptr, and that boundary here is the caller stack frame which retains the typed reference.

fast_func.go
```diff
@@ -0,0 +1,1103 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
+
+//go:build darwin || freebsd || linux || netbsd || windows
```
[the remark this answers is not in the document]
tmc: Good point. Will tighten the build tag to exclude Windows since the fast path is runtime-gated to non-Windows anyway.

func.go
```diff
@@ -218,6 +218,73 @@ func RegisterFunc(fptr any, cfn uintptr) {
...
+				if seenFloat {
+					trailingFloats = false
+				}
+				// Stack args must be pointer-sized for the 8-byte slot asm.
+				if numInts >= intRegCount && ty.In(i).Size() < unsafe.Sizeof(uintptr(0)) {
+					canFastPath = false
```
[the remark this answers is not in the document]
tmc: Same as the reply above -- the caller retains the typed pointer on its stack frame across the synchronous runtime_cgocall. The fast path eligibility check including Ptr/UnsafePointer is safe because the type erasure via forceFuncPtr does not affect the caller GC roots.

gen.go
```diff
@@ -4,3 +4,4 @@
 package purego
 
 //go:generate go run wincallback.go
+//go:generate go run gen_fast_func.go
```
hajimehoshi: I prefer `genfastfunc.go`

gen_fast_func.go
```diff
@@ -0,0 +1,883 @@
...
+			buf.WriteString("\t\t\t} else {\n")
+
+			fmt.Fprintf(buf, "\t\t\t\timpl := func(%s) {\n", interleavedParams(totalArgs, floatPos))
+			fmt.Fprintf(buf, "\t\t\t\t\tints := [%d]uintptr{%s}\n", cfg.MaxIntArgs, interleavedIntArgs(totalArgs))
+			buf.WriteString("\t\t\t\t\tvar floats [8]uintptr\n")
+			buf.WriteString("\t\t\t\t\tfloats[0] = uintptr(math.Float32bits(f1))\n")
+			fmt.Fprintf(buf, "\t\t\t\t\tfastCallIF(cfn, ints, %d, floats, 1)\n", numInts)
+			buf.WriteString("\t\t\t\t}\n")
+			buf.WriteString("\t\t\t\tforceFuncPtr(fn, &impl)\n")
+			buf.WriteString("\t\t\t}\n")
```
hajimehoshi: Why not using ` string literal?

gen_fast_func.go
```diff
@@ -0,0 +1,883 @@
...
+			fmt.Fprintf(buf, "\t\t\tif ty.In(i).Kind() != %s {\n", floatKind)
+			buf.WriteString("\t\t\t\tallMatch = false\n")
+			buf.WriteString("\t\t\t\tbreak\n")
+			buf.WriteString("\t\t\t}\n")
+			buf.WriteString("\t\t}\n")
+			buf.WriteString("\t\tif allMatch {\n")
+			fmt.Fprintf(buf, "\t\t\tregisterFastFunc%s(fn, cfn, numInts, hasReturn)\n", p.Name)
+			buf.WriteString("\t\t\treturn true\n")
+			buf.WriteString("\t\t}\n")
+			buf.WriteString("\t}\n\n")
```
hajimehoshi: ditto

genfastfunc.go
```diff
@@ -0,0 +1,720 @@
...
+			buf.WriteString("\ts.a9, s.a10, s.a11, s.a12, s.a13, s.a14, s.a15 = 0, 0, 0, 0, 0, 0, 0\n")
+		}
+
+		buf.WriteString("\ts.f1, s.f2, s.f3, s.f4, s.f5, s.f6, s.f7, s.f8 = 0, 0, 0, 0, 0, 0, 0, 0\n")
+		buf.WriteString("\ts.arm64_r8 = 0\n")
+		buf.WriteString("\truntime_cgocall(syscall15XABI0, unsafe.Pointer(s))\n")
+		buf.WriteString("\tr := s.a1\n")
+		buf.WriteString("\tthePool.Put(s)\n")
+		buf.WriteString("\treturn r\n")
+		buf.WriteString("}\n\n")
```
hajimehoshi: Can we use backquote string literals for multiple lines generations in this file, so that the maintaining would be easier?

tmc: yes

hajimehoshi:
Hmm, would it be difficult to use [text/template](text/template) for code generations?

tmc:
Adopted text/template -- open to any suggestions to improve readability/maintainability

hajimehoshi:
The tests fail now

tmc:
> The tests fail now

fixed the gofmt issue

gen.go
```diff
@@ -4,3 +4,4 @@
 package purego
 
 //go:generate go run wincallback.go
+//go:generate go run genfastfunc.go
```
hajimehoshi, line 7: Adding `//go:generate gofmt -s -w .` would be safer

genfastfunc.go
```diff
@@ -0,0 +1,855 @@
...
+		}
+	default:
+		s.a1 = ints[0]
+		s.a2 = ints[1]
+		s.a3 = ints[2]
+		s.a4 = ints[3]
+		s.a5 = ints[4]
+		s.a6 = ints[5]
+		s.a7 = ints[6]
+		s.a8 = 0
```
hajimehoshi, line 338: `ints[7]` is ignored, but is this intentional?

genfastfunc.go
```diff
@@ -0,0 +1,855 @@
...
+{{- end}}
+		}
+	default:
+		s.a1 = ints[0]
+		s.a2 = ints[1]
+		s.a3 = ints[2]
+		s.a4 = ints[3]
+		s.a5 = ints[4]
+		s.a6 = ints[5]
+		s.a7 = ints[6]
```
hajimehoshi, line 337: This assumes that ints[n] where n >= numInts is zero cleared, but is my understanding correct?

genfastfunc.go
```diff
@@ -0,0 +1,855 @@
...
+		return false
+	}
+
+	canFastPath := true
+	numFloats := 0
+	numInts := 0
+	hasStructs := false
+	hasSmallStackArg := false
+	trailingFloats := true
+	seenFloat := false
```
hajimehoshi: I prefer using `var foo int` or `var foo bool` for default values.

```go
var numFloats int
var numInts int
var hasStructs bool
var hasSmallStackArg bool
var seenFloat bool
trailingFloats := true
```

func.go
```diff
@@ -437,6 +441,151 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
...
+		case reflect.Bool:
+			out.SetBool(byte(r) != 0)
+		case reflect.UnsafePointer:
+			out.SetPointer(*(*unsafe.Pointer)(unsafe.Pointer(&r)))
+		case reflect.Pointer:
+			out = reflect.NewAt(ty.Out(0), unsafe.Pointer(&r)).Elem()
+		}
+		if len(args) > 0 {
+			args[0] = out
+			return args[:1]
```
hajimehoshi, line 584: Is it safe to reuse the slice of arguments? (just out of curiousity)

func.go
```diff
@@ -437,6 +441,151 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
...
+			}
+		case reflect.Int64, reflect.Int:
+			*(*int64)(unsafe.Pointer(&buf[off])) = f.Int()
+		case reflect.Uint64, reflect.Uint, reflect.Float64:
+			if f.Kind() == reflect.Float64 {
+				*(*uint64)(unsafe.Pointer(&buf[off])) = math.Float64bits(f.Float())
+			} else {
+				*(*uint64)(unsafe.Pointer(&buf[off])) = f.Uint()
+			}
+		case reflect.Struct:
```
hajimehoshi, line 523: Should we handle `reflect.Uintptr` in this switch?

genfastfunc.go
tmc: yes this is in the "numInts<8" branch

genfastfunc.go
tmc: yes - callers always use var ints [15]uintptr or [15]uintptr{...} composite literals, both of which zero-fill remaining elements per the Go spec

func.go
tmc: yes -- not touched by this pr but can fix here

func.go
tmc: callReflect allocates a fresh args slice per call (make([]Value, 0, numIn)) and never accesses it after the callback returns - it only reads the returned out slice. Reusing args[:1] avoids a []reflect.Value heap allocation per call. I believe this always been the case it's technically an undocumented implementation detail rather than a stated behavior.

hajimehoshi: OK thanks

genfastfunc.go
hajimehoshi: Acknowledged

genfastfunc.go
hajimehoshi: As `ints` is an argument, zeroing the values is the caller's responsibility, so this is a little confusing. At least we need a comment to explain that this assumes unused values are zero.

tmc: i can add that comment

hajimehoshi (approved):
LGTM, thanks!

@TotallyGamerJet PTAL

func.go
```diff
@@ -437,6 +441,153 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
...
+}
+
+// flattenStructToUintptr packs a small register-passed struct into one uintptr.
+func flattenStructToUintptr(v reflect.Value) uintptr {
+	var buf [8]byte
+	flattenStructFields(v, buf[:])
+	return *(*uintptr)(unsafe.Pointer(&buf[0]))
+}
+
+func flattenStructFields(v reflect.Value, buf []byte) {
```
TotallyGamerJet, line 487: Why isn't this just `copy(buf, unsafe.Slice((*byte)(v.UnsafePointer()), v.Type().Size()))`?

func.go
```diff
@@ -437,6 +441,153 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
...
+		return false
+	}
+	return true
+}
+
+// flattenStructToUintptr packs a small register-passed struct into one uintptr.
+func flattenStructToUintptr(v reflect.Value) uintptr {
+	var buf [8]byte
+	flattenStructFields(v, buf[:])
+	return *(*uintptr)(unsafe.Pointer(&buf[0]))
```
TotallyGamerJet, line 484: This definitely breaks unsafe rules. We can't return a pointer to a slice that may get GCd

TotallyGamerJet:
I'm not sure I feel comfortable with this change. It's a lot of extremely dense code to parse and review. Even the generator code is difficult. If I can't review it I can't be confident there isn't parts of it that break unsafe rules. I believe I've already found one in my cursory review. I lean towards rejection.

hajimehoshi:
Hmm, as this can be a maintainer burden, we should not go with this route. @tmc What do you think? Is this code complexity inevitable?

tmc:
I'm.looking into that unsafe violation now.

I'm using (attempting to use) purego for some ML operations and need zero allocation paths to make performance viable.

Open to any ideas for how to make this achievable, this has been my best cut at it so far.

tmc:
> I'm not sure I feel comfortable with this change. It's a lot of extremely dense code to parse and review. Even the generator code is difficult. If I can't review it I can't be confident there isn't parts of it that break unsafe rules. I believe I've already found one in my cursory review. I lean towards rejection.

That’s fair. I found and am fixing a real liveness gap in one of the Reflect fallback paths while reviewing this, so I agree the PR is carrying too much unsafe/ABI surface to review comfortably as-is.

I’m going to reduce the scope and make the safety argument more explicit. In particular, I’ll keep the mechanically generated pieces aligned with focused tests, avoid unsafe byte-buffer casts in the struct path, and either remove pointer signatures from the forceFuncPtr fast path or document a tighter argument before asking you to review that part again.

tmc:
I can work around this with a go:linkname kludge and may open a PR to expose a lower level api that is zero alloc (or just vendor or re-implement). Closing.

PR #445, opened 2026-04-30
feat: add fixed-arity `Syscall0`–`Syscall15` variants to avoid heap allocation

RomainMuller:
## Motivation

`SyscallN` is variadic (`args ...uintptr`). When called across a module
boundary the compiler must materialise the args as a `[]uintptr` slice at
the call site. Because of `//go:uintptrescapes` the compiler cannot prove
the slice doesn't escape, so it is heap-allocated even for small, short-lived
call sites (anyhow, since the call is cross-module, the slice always escapes).

This allocation can represent a significant cost (time + GC churn) when calls are frequently made using the low-level API. Having fixed-arity variants helps alleviate this and can result in a 75% overhead reduction (almost entirely from the slice allocation).

## What this PR does

Adds `Syscall0` through `Syscall15` — fixed-arity wrappers with explicit
named `uintptr` parameters. No slice is created at the call site; the
individual arguments are passed in registers, eliminating the allocation
entirely.

Internally each wrapper follows the same path as `SyscallN`:
- a zero-initialised `[maxArgs]uintptr` array for integer args and one for
  float args are allocated on the wrapper's own stack frame,
- the explicit arguments are assigned into those arrays,
- the existing `syscall_SyscallN` internal is called (no new assembly),
- the Windows path delegates to `syscall_syscallN` as before.

`//go:uintptrescapes` is retained on all wrappers for the same GC-pinning
reason it is present on `SyscallN`.

## Testing

Two C helpers are added to `testdata/abitest/abi_test.c`:
- `stack_0_uintptr()` — returns the constant `42` (smoke-tests `Syscall0`)
- `stack_15_uintptr(a1..a15)` — returns the sum of its 15 arguments

A `syscall_fixed` sub-test inside `TestABI_ArgumentPassing` exercises every
arity (0–15), checks the absolute expected value, and verifies parity with
the equivalent `SyscallN` call.

### DIFF
--- a/func_test.go
+++ b/func_test.go
@@ -469,6 +469,117 @@ func TestABI_ArgumentPassing(t *testing.T) {
 		}
 	})
 
+	t.Run("syscall_fixed", func(t *testing.T) {
+		fn0, err := load.OpenSymbol(lib, "stack_0_uintptr")
+		if err != nil {
+			t.Fatalf("OpenSymbol(stack_0_uintptr) failed: %v", err)
+		}
+		fn15, err := load.OpenSymbol(lib, "stack_15_uintptr")
+		if err != nil {
+			t.Fatalf("OpenSymbol(stack_15_uintptr) failed: %v", err)
+		}
+
+		// stack_0_uintptr takes no arguments and returns the constant 42.
+		{
+			got, _, _ := purego.Syscall0(fn0)
+			gotN, _, _ := purego.SyscallN(fn0)
+			if got != 42 {
+				t.Errorf("Syscall0: got %d, want 42", got)
+			}
+			if got != gotN {
+				t.Errorf("Syscall0 vs SyscallN: %d != %d", got, gotN)
+			}
+		}
+
+		// stack_15_uintptr sums its 15 arguments. Calling it with K explicit
+		// arguments (the remaining slots are zero-initialised) yields sum(1..K).
+		check := func(name string, got, gotN uintptr, want int) {
+			t.Helper()
+			if got != uintptr(want) {
+				t.Errorf("%s: got %d, want %d", name, got, want)
+			}
+			if got != gotN {
+				t.Errorf("%s vs SyscallN: %d != %d", name, got, gotN)
+			}
+		}
+
+		{
+			got, _, _ := purego.Syscall1(fn15, 1)
+			gotN, _, _ := purego.SyscallN(fn15, 1)
+			check("Syscall1", got, gotN, 1)
+		}
+		{
+			got, _, _ := purego.Syscall2(fn15, 1, 2)
+			gotN, _, _ := purego.SyscallN(fn15, 1, 2)
+			check("Syscall2", got, gotN, 3)
+		}
+		{
+			got, _, _ := purego.Syscall3(fn15, 1, 2, 3)
+			gotN, _, _ := purego.SyscallN(fn15, 1, 2, 3)
+			check("Syscall3", got, gotN, 6)
+		}
+		{
+			got, _, _ := purego.Syscall4(fn15, 1, 2, 3, 4)
+			gotN, _, _ := purego.SyscallN(fn15, 1, 2, 3, 4)
+			check("Syscall4", got, gotN, 10)
+		}
+		{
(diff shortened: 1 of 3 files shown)


qmuntal:
Why don't you remove `//go:uintptrescapes` from `SyscallN` instead, and add `//go:nosplit` + `//go:uintptrkeepalive`. This is how `syscall.SyscallN` is implemented on Windows, and it's alloc-free: https://github.com/golang/go/blob/17bd5ab8c650155dd2bd09f7005726552639eea0/src/syscall/dll_windows.go#L98.

RomainMuller:
> it's alloc-free

Is it? I've not found a way to get the compiler to not always heap-escape the variadic slice for a cross-module call?

Specifically, my benchmark _always_ shows 1 alloc that matches the vararg slice size... Only because it's cross-module.

eliottness (approved):
On our side this is the last allocation we have in the hot path for our FFI calls. Would be awesome to get alloc parity with CGO there

hajimehoshi:
> Is it? I've not found a way to get the compiler to not always heap-escape the variadic slice for a cross-module call?

So didn't //go:nosplit + //go:uintptrkeepalive work?

TotallyGamerJet:
> Why don't you remove `//go:uintptrescapes` from `SyscallN` instead, and add `//go:nosplit` + `//go:uintptrkeepalive`. This is how `syscall.SyscallN` is implemented on Windows, and it's alloc-free: https://github.com/golang/go/blob/17bd5ab8c650155dd2bd09f7005726552639eea0/src/syscall/dll_windows.go#L98.
> 
> 

The syscalln that it calls does have the noescape pragma which I think may be the reason?

RomainMuller:
The compiler rejects the use of `//go:uintptrkeepalive` (not sure why it works on Windows); claiming it's only supported in the standard library.

```
../../ebitengine/purego/syscall.go:62:3: //go:uintptrkeepalive is only allowed in the standard library
```

If I ignore that (i.e, just don't put `//go:uintptrkeepalive` at all), I am indeed no longer seeing the allocation. But am I changing semantics in a way that could be possibly breaking?

eliottness:
@TotallyGamerJet Any news on this one? Would be awesome to get to allocation parity with cgo here

TotallyGamerJet:
I think we were waiting on @RomainMuller to update the PR to avoid the allocations with a variadic SyscallN

RomainMuller:
@TotallyGamerJet sure... but then I still need to know if it's fine to _entirely drop_ the `uintptr` escaping semantics, since the `go:uintptrkeepalive` pragma is only available in the standard library...

tmc:
Re: whether it's safe to drop the escaping semantics — no, and the alloc-free variadic path isn't reachable from a third-party module anyway.

The gate is `noder.go`: `if flag == ir.UintptrKeepAlive && !base.Flag.Std`. Exact `==`, and `base.Flag.Std` is only set for `std`, so `syscall.dll_windows` gets the bare pragma and purego can't. `//go:uintptrescapes` maps to `UintptrEscapes | UintptrKeepAlive` (`lex.go`), so `!= UintptrKeepAlive`, so it slips the gate and *implies* keepalive — legal for us, bare keepalive isn't.

Dropping `uintptrescapes` without keepalive reopens a UAF: a `uintptr` that's really a stack pointer has nothing pinning it across the call, so stack growth/GC can move or reclaim it while C holds the raw address. That's the exact thing the doc comment above `SyscallN` says the pragma prevents.

Verified by compile, not just reading source — throwaway 3rd-party module:

| | go1.26.3 | go1.27-devel (tip) |
|---|---|---|
| bare `uintptrkeepalive` + `nosplit` | ❌ `only allowed in the standard library` | ❌ same |
| `uintptrescapes` | ✅ | ✅ |

Gate unchanged on tip, no open proposal to move it. So it's pick-two: variadic + `uintptrescapes` (1 alloc, ships today), variadic + keepalive (alloc-free, stdlib-only, won't compile), or these fixed-arity wrappers (alloc-free, compiles). The alloc-free variadic ask lands in the one cell that doesn't build outside `std`. Fixed-arity is the only alloc-free option third parties actually get. Happy to help draft a Go proposal to open the pragma for FFI if there's appetite but for better or worse the fixed arity seems to be the only zero-alloc path here.

PR #456, opened 2026-05-27
internal/fakecgo: remove runtime.load_g references

gdams:
## Summary

Go 1.27 ([`golang/go@aee6009`](https://github.com/golang/go/commit/aee6009ba5e1d71948b03ac0458fbc99e3a14ace)) enforces stricter `checklinkname` for assembly symbols. `runtime.load_g` is now `//go:linknamestd` (std-only), causing linker errors for non-std packages that reference it:

```
link: main: invalid reference to runtime.load_g
```

## Changes

**crosscall2 (6 architectures: arm, arm64, loong64, ppc64le, riscv64, s390x):**
Remove the `load_g` call — `cgocallback` already calls `load_g` internally, so the explicit call is unnecessary. The amd64 and 386 variants already did not call it.

**setg_trampoline (ppc64le):**
Replace the `load_g` call with a direct assignment to the `g` register from the saved `newg` argument, since `setg_gcc` has already stored it into TLS.

## Status

This resolves the `runtime.load_g` linker error. The `runtime.cgocallback` reference remains and will need an upstream Go change or the `-ldflags='-checklinkname=0'` workaround until resolved upstream.

### DIFF
--- a/internal/fakecgo/asm_arm.s
+++ b/internal/fakecgo/asm_arm.s
@@ -32,8 +32,6 @@ TEXT crosscall2(SB), NOSPLIT|NOFRAME, $0
 	MOVD F14, (13*4+8*7)(R13)
 	MOVD F15, (13*4+8*8)(R13)
 
-	BL runtime·load_g(SB)
-
 	// We set up the arguments to cgocallback when saving registers above.
 	BL runtime·cgocallback(SB)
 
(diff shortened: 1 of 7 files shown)


TotallyGamerJet (approved):
LGTM

hajimehoshi (approved):
LGTM

Issue #451, opened 2026-05-21
func.go: Concurrent calls swap return-value pointers (sync.Pool of *syscall15Args, regressed in #282)


nsaini-figma:
### PureGo Version

v0.9.0 and v0.9.1 (both confirmed). Likely all releases from v0.8.1 onward (the first release including #282).

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go 1.24 and go 1.26.1

### What steps will reproduce the problem?

Two Go goroutines calling the same `RegisterLibFunc`-registered function from a shared library, signature `func(uint64) *byte`. Within seconds the goroutines observe each other's return pointers.

**`ultra.c`** — 5 lines of payload, returns a malloc'd buffer with the seed embedded:

```c
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>

char *ultra_alloc(uint64_t seed) {
    char *buf = (char *)malloc(256);
    if (!buf) return NULL;
    snprintf(buf, 256, "{\"seed\":%lu,\"_\":\"%064lu\"}",
             (unsigned long)seed, (unsigned long)seed);
    return buf;
}

void ultra_free(char *s) { free(s); }
```

```sh
cc -O2 -shared -fPIC -o libultra.so ultra.c
```

**`ultra.go`** — N goroutines, each with a per-worker disjoint seed range, reads the JSON back and checks whether the embedded seed matches what it sent:

```go
package main

import (
    "flag"
    "fmt"
    "os"
    "strings"
    "sync"
    "sync/atomic"
    "time"
    "unsafe"

    "github.com/ebitengine/purego"
)

func main() {
    workers := flag.Int("workers", 32, "")
    flag.Parse()

    h, err := purego.Dlopen("./libultra.so", purego.RTLD_NOW|purego.RTLD_GLOBAL)
    if err != nil {
        fmt.Println(err)
        os.Exit(2)
    }

    var alloc func(uint64) *byte
    var free func(*byte)
    purego.RegisterLibFunc(&alloc, h, "ultra_alloc")
    purego.RegisterLibFunc(&free, h, "ultra_free")

    var ops, mismatches atomic.Uint64
    var wg sync.WaitGroup
    done := make(chan struct{})

    for i := 0; i < *workers; i++ {
        wg.Add(1)
        go func(id int) {
            defer wg.Done()
[...]

PR #457, opened 2026-05-27
func.go: restore sync.Pool

AlexanderYastrebov:

Updates #451

bug

## What this PR does | solves

Revert the sync.Pool removal #452 and copy syscall.a1 into a local variable before taking its address, so the reflect.Value points to an independent location that is not affected by pool recycling.

### DIFF
--- a/func.go
+++ b/func.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"reflect"
 	"runtime"
+	"sync"
 	"unsafe"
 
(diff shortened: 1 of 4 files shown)


hajimehoshi:
@TotallyGamerJet Thoughts?

TotallyGamerJet:
I think this looks good. I want to test some things with it before approving

TotallyGamerJet (approved):
LGTM

hajimehoshi (approved):

Issue #236, opened 2024-04-21
Struct support on Linux


ring-c:
### Operating System

- [ ] Windows
- [ ] macOS
- [X] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

### What feature would you like to be added?

[this](https://github.com/ebitengine/purego/blob/main/func.go#L169)

Can we please have support of sending structs as arguments on linux? 
I am willing to test code if required.

JupiterRider:
Maybe it is possible to write a wrapper using libffi. 
libffi.so is available on most linux systems, because it is a dependency of python and ruby.

TotallyGamerJet:
My biggest concern is maintaining the code for each platform. Structs support on Darwin is more complicated than handling all other return and argument types for all platforms combined. Even if most of the code is the same I do not have easy access to these platforms to test and validate the accuracy of the implementations. Just see the stalling of #234

The current goal right now is supporting Ebitengine and as this is not needed right now for it this feature will have to wait. Of course this is not a complete write off and it may happen in the future.

JupiterRider:
I finally have a working example:

```
bar/
├── ffi
│   └── ffi.go
├── go.mod
├── go.sum
└── main.go
```

`ffi/ffi.go`
```golang
//go:build linux && amd64
// +build linux,amd64

package ffi

import (
	"unsafe"

	"github.com/ebitengine/purego"
)

func init() {
	handle, err := purego.Dlopen("libffi.so", purego.RTLD_LAZY)
	if err != nil {
		panic(err)
	}
	purego.RegisterLibFunc(&PrepCif, handle, "ffi_prep_cif")
	purego.RegisterLibFunc(&Call, handle, "ffi_call")
}

type Abi uint32

const (
	DefaultAbi Abi = 2
)

type Status uint32

const (
	OK Status = 0
)

const (
	Void    = 0
	Double  = 3
	Float   = 2
	Sint32  = 10
	Pointer = 14
)

type Typ struct {
	Size      uint64
	Alignment uint16
	Typ       uint16
	Elements  **Typ
}

var (
	TypDouble  = Typ{8, 8, Double, nil}
	TypVoid    = Typ{1, 1, Void, nil}
	TypPointer = Typ{8, 8, Pointer, nil}
	TypSint32  = Typ{4, 4, Sint32, nil}
	TypFloat   = Typ{4, 4, Float, nil}
)

type Cif struct {
	Abi      uint32
	Nargs    uint32
[...]

PR #467, opened 2026-06-20
purego: support structs on Android and iOS

### LINKED ISSUE #466: Struct support on Android and iOS
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [x] Android
- [x] iOS

Support passing and returning C structs by value — in both `RegisterFunc`/`RegisterLibFunc` (Go calling C) and `NewCallback` callbacks (C calling Go) — on Android and iOS for `amd64` and `arm64`, matching the struct support already available on macOS, Linux, and Windows (#236, #237).

### Why is this needed?

Struct-by-value already works on darwin, linux, and windows (amd64/arm64), but `RegisterFunc` currently panics on iOS and Android: `ensureStructSupported` only permits `GOOS` darwin/linux/windows, so `GOOS=ios` and `GOOS=android` fall through and are rejected — even though their ABIs are identical to platforms that already work (iOS arm64 matches macOS arm64; the iOS simulator and Android amd64 are System V; Android arm64 is standard AAPCS64). The struct-marshalling code is already compiled into the iOS/Android builds and is only gated off by the `runtime.GOOS` string checks.

iOS and Android are Tier 1 platforms, and many C APIs — Apple frameworks and Android NDK libraries — pass and return small structs by value, so this gap blocks real-world FFI on mobile.

---

_Filed by Claude (Claude Code) on behalf of @hajimehoshi. An implementation is being prepared._

hajimehoshi:

Closes #466

feature

## What this PR does | solves

Enables passing and returning C structs by value on Android and iOS for
amd64 and arm64, in both `RegisterFunc`/`RegisterLibFunc` calls and `NewCallback`
callbacks — matching the support already available on macOS, Linux, and Windows
(#236, #237).

### Why it works

The struct-marshalling code was already compiled into the iOS/Android builds and
only gated off by `runtime.GOOS` string checks. Their ABIs are identical to
platforms that already work:

- iOS arm64 == macOS arm64 (Apple ABI); the iOS simulator (amd64) is System V.
- Android amd64 == System V; Android arm64 == standard AAPCS64.

So this routes `GOOS=ios` through the Darwin path via a new `isDarwin` helper,
treats `GOOS=android` like Linux, opens the `ensureStructSupported` gate to
ios/android, adds them to the `getCallbackStruct` callback switches, and extends
the Darwin ARM64 stack-argument reads in the callback path to iOS.

### Tests / CI

Adds an Android emulator CI job (amd64 and arm64) that runs the test suite.
Because the emulator has no C toolchain, the test libraries are cross-compiled on
the host with the NDK and loaded via a new `PUREGO_TEST_PREBUILT_LIBDIR` env var
instead of being compiled at run time. `getSystemLibrary` also learns Android's
libc (and iOS, for a future job).

### Reviewer notes

- I could not run the Android emulator job locally (no emulator/NDK in my
  environment), so it may need a CI iteration to settle image/arch specifics —
  notably the arm64-v8a image under full emulation on the x86_64 runner, and
  `dlopen("libc.so")` for the adb-shell user. The harness change itself is
  verified on the host: the full suite passes through `PUREGO_TEST_PREBUILT_LIBDIR`.
- An iOS Simulator job is left as a TODO in the workflow (lower marginal value,
  since the iOS arm64 ABI matches macOS arm64, which the macOS runner already
  exercises).

---

_Authored by Claude (Claude Code) on behalf of @hajimehoshi._

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -316,3 +316,77 @@ jobs:
             rm -fr /usr/local/go && tar -C /usr/local -xf go${{matrix.go}}.netbsd-amd64.tar.gz
             chmod +x $GITHUB_WORKSPACE/.github/scripts/bsd_tests.sh
           run: $GITHUB_WORKSPACE/.github/scripts/bsd_tests.sh
+
+  # Android is tested on amd64 only. The emulator requires the system image
+  # architecture to match the host, and GitHub's arm64 hosted runners do not
+  # provide a usable KVM device, so there is no hosted hardware-accelerated path
+  # to an Android arm64 emulator. The arm64 struct ABI is already covered by the
+  # linux/arm64 `test` job, and Android's Bionic/cgo integration by this job.
+  android:
+    strategy:
+      fail-fast: false
+      matrix:
+        go: ['1.25.x', '1.26.x']
+    name: Test with Go ${{ matrix.go }} on Android amd64
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: ${{ matrix.go }}
+
+      # The tests compile their C libraries at run time via `go env CC`, but the
+      # emulator has no C toolchain. Every library the tests load must therefore be
+      # cross-compiled here and loaded as a prebuilt library at run time instead.
+      - name: Cross-compile the test binary and C libraries
+        run: |
+          set -eux
+          NDK="$ANDROID_NDK_LATEST_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin"
+          CC="$NDK/x86_64-linux-android21-clang"
+          mkdir -p prebuilt
+          # Build every library the tests load. Keep this in sync with the
+          # buildSharedLib callers in the tests; a missing entry surfaces as a
+          # "prebuilt lib ... no such file" failure on the emulator.
+          "$CC" -shared -Wall -Werror -fPIC -o prebuilt/structtest.so testdata/structtest/struct_test.c
+          "$CC" -shared -Wall -Werror -fPIC -o prebuilt/structreturntest.so testdata/structtest/structreturn_test.c
+          "$CC" -shared -Wall -Werror -fPIC -o prebuilt/libcbtest.so testdata/libcbtest/callback_test.c
+          "$CC" -shared -Wall -Werror -fPIC -o prebuilt/abitest.so testdata/abitest/abi_test.c
+          env GOOS=android GOARCH=amd64 CGO_ENABLED=1 CC="$CC" go test -c -o purego.test .
+
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
+      - name: Run the test suite on the Android emulator
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 34
+          arch: x86_64
+          target: google_apis
+          emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim
+          script: |
+            set -eux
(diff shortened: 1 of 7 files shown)


struct_arm64.go
```diff
@@ -301,7 +301,7 @@ func isHVA(t reflect.Type) bool {
 // copyStruct8ByteChunks copies struct memory in 8-byte chunks to the provided callback.
 // This is used for Darwin ARM64's byte-level packing of non-HFA/HVA structs.
 func copyStruct8ByteChunks(ptr unsafe.Pointer, size uintptr, addChunk func(uintptr)) {
-	if runtime.GOOS != "darwin" {
+	if !isDarwin() {
 		panic("purego: should only be called on darwin")
 	}
```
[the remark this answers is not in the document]
hajimehoshi: I don't think this needs to be fixed.

struct_arm64.go
```diff
@@ -328,7 +328,7 @@ func copyStruct8ByteChunks(ptr unsafe.Pointer, size uintptr, addChunk func(uintp
 // For non-HFA/HVA structs, Darwin uses byte-level packing. We copy the struct memory in
 // 8-byte chunks, which works correctly for both register and stack placement.
 func placeRegistersDarwin(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
-	if runtime.GOOS != "darwin" {
+	if !isDarwin() {
 		panic("purego: placeRegistersDarwin should only be called on darwin")
 	}
```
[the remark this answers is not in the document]
hajimehoshi: I don't think this needs to be fixed.

hajimehoshi:
@TotallyGamerJet Please take a look, thanks

.github/workflows/test.yml
```diff
@@ -316,3 +316,79 @@ jobs:
...
+  # provide a usable KVM device, so there is no hosted hardware-accelerated path
+  # to an Android arm64 emulator. The arm64 struct ABI is already covered by the
+  # linux/arm64 `test` job, and Android's Bionic/cgo integration by this job.
+  android:
+    strategy:
+      fail-fast: false
+      matrix:
+        # Test only the latest two Go versions to save CI time.
+        # See https://go.dev/doc/devel/release#policy
+        go: ['1.25.x', '1.26.x']
```
TotallyGamerJet, line 329: If we're dropping support for the older versions I don't think the comment is necessary

README.md
```diff
@@ -30,27 +30,27 @@ except for float arguments and return values.
...
+- **Linux**: 386<sup>3</sup>, arm<sup>3</sup>, loong64<sup>3</sup>, ppc64le<sup>3</sup>, riscv64<sup>3</sup>, s390x<sup>1,3</sup>
+- **NetBSD**: amd64<sup>3,4</sup>, arm64<sup>3,4</sup>
+- **Windows**: 386<sup>3,5</sup>, arm<sup>3,5,6</sup>
 
 #### Support Notes
 
 1. These architectures require CGO_ENABLED=1 to compile
-2. These architectures do not support passing structs by value as arguments or return values
-3. These architectures support passing structs by value as arguments and return values when calling C functions, but not in callbacks created with `NewCallback`
+2. These architectures support passing structs by value as arguments and return values when calling C functions, but not in callbacks created with `NewCallback`
```
TotallyGamerJet, line 52: Why swap these?

func.go
```diff
@@ -505,11 +505,19 @@ func ensureStructSupported() {
...
+	case "android", "darwin", "ios", "linux", "windows":
+	default:
+		panic("purego: struct arguments/returns are only supported on android, darwin, ios, linux, and windows")
 	}
 }
 
+// isDarwin reports whether the platform uses Apple's calling convention.
+// iOS (GOOS=ios) shares it with macOS (GOOS=darwin).
+func isDarwin() bool {
+	return runtime.GOOS == "darwin" || runtime.GOOS == "ios"
```
TotallyGamerJet: Make this a constant instead of a function

README.md
hajimehoshi: This change sorts the indices in the order of their appearances.

.github/workflows/test.yml
hajimehoshi: Done

func.go
hajimehoshi: Done

TotallyGamerJet (approved):
LGTM

PR #431, opened 2026-03-05
purego: Support 32 arguments on 64bit

tmc:
# What issue is this addressing?
Closes #425 

## What _type_ of issue is this addressing?
feature

## What this PR does | solves
This increases the # of supported arguments to 32.

PR #473, opened 2026-07-05
purego: support structs on loong64

hajimehoshi:

Updates #474

feature

## What this PR does | solves

Enables passing and returning C structs by value through `RegisterFunc` /
`RegisterLibFunc` on **linux/loong64**.

The per-architecture code added in #431 was unreachable behind a guard that
only permitted amd64 and arm64. This PR makes that guard **direction-aware** and
adds loong64:

- `ensureStructSupported` gates the forward call path (`RegisterFunc`) and now
  allows loong64 in addition to amd64/arm64.
- A new `ensureCallbackStructSupported` keeps the callback path (`NewCallback`)
  restricted to amd64/arm64, where `getCallbackStruct` / `setStruct` are
  implemented. Struct callbacks remain unsupported on loong64.
- Struct return types are now validated when creating a callback, so an
  unsupported signature fails at `NewCallback` time instead of panicking later.

The loong64 `struct_loong64.go` implementation was rewritten to follow the
LoongArch hard-float calling convention: an aggregate uses FP registers only
when, after flattening nested structs/arrays and ignoring blank padding fields,
it has one or two floating-point members and nothing else, or exactly one
floating-point member with one integer member; everything else uses the integer
convention. `float32` values are NaN-boxed in FP registers. This was validated
against the real compiler with `zig cc -target loongarch64-linux-gnu -S` and by
the QEMU test job.

**Scope:** only loong64 is enabled here. ppc64le, riscv64, and s390x still use
the old `isAllSameFloat` heuristic, which mishandles floating-point aggregates
(ppc64le currently crashes under test). Each will be enabled in its own PR once
its calling convention is implemented correctly. `README.md` reflects this:
loong64 moves to "structs when calling C functions, but not in callbacks", while
the other three stay unsupported.

---

Authored by Claude (Claude Code) on behalf of @hajimehoshi.

### DIFF
--- a/struct_loong64.go
+++ b/struct_loong64.go
@@ -22,7 +22,7 @@ type loong64Leaf struct {
 func loong64Flatten(t reflect.Type, base uintptr, leaves *[]loong64Leaf) {
 	switch t.Kind() {
 	case reflect.Struct:
-		for i := 0; i < t.NumField(); i++ {
+		for i := range t.NumField() {
 			f := t.Field(i)
 			if f.Name == "_" {
 				// Blank fields are explicit padding to match the C layout, not
@@ -33,7 +33,7 @@ func loong64Flatten(t reflect.Type, base uintptr, leaves *[]loong64Leaf) {
 		}
 	case reflect.Array:
 		elem := t.Elem()
-		for i := 0; i < t.Len(); i++ {
+		for i := range t.Len() {
 			loong64Flatten(elem, base+uintptr(i)*elem.Size(), leaves)
 		}
 	default:
@@ -79,7 +79,10 @@ func getStruct(outType reflect.Type, syscall syscallArgs) reflect.Value {
 	}
 	if outSize > 16 {
 		// Returned indirectly through a pointer in the first integer register.
-		return reflect.NewAt(outType, *(*unsafe.Pointer)(unsafe.Pointer(&syscall.a1))).Elem()
+		// Read it from a local copy so the address does not alias the pooled
+		// syscallArgs.
+		a1 := syscall.a1
+		return reflect.NewAt(outType, *(*unsafe.Pointer)(unsafe.Pointer(&a1))).Elem()
 	}
 
 	var buf [16]byte
@@ -90,18 +93,18 @@ func getStruct(outType reflect.Type, syscall syscallArgs) reflect.Value {
 		var fi, ii int
 		for _, l := range leaves {
 			dst := unsafe.Add(base, l.offset)
-			if l.isFloat {
-				r := floatRegs[fi]
-				fi++
-				if l.kind == reflect.Float32 {
-					// A single-precision value is NaN-boxed in the register.
-					*(*uint32)(dst) = uint32(r)
-				} else {
-					*(*uint64)(dst) = uint64(r)
-				}
-			} else {
+			if !l.isFloat {
 				loong64StoreInt(dst, l.size, intRegs[ii])
 				ii++
+				continue
+			}
+			r := floatRegs[fi]
+			fi++
+			if l.kind == reflect.Float32 {
+				// A single-precision value is NaN-boxed in the register.
+				*(*uint32)(dst) = uint32(r)
+			} else {
+				*(*uint64)(dst) = uint64(r)
 			}
 		}
 	} else {
--- a/struct_loong64.go
+++ b/struct_loong64.go
@@ -79,10 +79,7 @@ func getStruct(outType reflect.Type, syscall syscallArgs) reflect.Value {
 	}
 	if outSize > 16 {
 		// Returned indirectly through a pointer in the first integer register.
-		// Read it from a local copy so the address does not alias the pooled
-		// syscallArgs.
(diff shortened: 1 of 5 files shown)


hajimehoshi:

struct_loong64.go
```diff
@@ -4,177 +4,193 @@
...
-		return reflect.New(outType).Elem()
-	case outSize <= 8:
-		r1 := syscall.a1
-		if isAllFloats, numFields := isAllSameFloat(outType); isAllFloats {
-			r1 = syscall.f1
-			if numFields == 2 {
-				r1 = syscall.f2<<32 | syscall.f1
+// loongLeaf is a scalar member of an aggregate after flattening nested structs
+// and arrays. offset is the byte offset of the member within the aggregate.
+type loongLeaf struct {
```
TotallyGamerJet: Call this loong64Leaf to avoid people thinking it's a misspelling of longLeaf.

Ditto for the other places loong is used

hajimehoshi: Done

struct_loong64.go
```diff
@@ -4,177 +4,193 @@
...
-				r2 = syscall.f3
-			case 2:
-				r1 = syscall.f1
-				r2 = syscall.f2
-			default:
-				panic("not reached")
-			}
+	case reflect.Array:
+		elem := t.Elem()
+		for i := 0; i < t.Len(); i++ {
```
TotallyGamerJet: Range? Also, might want to run `go fix` for any other changes

struct_loong64.go
```diff
@@ -4,177 +4,193 @@
...
+}
+
+func getStruct(outType reflect.Type, syscall syscallArgs) reflect.Value {
+	outSize := outType.Size()
+	if outSize == 0 {
+		return reflect.New(outType).Elem()
+	}
+	if outSize > 16 {
+		// Returned indirectly through a pointer in the first integer register.
 		return reflect.NewAt(outType, *(*unsafe.Pointer)(unsafe.Pointer(&syscall.a1))).Elem()
```
TotallyGamerJet: I think we should make a local variable instead of directly taking the address. Don't want to cause this issue again: https://github.com/ebitengine/purego/commit/d339eb0fd4d3c57f9b641f4781615650183c7043

struct_loong64.go
```diff
@@ -4,177 +4,193 @@
...
+
+	var buf [16]byte
+	base := unsafe.Pointer(&buf[0])
+	if leaves, useFP := loong64Classify(outType); useFP {
+		floatRegs := [2]uintptr{syscall.f1, syscall.f2}
+		intRegs := [2]uintptr{syscall.a1, syscall.a2}
+		var fi, ii int
+		for _, l := range leaves {
+			dst := unsafe.Add(base, l.offset)
+			if l.isFloat {
```
TotallyGamerJet: Can we invert the if statement to avoid a ton of indents?
```go
if !l.isFloat {
     loong64StoreInt(dst, l.size, intRegs[ii])
    ii++
    continue
}
r := floatRegs[fi]
fi++
if l.kind == reflect.Float32 {
    // A single-precision value is NaN-boxed in the register.
     *(*uint32)(dst) = uint32(r)
} else {
     *(*uint64)(dst) = uint64(r)
}
```

struct_test.go
```diff
@@ -83,6 +83,11 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 			// not support struct arguments or returns.
 			continue
 		}
+		if imp.usesCallbacks && runtime.GOARCH != "amd64" && runtime.GOARCH != "arm64" {
+			// Struct arguments and returns in callbacks are only supported on
```
TotallyGamerJet: Not sure this comment is necessary. Seem obvious from the code what it's checking for

struct_test.go
```diff
@@ -908,6 +913,11 @@ func TestRegisterFunc_structReturns(t *testing.T) {
...
```
TotallyGamerJet: Ditto

struct_loong64.go
hajimehoshi: I run `go fix` but this was not affected. That's mysterious. Done

struct_loong64.go
hajimehoshi: Done

struct_loong64.go
hajimehoshi: Done

struct_test.go
hajimehoshi: Done

struct_test.go
hajimehoshi: Done

struct_loong64.go
hajimehoshi: Hm? syscall.a1 is not pooled now (so my new comment is not correct), but you mean we need a local variable for defensive code?

hajimehoshi: As syscall is passed by value at getStruct, I don't think we need this, so I keep this as it is. Let me know what you think.

TotallyGamerJet (approved):
LGTM

Issue #474, opened 2026-07-05
purego: add structs for more architectures


hajimehoshi:
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] Android
- [ ] iOS

### What feature would you like to be added?

Now structs are supported only for `amd64` and `arm64`. I propose to support more architectures.

Basically there are two types of struct usages: passing/receiving structs via `Register(Lib)Func`, and in callbacks of `NewCallback`. We don't have to support both at the same time, so let's do it one by one.

### Why is this needed?

I found an X API that actually passes a struct value `XSyncValue` (
https://xorg.freedesktop.org/archive/X11R7.7/doc/libXext/synclib.html#c_functions). This is needed for Ebitengine though this is an optional. [Now Ebitengine limits architectures for this API usage](https://github.com/hajimehoshi/ebiten/commit/d8e8661d550674b42555cc557530ddf6fabff232#diff-d7126f0ec82312190c5f63c7297056f13aea2cc1aacbdcd0c52e0cd6990d4d1bR656-R664), but I'd like to support more platforms if possible.

hajimehoshi:
> [Now Ebitengine limits architectures for this API usage](https://github.com/hajimehoshi/ebiten/commit/d8e8661d550674b42555cc557530ddf6fabff232#diff-d7126f0ec82312190c5f63c7297056f13aea2cc1aacbdcd0c52e0cd6990d4d1bR656-R664)

For now, Ebitengine decided to use a 64bit integer instead of a struct (2 32bit integers). We still need to implement features for 32bit architectures

PR #475, opened 2026-07-07
purego: support structs on ppc64le

hajimehoshi:

Updates #474

feature

## What this PR does | solves

Adds struct-by-value support for `linux/ppc64le` when calling C
functions, following the loong64 work in #473.

Under the ELFv2 ABI:

- Homogeneous floating-point aggregates (1–8 members, all the same
  floating-point type) are passed and returned in the floating-point
  registers.
- Other aggregates are passed by value across the general-purpose
  registers and the parameter save area. A floating-point argument
  shadows and skips a GPR, so an HFA argument advances the integer
  registers accordingly.
- A non-HFA aggregate larger than two doublewords is returned through a
  caller-allocated buffer whose pointer is passed and returned in `r3`.

Struct arguments and returns in callbacks (`NewCallback`) remain
unsupported, as on the other non-amd64/arm64 architectures.

### Not yet handled (skipped in the tests on ppc64le)

- Calls needing more than eight floating-point argument registers
  (`LargeFloatStructWithRegs`, `RectangleWithRegs`); purego wires f1–f8.
- A struct argument that follows a scalar floating-point argument
  (`IdentityTwoInt64AfterPrims`): the scalar float shadows a GPR, which
  purego does not model for scalars without also changing the callback
  path.

### Testing

The register/stack placement and return-value reconstruction were
validated against GCC/Clang codegen by disassembling the test C for
ppc64le (ELFv2) with `zig cc -target powerpc64le-linux-gnu`, plus the
full host test suite. The QEMU-based ppc64le CI job exercises the struct
tests.

### DIFF
--- a/func.go
+++ b/func.go
@@ -199,7 +199,7 @@ func RegisterFunc(fptr any, cfn uintptr) {
 			ensureStructSupported()
 			outType := ty.Out(0)
 			checkStructFieldsSupported(outType)
-			if structReturnInMemory(outType.Size()) {
+			if structReturnInMemory(outType) {
 				// A struct returned in memory is allocated by the caller and its
 				// pointer is passed as a hidden first integer argument. When the
 				// integer registers are already full, prepending it spills a
@@ -292,7 +292,7 @@ func RegisterFunc(fptr any, cfn uintptr) {
 		var arm64_r8 uintptr
 		if ty.NumOut() == 1 && ty.Out(0).Kind() == reflect.Struct {
 			outType := ty.Out(0)
-			if structReturnInMemory(outType.Size()) {
+			if structReturnInMemory(outType) {
 				// The caller allocates the return value and passes its pointer
 				// as a hidden first integer argument.
 				val := reflect.New(outType)
@@ -431,10 +431,14 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
 			addInt(0)
 		}
 	case reflect.Float32:
-		// On S390X big-endian, float32 goes in upper 32 bits of 64-bit FP register
-		if runtime.GOARCH == "s390x" {
+		switch runtime.GOARCH {
+		case "ppc64le":
+			// A single-precision argument occupies a floating-point register in double format on Power.
+			addFloat(uintptr(math.Float64bits(v.Float())))
+		case "s390x":
+			// S390X big-endian: float32 goes in the upper 32 bits of the 64-bit FP register.
 			addFloat(uintptr(math.Float32bits(float32(v.Float()))) << 32)
-		} else {
+		default:
 			addFloat(uintptr(math.Float32bits(float32(v.Float()))))
 		}
 	case reflect.Float64:
@@ -510,9 +514,9 @@ func checkStructFieldsSupported(ty reflect.Type) {
 // a C function is unsupported on the current platform.
 func ensureStructSupported() {
 	switch runtime.GOARCH {
-	case "amd64", "arm64", "loong64":
+	case "amd64", "arm64", "loong64", "ppc64le":
 	default:
-		panic("purego: struct arguments/returns are only supported on amd64, arm64, and loong64")
+		panic("purego: struct arguments/returns are only supported on amd64, arm64, loong64, and ppc64le")
 	}
 	switch runtime.GOOS {
 	case "android", "darwin", "ios", "linux", "windows":
--- a/struct_ppc64le.go
+++ b/struct_ppc64le.go
@@ -4,84 +4,76 @@
 package purego
 
 import (
+	"math"
 	"reflect"
 	"unsafe"
 )
 
-// structReturnInMemory reports whether a struct return value of the given size
-// is returned through a caller-allocated hidden pointer passed as the first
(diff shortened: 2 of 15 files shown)


struct_ppc64le.go
```diff
@@ -86,58 +73,105 @@ func placeSmallAggregatePPC64LE(
...
+	// An HFA passes each member in its own floating-point register.
+	if leaves, isHFA := ppc64leClassifyHFA(v.Type()); isHFA {
+		for _, l := range leaves {
+			src := unsafe.Add(ptr, l.offset)
+			if l.kind == reflect.Float32 {
+				// Single precision occupies the register in double format.
+				addFloat(uintptr(math.Float64bits(float64(*(*float32)(src)))))
+			} else {
+				addFloat(uintptr(*(*uint64)(src)))
+			}
```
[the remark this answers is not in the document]
hajimehoshi: This is intentional and correct for ppc64le. Power FP registers hold values in double-precision format — the call path loads argument FPRs with FMOVD (lfd), and disassembling the test C with zig cc -target powerpc64le-linux-gnu shows the callee accesses HFA float members with lfs/stfs, i.e. it expects the single value stored in double format. Raw 32-bit bits would make the callee read a denormal instead.

The real inconsistency was the scalar float32 path in addValue, which used raw single-bits — a genuine bug that made the struct-return tests fail. The follow-up makes scalar float32 args, HFA float32 members, and float32 callback reads all use double format consistently.

Reply written by Claude (Claude Code) on behalf of @hajimehoshi.

struct_ppc64le.go
```diff
@@ -86,58 +73,105 @@ func placeSmallAggregatePPC64LE(
...
 
-	if isFloats, _ := isAllSameFloat(v.Type()); isFloats {
-		addFloat(w0)
-		if size > 8 {
-			addFloat(w1)
+	// An HFA returns each member in its own floating-point register.
+	if isFloats, numFields := isAllSameFloat(outType); isFloats && numFields <= 8 {
+		floatRegs := [8]uintptr{
+			syscall.f1, syscall.f2, syscall.f3, syscall.f4,
+			syscall.f5, syscall.f6, syscall.f7, syscall.f8,
```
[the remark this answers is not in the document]
hajimehoshi: Fixed. getStruct now classifies with ppc64leClassifyHFA (the same array/blank-aware flattening used for arguments) and writes each member at its actual leaf offset instead of packing contiguously, so nested structs, arrays, and blank padding round-trip. float32 members are decoded via Float64frombits because the register holds the value in double format (see the other thread).

Reply written by Claude (Claude Code) on behalf of @hajimehoshi.

func.go
```diff
@@ -287,11 +287,20 @@ func RegisterFunc(fptr any, cfn uintptr) {
...
+			} else if runtime.GOARCH == "ppc64le" && outType.Size() > maxRegAllocStructSize {
+				// ppc64le returns an HFA of up to eight members in the
+				// floating-point registers; only other aggregates need a buffer.
+				isAllFloats, numFields := isAllSameFloat(outType)
+				if !isAllFloats || numFields > 8 {
+					val := reflect.New(outType)
+					keepAlive = append(keepAlive, val)
+					addInt(val.Pointer())
+				}
 			} else if runtime.GOARCH == "arm64" && outType.Size() > maxRegAllocStructSize {
```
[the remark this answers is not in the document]
hajimehoshi: Fixed. ppc64leClassifyHFA now lives in func.go, so the hidden-buffer decision, argument passing, and getStruct share one classifier. HFAs defined via arrays or with blank padding are classified consistently, so no unexpected hidden pointer is passed.

Reply written by Claude (Claude Code) on behalf of @hajimehoshi.

hajimehoshi:
@TotallyGamerJet PTAL

func.go
```diff
@@ -480,6 +492,54 @@ func isAllSameFloat(ty reflect.Type) (allFloats bool, numFields int) {
 	return allFloats, numFields
 }
 
+// ppc64leLeaf is a scalar member of a flattened aggregate at byte offset offset.
```
TotallyGamerJet: Shouldn't this be in struct_ppc64?

struct_notppc64le.go
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
+
+//go:build !ppc64le
+
+package purego
+
+import "reflect"
+
+func ppc64leStructReturnInMemory(reflect.Type) bool {
```
hajimehoshi: Unfortunately we need a stub file. Another solution is to define stub functions for each `struct_*.go` files. Thoughts? @TotallyGamerJet

TotallyGamerJet: At this point it might make sense to make a `structReturnInMemory` for each platform inside the struct_GOARCH.go instead of doing the checks on runtime.GOARCH

hajimehoshi: I see. I think I'll make a separate PR for this refactoring (e.g. amd64StructReturnInMemory) before this PR.

hajimehoshi: Updated

TotallyGamerJet (approved):
LGTM

PR #478, opened 2026-07-14
purego: support RegisterFunc and typed callbacks on Windows 386

nekohasekai:

There is no existing issue. This expands the Windows/386 support currently described in the platform support table.

feature

## What this PR does | solves

Windows/386 previously only officially supported `SyscallN` and `NewCallback`. The partial `RegisterFunc` path also treated 64-bit values as a single `uintptr`, only observed `EAX` for integer results, and did not retrieve floating-point results from x87.

This change adds the Windows/386 ABI handling needed for typed `RegisterFunc`/`RegisterLibFunc` calls and typed callbacks:

- lays out integer, pointer, bool, string, `float32`, and `float64` arguments on the 32-bit stack;
- splits 64-bit arguments into little-endian low/high slots and rebuilds integer results from `EDX:EAX`;
- saves direct floating-point results from x87 `ST(0)` and restores callback floating-point results there;
- decodes 64-bit and floating-point callback arguments, supports 64-bit callback results, and preserves both stdcall and cdecl stack cleanup;
- passes up to the declared 32 stack slots instead of the old 15-slot fallback;
- supports by-value struct arguments and returns for direct calls, matching the existing Windows limitation that structs are not supported in callbacks; and
- adds Windows/386 ABI, x87 edge-case, stack-limit, concurrency, reentrancy, native-thread callback, and callback-table coverage.

The Windows CI jobs now use an explicit MSYS2 `MINGW32` GNU toolchain, verify that it can build a 32-bit DLL, and run both Cgo and non-Cgo Windows/386 suites (including unoptimized builds) on Go 1.25, 1.26, and 1.27rc1.

Validation:

- `go test ./...`
- Windows/386 tests in a real 32-bit Windows process with LLVM-MinGW
- full GitHub Actions matrix, including all three `windows-latest` jobs: https://github.com/SagerNet/purego/actions/runs/29236606685


.github/workflows/test.yml
```diff
@@ -124,14 +136,20 @@ jobs:
...
 
       - name: go test (Windows 386)
-        if: runner.os == 'Windows'
+        if: runner.os == 'Windows' && runner.arch != 'ARM64'
+        shell: msys2 {0}
         run: |
-          env CGO_ENABLED=0 GOARCH=386 go test -shuffle=on -v -count=10 ./...
+          CC="$(cygpath -w "$(command -v gcc)")"
+          env CGO_ENABLED=0 GOARCH=386 CC="$CC" go test -shuffle=on -v -count=10 ./...
+          env CGO_ENABLED=0 GOARCH=386 CC="$CC" go test "-gcflags=all=-N -l" -v ./...
```
hajimehoshi, line 144: I was wondering why CC is needed even when CGO_ENABLED=0

nekohasekai: The tests invoke CC directly to build 32-bit DLL fixtures, even with `CGO_ENABLED=0`.

TotallyGamerJet:
Can you create an issue to track this change?

hajimehoshi:
Update README.md too

nekohasekai:
The README is already updated in this PR — the platform table now lists Windows 386 with footnote 2 (structs supported as arguments/return values in calls, but not in callbacks), replacing the previous footnotes 3 and 6. Let me know if you'd like it documented differently.

hajimehoshi:
Ah sorry, I missed that

callback_windows_386.go
```diff
@@ -0,0 +1,179 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
+
+//go:build windows && 386
```
hajimehoshi: As the filename already specified GOOS and GOARCH, we don't need this

struct_windows_386.go
```diff
@@ -0,0 +1,107 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
+
+//go:build windows && 386
```
hajimehoshi: As the filename already specifies GOOS and GOARCH, we don't need this

sys_windows_386.s
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
+
+//go:build windows && 386
```
hajimehoshi: As the filename already specifies GOOS and GOARCH, we don't need this

.github/workflows/test.yml
```diff
@@ -34,6 +34,18 @@ jobs:
       - name: Set up Msys2
         if: runner.os == 'Windows' && runner.arch != 'ARM64'
         uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW32
```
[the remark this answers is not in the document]
nekohasekai: The MINGW32 gcc is only visible to the two Windows/386 steps that run with `shell: msys2 {0}`. The native test steps use `shell: bash` (Git Bash), whose PATH still resolves to the runner's preinstalled x86_64-w64-mingw32 gcc — `path-type: inherit` only makes the outer PATH visible *inside* the MSYS2 shell, not the other way around. The passing windows-latest amd64 Cgo jobs confirm this: they build and load 64-bit test DLLs successfully.

func.go
```diff
@@ -290,8 +303,21 @@ func RegisterFunc(fptr any, cfn uintptr) {
 		}()
 
 		var arm64_r8 uintptr
+		if runtime.GOARCH == "386" && ty.NumOut() == 1 &&
+			(ty.Out(0).Kind() == reflect.Float32 || ty.Out(0).Kind() == reflect.Float64) {
+			// The 386 bridge uses this to avoid popping x87 ST(0) after
+			// non-floating calls.
+			arm64_r8 = 1
+		}
```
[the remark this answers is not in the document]
nekohasekai: Done — renamed the local to the architecture-neutral `aux` and added a comment describing its meaning per architecture (arm64: hidden struct-return pointer in R8; 386: x87 float-return mode). The per-arch `syscallArgs` field names (`arm64_r8` / `floatReturn`) are unchanged since those files are already architecture-specific.

hajimehoshi:
> This branch has conflicts that must be resolved

This needs to be fixed

Issue #482, opened 2026-07-18
0.10 branch: macOS CI fails on macos-latest (macOS 26) with Go 1.20-1.23: missing LC_UUID load command


hajimehoshi:
### PureGo Version

`0.10` branch (v0.10.0)

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

Go 1.20.x through 1.23.x (reproduced with go1.20.14, go1.22.12 and go1.23.11 on darwin/arm64)

### What steps will reproduce the problem?

CI on the `0.10` branch fails now that GitHub's `macos-latest` label has migrated to macOS 26:
https://github.com/ebitengine/purego/actions/runs/29646451569

The `Test with Go 1.22.x on macos-latest` job fails in the `go test` step. Every other error in that run is `The operation was canceled.` from fail-fast.

Locally, on macOS 26.5 (arm64):

```
git checkout 0.10
env CGO_ENABLED=1 go1.22.12 test -c -o purego.test .
./purego.test -test.run TestOS
```

### What is the expected result?

The test binary runs, as it did while `macos-latest` was macOS 15. The same job passed on
https://github.com/ebitengine/purego/actions/runs/26362358247 (2026-05-24), where the runner image was `macos-15-arm64` / macOS 15.7.7.

### What happens instead?

The binary is rejected by dyld before `main` runs:

```
dyld[9810]: missing LC_UUID load command in /var/folders/.../b001/purego.test
dyld[9810]: missing LC_UUID load command
signal: abort trap
FAIL	github.com/ebitengine/purego	0.003s
```

### Anything else you feel useful to add?

**Cause**

This is not a purego bug. On macOS 26, dyld refuses to exec a Mach-O binary that has no `LC_UUID` load command when the binary targets the macOS 26 SDK and links CoreFoundation. Go's internal linker only emits `LC_UUID` when `-B` is passed, and `-B gobuildid` became the default only in Go 1.24 (`cmd/link/internal/ld/main.go`: `if *flagHostBuildid == "" && *flagBuildid != "" { *flagHostBuildid = "gobuildid" }`, present in go1.24.0, absent in go1.23.12).

CoreFoundation is pulled in by the cgo-enabled `net` package, which is why `CGO_ENABLED=0` passes and `CGO_ENABLED=1` fails within the same step. The `-race` steps fail for the same reason.

[...]

PR #469, opened 2026-06-20
all: drop support for Go 1.18 through 1.24

hajimehoshi:
# What issue is this addressing?

Updates #468

## What _type_ of issue is this addressing?

feature

## What this PR does | solves

Drops support for Go 1.18 through 1.24, raising the minimum to Go 1.25.

The CI matrix exercised every Go release back to 1.18, which makes test
runs long. That cost only grows as more targets are added (e.g. a future
`windows/arm64`). Narrowing the supported range also lets us drop
version-gated workarounds, reducing maintenance burden.

Changes:

- **go.mod**: language version `1.18` → `1.25.0`.
- **CI** (`.github/workflows/test.yml`): trim the matrix to the two most
  recent Go releases (`1.25.x`, `1.26.x`), and drop the now-always-true
  `!startsWith(matrix.go, '1.18'/'1.19')` guards on the race steps.
- Remove compatibility shims that only existed for older toolchains:
  - `internal/xreflect` (polyfilled `reflect.TypeAssert`) — callers now
    use the stdlib `reflect.TypeAssert` directly.
  - the `objc` `clone` helper — replaced by `slices.Clone`.
- Modernize the remaining code to the new baseline: integer `range`
  loops, `reflect.TypeFor`, and an `iter.Seq`-based field iterator in
  the struct tests.

Verified locally: `go vet`, `gofmt -s`, `go build` (native +
cross-compile to linux/windows/freebsd), and the full test suite
(`CGO_ENABLED=1`) all pass.

---
*Authored by Claude (Claude Code), on behalf of @hajimehoshi.*

PR #483, opened 2026-07-18
.github/workflows: use macos-15 instead of macos-latest

hajimehoshi:

Updates #482

bug

## What this PR does | solves

CI on this branch is red because GitHub's `macos-latest` label has migrated to macOS 26. On macOS 26, dyld refuses to exec a Mach-O binary that has no `LC_UUID` load command, so the test binaries abort before `main` runs:

```
dyld[9810]: missing LC_UUID load command in .../b001/purego.test
signal: abort trap
FAIL	github.com/ebitengine/purego	0.003s
```

The Go linker emits `LC_UUID` only when `-B` is passed, and `-B gobuildid` became the default in Go 1.24 (`cmd/link/internal/ld/main.go`). Go 1.20 through 1.23 therefore produce binaries that macOS 26 will not start. CoreFoundation, pulled in by the cgo-enabled `net` package, is what triggers dyld's check, which is why `CGO_ENABLED=0` passes and `CGO_ENABLED=1` fails within the same step.

Measured on macOS 26.5 with this branch's tree:

| Go | `LC_UUID` | `LC_BUILD_VERSION` | `CGO_ENABLED=1` test binary |
|---|---|---|---|
| 1.18.10 | absent | minos 12.0 | runs |
| 1.19.13 | absent | minos 12.x | runs |
| 1.20.14 | absent | minos 26.0 | dyld: missing LC_UUID |
| 1.21.x | absent | — | expected to fail |
| 1.22.12 | absent | minos 26.x | dyld: missing LC_UUID |
| 1.23.11 | absent | minos 26.x | dyld: missing LC_UUID |
| 1.25.7 / 1.26.5 | present | minos 26.x | runs |

Go 1.18/1.19 survive only because their linker stamps an old SDK version in `LC_BUILD_VERSION`, so dyld applies the pre-26 rules. That is incidental, not something to rely on.

This branch supports Go 1.18 and later, so the affected Go versions cannot simply be dropped from the matrix. Pinning the runner to `macos-15` keeps the full matrix working. `macos-15` is the arm64 image, the same architecture `macos-latest` resolved to before and after the macOS 26 migration, so test coverage is unchanged.

### Notes for reviewers

- `main` needs no equivalent change: #469 reduced its matrix to Go 1.25+, all of which emit `LC_UUID` by default. The same commit passed on `main` and failed on this branch on the same day.
- Alternatives considered and rejected:
  - `-ldflags=-B=gobuildid` on the macOS test commands works for Go 1.22/1.23, but Go 1.20/1.21 reject the flag (`-B argument must start with 0x: gobuildid`) and cannot emit `LC_UUID` at all.
  - Dropping Go 1.20-1.23 from the macOS matrix loses coverage that `macos-15` can still provide.
- This fix lasts only as long as GitHub offers `macos-15`. Once that label is retired, the Go 1.20-1.23 macOS cells will have to be dropped rather than fixed.
- Job names change from `... on macos-latest` to `... on macos-15`. This branch has no branch protection, so no required status check is pinned to the old names.

---

Authored by Claude (Claude Code), on behalf of @hajimehoshi.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,10 @@ jobs:
   test:
     strategy:
       matrix:
-        os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
+        # macos-15 is used instead of macos-latest. On macOS 26, dyld refuses to
+        # run a binary without an LC_UUID load command, which the Go linker emits
+        # only since Go 1.24. See #482.
+        os: [ubuntu-latest, ubuntu-24.04-arm, macos-15, windows-latest]
         go: ['1.18.x', '1.19.x', '1.20.x', '1.21.x', '1.22.x', '1.23.x', '1.24.x', '1.25.x', '1.26.x']
     name: Test with Go ${{ matrix.go }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}


hajimehoshi:
@TotallyGamerJet PTAL. The merge target is 0.10, not main

TotallyGamerJet (approved):
LGTM

PR #486, opened 2026-07-31
internal/fakecgo: preserve frame pointer across mstart call in threadentry

### LINKED ISSUE #485: internal/fakecgo: threadentry SIGSEGVs in its LEAVE epilogue when a fakecgo-created OS thread exits (amd64, Go 1.28 frame-pointer epilogue)
### PureGo Version

v0.11.0-alpha.8 (commit d476432ce7e3d8bb64b7cd257d332ee221ed6e7b)

### Operating System

Linux (verified). Also affects macOS, FreeBSD and NetBSD on amd64 — the faulting code (`threadentry` + the amd64 frame-pointer epilogue) is shared by all non-Windows amd64 targets. Windows and non-amd64 arches are not affected.

### Go Version (`go version`)

`go1.28-devel` (linux/amd64). The trigger is the amd64 compiler emitting a frame-pointer `LEAVE` epilogue (`MOV BP,SP; POP BP`) for `threadentry`, which is the case on current Go 1.28-devel/tip. Older toolchains that restore SP arithmetically instead of via `LEAVE` do not crash.

Build any `CGO_ENABLED=0` program that links `internal/fakecgo` (so `iscgo` is forced true) and make the Go runtime **create and then exit** OS threads. New threads are routed through `_cgo_thread_start → threadentry_trampoline → threadentry → runtime.mstart`; when such a thread's M exits, `mstart` returns back into `threadentry`.

Minimal reproducer — drop this into the package and run it with cgo disabled:

`internal/fakecgo/threadentry_repro_test.go`
```go
//go:build !cgo && (linux || darwin || freebsd || netbsd)

package fakecgo

import (
	"runtime"
	"sync"
	"testing"
)

// Each goroutine locks its OS thread and then returns without unlocking, so the
// locked (fakecgo-created) M exits and runtime.mstart returns into threadentry.
func TestThreadEntryReturn(t *testing.T) {
	const rounds, workers = 50, 64
	for r := 0; r < rounds; r++ {
		var wg sync.WaitGroup
		wg.Add(workers)
		for i := 0; i < workers; i++ {
			go func() {
				defer wg.Done()
				runtime.LockOSThread()
			}()
		}
		wg.Wait()
	}
}
```

Run:
```sh
CGO_ENABLED=0 go test -run TestThreadEntryReturn ./internal/fakecgo/
```

The test passes and the process exits cleanly — locked OS threads are created and torn down without incident.

### What happens instead?

The process crashes:

```
=== RUN   TestThreadEntryReturn
signal: segmentation fault
FAIL    github.com/ebitengine/purego/internal/fakecgo   0.012s
```

A core dump shows the fault is in `threadentry`'s epilogue, executing `leave` with `RBP = 0`:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  ...fakecgo.threadentry (...) at internal/fakecgo/go_linux.go:56    return nil
#1  threadentry_trampoline () at internal/fakecgo/trampolines_amd64.s:81
#2  start_thread (pthread_create.c:447)

rip  <fakecgo.threadentry+142>
rbp  0x0
=> <fakecgo.threadentry+142>:   leave       // MOV RBP,RSP ; POP RBP  ->  RSP=0, POP [0] -> NULL deref
```

`si_addr` is NULL (`SEGV_MAPERR`). Several other threads are simultaneously mid-`threadentry`/`mstart`.

**Root cause.** `threadentry` (`go_*.go`) calls `runtime.mstart` directly via the closure trick:

```go
fn := uintptr(unsafe.Pointer(&ts.fn))
(*(*func())(unsafe.Pointer(&fn)))()
```

For a fakecgo thread the M's stack is system-allocated, so when the M exits the runtime does `mexit(osStack=true)`, which **returns** from `mstart` back into `threadentry` (this is by design — it lets the pthread unwind and exit). But `mstart` runs the scheduler and comes back with `BP = 0`. On Go 1.28 the amd64 compiler emits a frame-pointer `LEAVE` epilogue for `threadentry`, and `LEAVE` trusts `BP` — so with `BP = 0` it sets `SP = 0` and faults.

Real cgo does not hit this because its `x_cgo_threadentry` invokes `mstart` through `crosscall_amd64`, which **saves and restores the callee-saved registers (including `BP`) on the stack** around the call. `fakecgo`'s `threadentry` skips that step.

**Why it's mostly latent.** Ms normally park and get reused rather than exiting, so `mstart` rarely returns. The bug surfaces when Ms actually **exit** — e.g. a `runtime.LockOSThread` goroutine that finishes (as in the standard library's `os.TestProgWideChdir`), which is exactly how we first hit it. A `LockOSThread` busy-loop that never exits does *not* reproduce it.

**Suggested fix.** Route the `ts.fn`/`mstart` call through a small assembly shim that saves/restores the frame pointer (mirroring `crosscall_amd64`), e.g. using the existing `PUSH_REGS_HOST_TO_ABI0` / `POP_REGS_HOST_TO_ABI0` macros. Note the existing `call5` helper is **not** reusable here because it stashes `SP` in `BP` (`MOV SP,BP`), which `mstart` zeroes — `BP` must be preserved on the stack. I have a working patch (verified: the reproducer above plus `os`/`syscall`/`runtime` all pass with cgo disabled) and will open a PR.

gdams:
Fixes #485.

## Problem

`internal/fakecgo`'s `threadentry` calls `runtime.mstart` directly. For a fakecgo-created thread the M's stack is system-allocated, so when the M exits the runtime does `mexit(osStack=true)`, which **returns** from `mstart` back into `threadentry` (by design, so the pthread can unwind and exit). But `mstart` runs the scheduler and comes back with `BP = 0`.

On Go 1.28 the amd64 compiler emits a frame-pointer `LEAVE` epilogue (`MOV BP,SP; POP BP`) for `threadentry`. With `BP = 0`, `LEAVE` sets `SP = 0` and the following `RET` dereferences a NULL pointer:

```
#0  fakecgo.threadentry (...) at internal/fakecgo/go_linux.go:56    return nil
    => leave        rbp = 0x0        // SIGSEGV, si_addr = NULL
#1  threadentry_trampoline (trampolines_amd64.s)
#2  start_thread
```

Real cgo doesn't hit this because its `x_cgo_threadentry` invokes `mstart` through `crosscall_amd64`, which saves/restores the callee-saved registers (including `BP`) on the stack around the call. fakecgo's `threadentry` skipped that step.

This is mostly latent because Ms usually park and get reused rather than exiting. It surfaces whenever an M actually **exits** — e.g. a `runtime.LockOSThread` goroutine that finishes (which is how the Go standard library's `os.TestProgWideChdir` tripped it under `CGO_ENABLED=0`).

## Fix

Route the `ts.fn` (`runtime.mstart`) call through a small assembly shim, `callThreadEntryFn`, that saves and restores the callee-saved registers on the stack around the call (using the existing `PUSH_REGS_HOST_TO_ABI0` / `POP_REGS_HOST_TO_ABI0` macros), mirroring `crosscall_amd64`. The existing `call5` helper can't be reused because it stashes `SP` in `BP` (`MOV SP,BP`), which `mstart` zeroes.

- `go_linux.go`, `go_darwin.go`, `go_freebsd.go`, `go_netbsd.go`: call `callThreadEntryFn(ts.fn)`.
- `trampolines_amd64.s`: the shim.
- `threadentry_amd64.go`: asm function declaration.
- `threadentry_noasm.go`: plain-call fallback for non-amd64 (unaffected — those epilogues don't rely on `BP`).
- `threadentry_test.go`: regression test that forces locked OS threads to exit.

## Testing

- New `TestThreadEntryReturn` crashes reliably before this change and passes after.
- With the fix vendored into a `CGO_ENABLED=0` + Go 1.28-devel toolchain, the standard library's `os`, `syscall` and `runtime` suites pass again.
- Cross-builds verified for `linux/amd64`, `linux/arm64`, `linux/386`, `darwin/amd64`, `netbsd/amd64`; `gofmt` clean.


internal/fakecgo/threadentry_test.go
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: Apache-2.0
```
hajimehoshi, line 1: Rename this to `threadentry_test.go`

gdams: updated PTAL

internal/fakecgo/threadentry_test.go
```diff
@@ -0,0 +1,47 @@
...
+// this used to crash, because mstart returns with BP clobbered.
+//
+// Locking an OS thread and then letting the goroutine exit (without calling
+// UnlockOSThread) forces the locked fakecgo M to exit, driving mstart to return
+// into threadentry. A busy loop that never lets the goroutine exit does NOT
+// reproduce the crash, because the M never exits and mstart never returns.
+func TestThreadEntryReturn(t *testing.T) {
+	const rounds = 50
+	const workers = 64
+	for r := 0; r < rounds; r++ {
```
hajimehoshi: `for range rounds {`

internal/fakecgo/threadentry_test.go
```diff
@@ -0,0 +1,47 @@
...
+		var wg sync.WaitGroup
+		wg.Add(workers)
+		for i := 0; i < workers; i++ {
```
hajimehoshi: for range workers {

internal/fakecgo/threadentry_test.go
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
+
+// The frame-pointer LEAVE epilogue that this regression guards against is
```
TotallyGamerJet: Wouldn't it be good to test that closing an M on other platforms works?

gdams: yeah good point, I've pushed an update

TotallyGamerJet (approved):
LGTM

internal/fakecgo/go_darwin.go
```diff
@@ -55,9 +55,11 @@ func threadentry(v unsafe.Pointer) unsafe.Pointer {
 	//#endif
 	setg_trampoline(setg_func, uintptr(unsafe.Pointer(ts.g)))
 
-	// faking funcs in go is a bit a... involved - but the following works :)
-	fn := uintptr(unsafe.Pointer(&ts.fn))
-	(*(*func())(unsafe.Pointer(&fn)))()
+	// Call ts.fn (runtime.mstart) through an assembly shim that saves and
+	// restores the frame pointer around the call. mstart returns with BP
+	// clobbered, which would otherwise make this function's frame-pointer
+	// epilogue (LEAVE on amd64) fault when it returns.
```
hajimehoshi: This comment is redundant. callThreadEntryFn's godoc already explains enough. Let's remove this.

internal/fakecgo/go_freebsd.go
```diff
@@ -49,9 +49,11 @@ func threadentry(v unsafe.Pointer) unsafe.Pointer {
 
 	setg_trampoline(setg_func, uintptr(unsafe.Pointer(ts.g)))
 
-	// faking funcs in go is a bit a... involved - but the following works :)
-	fn := uintptr(unsafe.Pointer(&ts.fn))
-	(*(*func())(unsafe.Pointer(&fn)))()
+	// Call ts.fn (runtime.mstart) through an assembly shim that saves and
+	// restores the frame pointer around the call. mstart returns with BP
+	// clobbered, which would otherwise make this function's frame-pointer
+	// epilogue (LEAVE on amd64) fault when it returns.
```
hajimehoshi: ditto

internal/fakecgo/go_linux.go
```diff
@@ -49,9 +49,11 @@ func threadentry(v unsafe.Pointer) unsafe.Pointer {
 
 	setg_trampoline(setg_func, uintptr(unsafe.Pointer(ts.g)))
 
-	// faking funcs in go is a bit a... involved - but the following works :)
-	fn := uintptr(unsafe.Pointer(&ts.fn))
-	(*(*func())(unsafe.Pointer(&fn)))()
+	// Call ts.fn (runtime.mstart) through an assembly shim that saves and
+	// restores the frame pointer around the call. mstart returns with BP
+	// clobbered, which would otherwise make this function's frame-pointer
+	// epilogue (LEAVE on amd64) fault when it returns.
```
hajimehoshi: ditto

internal/fakecgo/go_netbsd.go
```diff
@@ -60,9 +60,11 @@ func threadentry(v unsafe.Pointer) unsafe.Pointer {
 
 	setg_trampoline(setg_func, uintptr(unsafe.Pointer(ts.g)))
 
-	// faking funcs in go is a bit a... involved - but the following works :)
-	fn := uintptr(unsafe.Pointer(&ts.fn))
-	(*(*func())(unsafe.Pointer(&fn)))()
+	// Call ts.fn (runtime.mstart) through an assembly shim that saves and
+	// restores the frame pointer around the call. mstart returns with BP
+	// clobbered, which would otherwise make this function's frame-pointer
+	// epilogue (LEAVE on amd64) fault when it returns.
```
hajimehoshi: ditto

hajimehoshi (approved):
LGTM % comments, thanks!

gdams:
@hajimehoshi updated PTAL

hajimehoshi (approved):
Still LGTM

PR #488, opened 2026-08-03
purego: improve support of 'arm' architecture

### LINKED ISSUE #489: 64-bit arguments and return values are not fully supported on 32-bit ARM
### PureGo Version

v0.10.2

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

1.26.5

For C functions with signatures like

```c
int64_t some_fn(uintptr_t a1, int64_t a2)
```
or

```c
int64_t some_fn(uintptr_t a1, uintptr_t a2, uintptr_t a3, uintptr_t a4, uintptr_t a5, int64_t a6)
```
arguments and return values are corruped when called via purego. This also applies to callbacks.

However, signatures like 
```c
int64_t some_fn(uintptr_t a1, uintptr_t a2, int64_t a3)
```

or

```c
int64_t some_fn(uintptr_t a1, uintptr_t a2, uintptr_t a3, uintptr_t a4, uintptr_t a5, uintptr_t a6, int64_t a7)
```
will work.

64-bit function arguments are properly serialized and return values are properly deserialized

### What happens instead?

I'm getting invalid values

This happens because purego does not comply with 32-bit [AAPCS](https://github.com/ARM-software/abi-aa/blob/main/aapcs32/aapcs32.rst#6111handling-values-larger-than-32-bits). Exact points:
> C.3 If the argument requires double-word alignment (8-byte), the NCRN is rounded up to the next even register number.
   C.7 If the argument required double-word alignment (8-byte), then the NSAA is rounded up to the next double-word address.

### LINKED ISSUE #490: All function/callback calls triggers SIGILL on ARMv5 cpus without FPU
### PureGo Version

v0.10.2

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

1.26.5

Run any example on ARMv5 CPU without FPU. I.e. Marwell Kirkwood which is used on NASes.

Examples are crashing with SIGILL

### What happens instead?

Everything works, functions containing float32/64 arguments and return values can be called via RegisterFunc/RegisterCallback.

Reason: all assembly trampolines in purego for 'arm' are unconditionally saving/restoring FP registers. Go runtime has special variable `goarmsoftfp` to detect soft-float usage and skip FP-related instructions.

Linux/ARM declared as Tier 2 supported target in README and IMHO should be supported equal to Go itself: recognize soft-float usage and act properly.

xakep666:
Closes https://github.com/ebitengine/purego/issues/489 (improper alignment)
Closes https://github.com/ebitengine/purego/issues/490 (soft-float)

Fixes bugs and add feature.

## What this PR does | solves
This PR improves support of 'arm' architecture stated as 'Tier 2' in README.
* **Properly handle 64-bit integers**:
  1. Obey 8-byte alignment on stack and even-register placement required by [AAPCS](https://github.com/ARM-software/abi-aa/blob/main/aapcs32/aapcs32.rst#6111handling-values-larger-than-32-bits). Exact points:
  > C.3 If the argument requires double-word alignment (8-byte), the NCRN is rounded up to the next even register number.
     C.7 If the argument required double-word alignment (8-byte), then the NSAA is rounded up to the next double-word address.
  2. Use two registers (R0, R1) to make function return value. This will work for '386' architecture too.
* **Support for soft-float ABI**:
  1. Pull runtime variable 'goarmsoftfp' to detect if soft-float ABI is used.
  2. Assembly trampolines skips load/store to FP registers if soft-float is detected. This was implemented in exactly same way how it's done in [Go runtime](https://github.com/golang/go/blob/c19862e5f8415b4f24b189d065ed739517c548ba/src/runtime/asm_arm.s#L42).
  4. Apply integer passing rules for float arguments/return values.
* **Run tests for GOARM=5 (no FPU)** in test workflow and add some arm-specific test cases.

**Possible concerns**: a way soft-float ABI usage is detected. I've decided to import a variable from 'runtime' package via `//go:linkname` in a specific way to ensure that `&purego.runtime_goarmsoftfp == &runtime.goarmsoftfp`. Last one has `//go:linkname` too because it's widely used across assembly routines.

**Verification**:
* Added test running for with soft-float ABI to test workflow and added specific test cases.
* Verified on real ARMv5TE CPU (Marvell Kirkwood, no FPU) by @Hirador (co-author of commit, pin-pointed some places to fix).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -146,7 +146,9 @@ jobs:
         if: ${{ runner.os == 'Linux' && runner.arch == 'ARM64' }}
         run: |
           sudo apt-get update
-          sudo apt-get install -y gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf libc6-armhf-cross
+          sudo apt-get install -y \
+            gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf libc6-armhf-cross \
+            gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libc6-armel-cross
 
           # Install the latest QEMU 10.x package from Debian. Ubuntu 24.04 ships QEMU 8.2
           # which has a bug where inter-thread signal delivery (tgkill) hangs,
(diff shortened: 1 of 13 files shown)


hajimehoshi:
> Didn't create an issue. Will do if necessary.

Create an issue first please.

xakep666:
> Create an issue first please.

Done. One issue for improper argument alignment and one for soft-float support.

hajimehoshi:
Thanks.

This PR fixes the two issues at the same time, but can we split the changes?

xakep666:
It is possible. However it's tested together. I've started with soft-float support (found because of SIGILLs) but then found alignment problem.

hajimehoshi:
> It is possible. However it's tested together. I've started with soft-float support (found because of SIGILLs) but then found alignment problem.

Then let's do one by one. Make a separate PR. Thanks,

PR #496, opened 2026-08-12
all: remove zcallback_*.s from the external code list

NotAFlightRisk:

None, spotted it while working out which bits came from the Go runtime.

bug (well, not really - just a small docs issue)

## What this PR does | solves

~~The external code list points at `zcallback_darwin_*.s`, but those got renamed to `zcallback_<GOARCH>.s` in #98 when darwin and linux were merged, so the entry doesn't name a file that exists any more. Every other entry in the list still resolves, so this is the only one.~~

Turns out the line never attributed anything, the files are generated locally by `wincallback.go` which is already listed. Dropping it instead, thanks @hajimehoshi.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,6 @@ This is a list of the copied files:
 
 * `abi_*.h` from package `runtime/cgo`
 * `wincallback.go` from package `runtime`
-* `zcallback_*.s` from package `runtime`
 * `internal/fakecgo/abi_*.h` from package `runtime/cgo`
 * `internal/fakecgo/asm_GOARCH.s` from package `runtime/cgo`
 * `internal/fakecgo/callbacks.go` from package `runtime/cgo`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,7 +108,7 @@ This is a list of the copied files:
 
 * `abi_*.h` from package `runtime/cgo`
 * `wincallback.go` from package `runtime`
-* `zcallback_darwin_*.s` from package `runtime`
+* `zcallback_*.s` from package `runtime`
```
hajimehoshi: Actually the current zcallback_*.s in purego do not correspond to files in the Go repository, if I understand correctly. Then I think we should remove this line

NotAFlightRisk: Ah yeah, you're right. I saw the generated-by header, went straight to fixing the glob. Removed it, cheers @hajimehoshi

hajimehoshi (approved):
LGTM

@TotallyGamerJet PTAL

NotAFlightRisk:
Cheers @hajimehoshi :)

TotallyGamerJet (approved):
LGTM

PR #499, opened 2026-08-20
objc: encode 64-bit integers as q and Q

### LINKED ISSUE #493: objc: encodeType emits 32-bit encodings for 64-bit integer types
### PureGo Version

v0.11.0-alpha.8 (also present at main @ ecac7f7)

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go version go1.26.3 darwin/arm64

Register a class via objc.RegisterClass with a method whose Go signature contains int, int64, uint, or uint64 arguments, then read back the recorded encoding:

```go
class, _ := objc.RegisterClass("Oracle", objc.GetClass("NSObject"), nil, nil,
    []objc.MethodDef{{
        Cmd: objc.RegisterName("takeStackA:b:c:d:e:f:g:h:i:"),
        Fn:  func(_ objc.ID, _ objc.SEL, a1, a2, a3, a4, a5, a6, a7, a8 uint64, a9 uint16) {},
    }})
// read back via class_getInstanceMethod + method_getTypeEncoding
```

Compare against clang:

```objc
printf("%s %s\n", @encode(long), @encode(unsigned long));  // prints: q Q
```

The runtime records `v@:QQQQQQQQS` — on LP64 darwin (every darwin target Go supports), clang @encode spells 64-bit integers "q" (long, long long) and "Q" (unsigned long, unsigned long long).

### What happens instead?

The runtime records `v@:LLLLLLLLS`. objc.encodeType maps reflect.Int to "l" and reflect.Int64, reflect.Uint, and reflect.Uint64 all to "L" — the 32-bit encodings (reflect.Int64 additionally gets an unsigned spelling).

Argument passing is unaffected — purego marshals from the Go types, and Q/L agree on register width — so ordinary objc_msgSend calls work. The wrong string bites consumers that read the declared width from runtime metadata: NSInvocation (getArgument:/setArgument: copy 4 bytes for an L-declared slot, silently truncating 64-bit values), NSMethodSignature and any forwardInvocation:-based proxy, KVC autoboxing from ivar encodings, and block-signature validation (the same derivation feeds block descriptors and ivars).

Suggested fix: map Int and Int64 to "q", Uint and Uint64 to "Q" (constants encLongLong/encULongLong), and delete encLong/encULong, which then have no remaining uses. A test comparing encodeType output per integer kind against clang @encode would prevent the family. I have a fix with such an oracle test ready to send as a PR.

Severity framing: correct-by-@encode fix, oracle-tested, no behavior change for argument passing; fixes silent truncation for NSInvocation/KVC/introspection consumers. The failure shape is silent wrong values (only above 2^32), so it survives small-value testing.

TotallyGamerJet:
Closes #493
bug
## What this PR does | solves

encodeType mapped Go's 64-bit integer kinds to the 32-bit @encode spellings: int to "l", and int64, uint and uint64 all to "L", which additionally gave int64 an unsigned spelling. On darwin, which is LP64 on every architecture Go supports, clang spells 64-bit integers "q" and "Q".

Argument passing is unaffected because purego marshals from the Go types, but consumers that read the declared width back from the runtime metadata - NSInvocation, NSMethodSignature and forwardInvocation: based proxies, KVC autoboxing from ivar encodings - silently truncate values above 2^32.

Check the encoding of each type against clang @encode itself so that the expected encodings cannot drift from the compiler, and assert how encodeFunc assembles a full method signature.

### DIFF
--- a/objc/encoding_darwin_test.go
+++ b/objc/encoding_darwin_test.go
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
 
 package objc
 
--- a/objc/encoding_darwin_test.go
+++ b/objc/encoding_darwin_test.go
@@ -1,0 +1,184 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
+
+package objc
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"reflect"
+	"strings"
+	"testing"
+	"unsafe"
+)
+
+type encodeTypeTestStruct struct {
+	A int32
+	B float64
+}
+
+var encodeTypeTests = []struct {
+	typ   reflect.Type
+	cType string
+	want  string
+}{
+	{reflect.TypeFor[bool](), "_Bool", "B"},
+	{reflect.TypeFor[int8](), "signed char", "c"},
+	{reflect.TypeFor[uint8](), "unsigned char", "C"},
+	{reflect.TypeFor[int16](), "short", "s"},
+	{reflect.TypeFor[uint16](), "unsigned short", "S"},
+	{reflect.TypeFor[int32](), "int", "i"},
+	{reflect.TypeFor[uint32](), "unsigned int", "I"},
+	{reflect.TypeFor[int64](), "long long", "q"},
+	{reflect.TypeFor[uint64](), "unsigned long long", "Q"},
+	{reflect.TypeFor[int](), "long", "q"},
+	{reflect.TypeFor[uint](), "unsigned long", "Q"},
+	{reflect.TypeFor[float32](), "float", "f"},
+	{reflect.TypeFor[float64](), "double", "d"},
+	{reflect.TypeFor[string](), "char *", "*"},
+	{reflect.TypeFor[unsafe.Pointer](), "void *", "^v"},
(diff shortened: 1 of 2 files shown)


objc/encoding_darwin_test.go
```diff
@@ -0,0 +1,184 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 The Ebitengine Authors
```
hajimehoshi: 2026

objc/objc_runtime_darwin.go
```diff
@@ -402,8 +402,8 @@ const (
 	encUShort      = "S"
 	encInt         = "i"
 	encUInt        = "I"
-	encLong        = "l"
-	encULong       = "L"
```
hajimehoshi, line 406: Does this mean we never need `long` or `unsigned long`?

TotallyGamerJet: endlong and encULong both have differing values on 32vs64 bit architectures. It's more clear to use encLongLong and encULongLong because irrespective of arch it is always q/Q encoding. Since Go only supports 64bit architectures for Darwin this made the most sense. The test includes a test for `long` and `unsigned long`. If however, you want it to duplicate the q/Q constants to keep encLong then I can

hajimehoshi (approved):
LGTM, thanks!

objc/objc_runtime_darwin.go
hajimehoshi: That's fine. I don't think we need to keep encLong and encULong for non-existing 32bit architectures.

PR #500, opened 2026-08-23
purego: fix returning a C function pointer as a Go func

NotAFlightRisk:

#501

bug

## What this PR does | solves

The `RegisterFunc` type table has `func <=> C function` in it, but that only holds in the argument direction. Declare a C function as returning a func, and calling it panics:

```
panic: reflect.MakeFunc: value of type *func(int32, int32) int32 is not assignable to type func(int32, int32) int32
```

Registering is fine, it's the call that goes, and it goes for every signature. Every other case in that return switch leaves the result as the declared type. The `reflect.Func` case swapped in a `*func`, registered into that, and handed the pointer back to `MakeFunc`, which wont take it.

If the C function returns NULL the inner `RegisterFunc` still panics `purego: cfn is nil`. Left that alone, since whether it should hand you a nil Go func instead looked like your call rather than a bug.

Test is a block in `TestABI` plus a `return_func_ptr` helper in `abi_test.c`, and it fails on main with teh panic above. Been broken since #49 back in 2022, which is probably why nobody's hit it: you'd normally take the pointer back as a `uintptr` and register it youself, and nothing in the repo declares a func return.

### DIFF
--- a/func.go
+++ b/func.go
@@ -409,7 +409,9 @@ func RegisterFunc(fptr any, cfn uintptr) {
 			v = reflect.NewAt(outType, unsafe.Pointer(&a1)).Elem()
 		case reflect.Func:
 			// wrap this C function in a nicely typed Go function
-			RegisterFunc(v.Addr().Interface(), syscall.a1)
+			if syscall.a1 != 0 {
+				RegisterFunc(v.Addr().Interface(), syscall.a1)
+			}
 		case reflect.String:
 			v.SetString(strings.GoString(syscall.a1))
 		case reflect.Float32:
--- a/func.go
+++ b/func.go
@@ -409,8 +409,7 @@ func RegisterFunc(fptr any, cfn uintptr) {
 			v = reflect.NewAt(outType, unsafe.Pointer(&a1)).Elem()
 		case reflect.Func:
 			// wrap this C function in a nicely typed Go function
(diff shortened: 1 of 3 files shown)


func.go
```diff
@@ -409,8 +409,7 @@ func RegisterFunc(fptr any, cfn uintptr) {
 			v = reflect.NewAt(outType, unsafe.Pointer(&a1)).Elem()
 		case reflect.Func:
 			// wrap this C function in a nicely typed Go function
-			v = reflect.New(outType)
-			RegisterFunc(v.Interface(), syscall.a1)
+			RegisterFunc(v.Addr().Interface(), syscall.a1)
```
hajimehoshi:
```go
if syscall.a1 != 0 {
    RegisterFunc(v.Addr().Interface(), syscall.a1)
}
```

and add a test for NULL case

NotAFlightRisk: Done, cheers @hajimehoshi. I'd only tested it with a real function pointer so never hit the null path - the test asserts the func stays nil now.

TotallyGamerJet:
> None. Happy to file one if you'd rather have it tracked.

Yes please create one

hajimehoshi (approved):
LGTM

@TotallyGamerJet PTAL

TotallyGamerJet (approved):
LGTM

PR #502, opened 2026-08-24
all: use structs.HostLayout on structs that must match the C ABI

### LINKED ISSUE #259: Use Go 1.23's new structs.HostLayout

TotallyGamerJet:
Closes #259

feature

## What this PR does | solves
Go 1.23 added structs.HostLayout, a zero-sized marker field that tells the compiler a struct must be laid out the way the host's C ABI expects rather than by Go's own rules. Until now purego relied on Go's current layout happening to match C.

Declare the marker on every struct that stands in for a C one: objc_super, MethodDescription, PropertyAttribute, blockDescriptor and blockLayout in package objc; G, ThreadStart, the Darwin pthread types, the NetBSD stack_t and argset in internal/fakecgo; syscallArgs, which internal/cgo hands to C as a struct syscallArgs; the window example's NSPoint/NSSize/NSRect and WNDCLASSEX/RECT/POINT/MSG; and the C-representative structs in the tests. Structs that never cross into C, such as MethodDef, FieldDef and the callback argument block filled in by Go assembly, are left alone.

The marker field is zero-sized, so it takes no part in the C ABI. Teach the argument classification to skip zero-sized fields, otherwise a struct that declares it is misclassified: isAllSameFloat would panic descending into the marker, isHFA/isHVA would stop recognizing homogeneous aggregates, and objc's @encode would emit the marker as a member.

### DIFF
--- a/func.go
+++ b/func.go
@@ -566,7 +566,7 @@ func isABIField(f reflect.StructField) bool {
 // numABIFields returns how many of ty's fields take part in the C ABI.
 func numABIFields(ty reflect.Type) int {
 	var n int
-	for i := 0; i < ty.NumField(); i++ {
+	for i := range ty.NumField() {
 		if isABIField(ty.Field(i)) {
 			n++
 		}
@@ -577,7 +577,7 @@ func numABIFields(ty reflect.Type) int {
 // abiField returns the i'th field of ty that takes part in the C ABI. It panics
 // if ty has fewer than i+1 such fields.
 func abiField(ty reflect.Type, i int) reflect.StructField {
-	for j := 0; j < ty.NumField(); j++ {
+	for j := range ty.NumField() {
 		f := ty.Field(j)
 		if !isABIField(f) {
 			continue
--- a/func.go
+++ b/func.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"reflect"
 	"runtime"
+	"structs"
 	"sync"
 	"unsafe"
 
@@ -555,12 +556,12 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
 // If you change this make sure to update it in objc_runtime_darwin.go
 const maxRegAllocStructSize = 16
 
+var hostLayoutType = reflect.TypeFor[structs.HostLayout]()
+
 // isABIField reports whether f takes part in the C ABI of the struct that
-// contains it. Zero-sized fields occupy no storage, so a C compiler never sees
-// them as members. The structs.HostLayout marker that a C-representative
-// struct declares is the usual example.
+// contains it. Only the structs.HostLayout marker does not.
 func isABIField(f reflect.StructField) bool {
-	return f.Type.Size() != 0
+	return !f.Type.ConvertibleTo(hostLayoutType)
 }
 
 // numABIFields returns how many of ty's fields take part in the C ABI.
--- a/func.go
+++ b/func.go
@@ -555,17 +555,61 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
 // If you change this make sure to update it in objc_runtime_darwin.go
 const maxRegAllocStructSize = 16
 
+// isABIField reports whether f takes part in the C ABI of the struct that
+// contains it. Zero-sized fields occupy no storage, so a C compiler never sees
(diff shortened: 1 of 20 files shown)


func.go
```diff
@@ -555,17 +555,61 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
...
+			n++
+		}
+	}
+	return n
+}
+
+// abiField returns the i'th field of ty that takes part in the C ABI. It panics
+// if ty has fewer than i+1 such fields.
+func abiField(ty reflect.Type, i int) reflect.StructField {
+	for j := 0; j < ty.NumField(); j++ {
```
hajimehoshi: Use for-range

func.go
```diff
@@ -555,17 +555,61 @@ func addValue(v reflect.Value, keepAlive []any, addInt func(x uintptr), addFloat
...
+// them as members. The structs.HostLayout marker that a C-representative
+// struct declares is the usual example.
+func isABIField(f reflect.StructField) bool {
+	return f.Type.Size() != 0
+}
+
+// numABIFields returns how many of ty's fields take part in the C ABI.
+func numABIFields(ty reflect.Type) int {
+	var n int
+	for i := 0; i < ty.NumField(); i++ {
```
hajimehoshi: Use for-range

hajimehoshi (commented):
I think adding a check of `isABIField` can be a separate change from introducing `LayoutHost`, as `isABIField` seems to be related to other zero-sized structs. What do you think?

TotallyGamerJet:
> I think adding a check of isABIField can be a separate change from introducing LayoutHost, as isABIField seems to be related to other zero-sized structs. What do you think?

We do need to exclude HostLayout from the calculated ABI otherwise we get incorrect argument passing. It's now specific only to HostLayout though

objc/objc_runtime_darwin.go
```diff
@@ -470,6 +474,13 @@ func encodeType(typ reflect.Type, insidePtr bool) (string, error) {
 		encoding.WriteString("=")
 		for i := 0; i < typ.NumField(); i++ {
 			f := typ.Field(i)
+			if f.Type.ConvertibleTo(hostLayoutType) {
```
hajimehoshi, line 477: `!isABIField(f)`?

struct_amd64.go
```diff
@@ -66,23 +66,24 @@ func getStruct(outType reflect.Type, syscall syscallArgs) (v reflect.Value) {
...
+			if f1.Kind() == reflect.Float64 || f1.Kind() == reflect.Float32 && abiField(outType, 1).Type.Kind() == reflect.Float32 {
 				r1 = syscall.f1
 				hasFirstFloat = true
 			}
 
 			// find index of the field that starts the second 8 bytes
 			var i int
-			for i = 0; i < outType.NumField(); i++ {
-				if outType.Field(i).Offset == 8 {
+			for i = 0; i < numFields; i++ {
```
hajimehoshi, line 78: Optional: `for i := range numFields`

(There are a lot of for-range candidates, so we can do them once later)

objc/objc_runtime_darwin.go
TotallyGamerJet: yes but isABIField is defined in func.go so isn't accessible here

hajimehoshi: Ah, I see, then that's fine

struct_amd64.go
TotallyGamerJet: yeah I think doing it separately is okay

hajimehoshi (approved):
LGTM

PR #503, opened 2026-08-29
all: use range-over-int in counting loops

hajimehoshi:
None; this is a standalone cleanup.

refactor (no behavior change)

## What this PR does | solves
Go 1.22 added range over an integer. This converts the 37 three-clause
counting loops that translate directly, most of them walking reflect fields
or function arguments, and drops the index where the body never used it.

Found with `go fix -rangeint ./...` run across the GOOS/GOARCH matrix so the
per-architecture files were covered, then extended by hand to the cases that
analyzer skips conservatively:

- Call limits it cannot prove invariant: `NumIn`, `NumField`, `Len`, `NumCPU`
  and `numABIFields`. All are pure, so evaluating them once instead of on
  every iteration does not change behavior. `func.go` already used this form.
- `numFields` in the `place` closures of `struct_amd64.go` and
  `struct_arm64.go`, assigned in two branches before the loop and never
  inside it.
- `wincallback.go`, which carries `//go:build ignore` and so is never
  analyzed. It generates the `zcallback_*.s` files; regenerating with the
  converted loops reproduces all eight byte for byte.

Loops a range cannot express are left as they are: the 8-byte strides in
`copyStruct8ByteChunks`, `placeStack` and `addStruct`; the field scan in
amd64 `getStruct`, whose index is read after the loop; the compound
condition in `setStruct`; and the one-based loops in `objc` that skip the
first argument, emitted separately as `encId`.

`gofmt -s -l` is clean, `go vet` is clean on 10 platforms and `go build` on
all 15 that build today, and `go test ./...` passes on darwin/arm64.

---
Authored by Claude (Claude Code), on behalf of @hajimehoshi.

### DIFF
--- a/func.go
+++ b/func.go
@@ -151,14 +151,14 @@ func RegisterFunc(fptr any, cfn uintptr) {
 		floatArgRegs := numOfFloatRegisters()
 		ptrSize := unsafe.Sizeof(uintptr(0))
 		var stack int
-		for i := 0; i < ty.NumIn(); i++ {
+		for i := range ty.NumIn() {
 			arg := ty.In(i)
 			switch arg.Kind() {
 			case reflect.Func:
 				// This only does preliminary testing to ensure the CDecl argument
 				// is the first argument. Full testing is done when the callback is actually
 				// created in NewCallback.
-				for j := 0; j < arg.NumIn(); j++ {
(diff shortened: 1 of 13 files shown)


TotallyGamerJet (approved):
LGTM

PR #504, opened 2026-08-30
purego: apply go fix modernizations

hajimehoshi:
None; no issue filed.

Neither bug, feature, nor security: a maintenance/cleanup change.

## What this PR does | solves
Go 1.27's `go fix` rewrites constructs that newer standard library APIs express better. Running it over every GOOS/GOARCH this module builds for turns up three sites, all mechanical:

- `compareStatus` walks the lines of a `/proc` status file, so the slice from `strings.Split` only exists to be ranged over; `SplitSeq` yields the same lines without materializing it.
- The same function sorts a field list with `sort.Slice` and a less function that is plain string comparison, which is what `slices.Sort` does for an ordered element type.
- `NewCallback` asks whether an argument is assignable to `CDecl`, a type known at compile time, so `reflect.TypeFor` names it directly instead of building a value to take the type of.

`go fix` only analyzes the files selected by the current build constraints, so it was run for each of darwin, linux, freebsd, netbsd, windows, android, ios and js across the architectures purego supports, with cgo both on and off, plus the `faketime` tag. Only the two files here reported anything.

The cgo-enabled cross builds could not be analyzed on the machine this ran on, which has no cross C toolchain; darwin/amd64 and darwin/arm64 with cgo were covered and reported nothing.

Verified with `gofmt -l -s` (clean), `go vet` for linux and windows, `go build` for windows, and `go test ./...` on darwin/arm64.

### DIFF
--- a/syscall_linux_test.go
+++ b/syscall_linux_test.go
@@ -10,7 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"sort"
+	"slices"
 	"strings"
 	"syscall"
 	"testing"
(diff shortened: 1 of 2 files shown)


TotallyGamerJet (approved):
LGTM

PR #510, opened 2026-09-05
purego: stop echoing input a3 as errno on non-darwin

### LINKED ISSUE #505: SyscallN third return value echoes input a3 instead of errno on non-darwin
### PureGo Version

main @ 63002e0 (after #504)

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [x] FreeBSD
- [x] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64

`SyscallN`'s third return value is documented as "the libc error code", but on every OS except darwin it just echoes the *input* `a3` argument. The errno write-back in `sys_*.s` is guarded by `#ifdef GOOS_darwin`:

```asm
// sys_amd64.s:125-129
#ifdef GOOS_darwin
	CALL purego_error(SB)
	MOVQ PTR_ADDRESS(SP), DI      // reload (DI clobbered by call)
	MOVQ (AX), AX
	MOVQ AX, syscallArgs_a3(DI) // save errno
#endif
```

while `syscall.go:89` unconditionally returns it:

```go
s := syscall_SyscallN(fn, tmp[:], floats[:], 0)
defer thePool.Put(s)
return s.a1, s.a2, s.a3   // s.a3 was never written on Linux/FreeBSD/NetBSD
```

Repro (Linux):

```go
libc, _ := purego.Dlopen("libc.so.6", purego.RTLD_NOW|purego.RTLD_GLOBAL)
sym, _ := purego.Dlsym(libc, "open")
path := append([]byte("/nonexistent_purego_xyz"), 0)
r1, _, errno := purego.SyscallN(sym, uintptr(unsafe.Pointer(&path[0])), 0)
runtime.KeepAlive(path)
fmt.Println(r1, errno) // r1 = -1 (0xffff...), errno = 0; want errno = ENOENT (2)
```

Actual output on Linux: `open: r1=-1 errno=0`. The `0` is just the input `a3` (`flags=0`) echoed back, not `ENOENT`.

The same `#ifdef GOOS_darwin` pattern exists in `sys_arm64.s:126-129` (and the other arch trampolines), so Linux/arm64, FreeBSD, and NetBSD are all affected.

On Linux/FreeBSD/NetBSD the third return value should be the libc error code (`ENOENT == 2` in the repro above), matching the documented behavior of `SyscallN` and the darwin behavior.

### What happens instead?

The caller observes whatever was passed as the third argument (`a3`). Any `errno != 0` check against a successful call, or any error-code inspection after a failed call, reads garbage on non-darwin platforms.

Minimized reproducible case: the program above (needs only `libc.so.6`, no C toolchain). Verified failing on `main @ 63002e0`, Linux/amd64, Go 1.26.7.

Suggested fix: on non-darwin either return `0` explicitly and document that `SyscallN` only reports errno on darwin, or fetch the error via `__errno_location` / `___errno` per-OS like the darwin path does with `purego_error`. At minimum the doc comment on `SyscallN` ("It returns the result and the libc error code if there is one") must stop promising errno on platforms that never set it.

kumagi:
#505

bug

## What this PR does | solves
The syscall trampolines only write errno back into the args block on darwin (see the GOOS_darwin guards in sys_*.s), so on Linux, FreeBSD and NetBSD SyscallN returned the caller's third argument as a fake error code. Return 0 there instead and document that err is only the libc error code on darwin and Windows.

Closes #505


hajimehoshi:
The new Darwin-only condition in `syscall.go` also discards valid errno values from the Linux C fallback. For example, linux/mips with CGO_ENABLED=1 uses `syscall_cgo_linux.go`, which routes calls through `internal/cgo/syscall_cgo_unix.go`; that implementation already writes `errno` into `args->a3`. Returning zero here therefore regresses error reporting on those targets.

Could this preserve the C fallback result and only return zero for trampolines that do not capture errno? A regression test covering the fallback would also help.

I confirmed the linux/mips source selection with `go list`; I did not run runtime tests.

Posted by Codex (OpenAI), on behalf of @hajimehoshi.

syscall.go
```diff
@@ -49,7 +49,9 @@ func syscall_SyscallN(fn uintptr, sysargs []uintptr, floats []uintptr, r8 uintpt
 
 // SyscallN takes fn, a C function pointer and a list of arguments as uintptr.
 // There is an internal maximum number of arguments that SyscallN can take. It panics
-// when the maximum is exceeded. It returns the result and the libc error code if there is one.
+// when the maximum is exceeded. It returns the result and, on darwin and
+// Windows, the libc error code if there is one. On other platforms errno is
+// not captured so err is always 0.
```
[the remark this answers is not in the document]
kumagi: Fixed. The comment now says that on darwin, and on the Linux architectures that go through the C fallback, `err` is the libc errno, while on Windows it is what `syscall.SyscallN` returns as its error, which is a GetLastError-style code rather than a libc errno.

syscall.go
```diff
@@ -86,5 +88,10 @@ func SyscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 	copy(floats[:], tmp[:])
 	s := syscall_SyscallN(fn, tmp[:], floats[:], 0)
 	defer thePool.Put(s)
-	return s.a1, s.a2, s.a3
+	if runtime.GOOS == "darwin" {
+		return s.a1, s.a2, s.a3
+	}
+	// The trampolines only save errno on darwin. Returning s.a3 here
+	// would echo the caller's third argument back as a fake error code.
+	return s.a1, s.a2, 0
```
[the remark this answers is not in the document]
kumagi: Good point about the fallback: the run-time Darwin check here was also throwing away the errno that `internal/cgo` writes into `a3` on the Linux architectures without an assembly trampoline. The condition is now the build-tagged `capturesErrno` constant, which is true on darwin and under the build constraint of `syscall_cgo_linux.go`, so the C fallback result is preserved.

There is also a test now: `TestSyscallNErrnoIsNotAnInputArgument` calls `open` with a non-zero third argument and requires `err == 0` on platforms that do not capture errno; with the old code it returned 384. `TestErrno` covers the opposite case and now runs on the fallback instead of being Darwin-only.

syscall_32bit.go
```diff
@@ -51,7 +51,9 @@ func syscall_SyscallN(fn uintptr, sysargs []uintptr, floats []uintptr, r8 uintpt
 
 // SyscallN takes fn, a C function pointer and a list of arguments as uintptr.
 // There is an internal maximum number of arguments that SyscallN can take. It panics
-// when the maximum is exceeded. It returns the result and the libc error code if there is one.
+// when the maximum is exceeded. It returns the result and, on darwin and
+// Windows, the libc error code if there is one. On other platforms errno is
+// not captured so err is always 0.
```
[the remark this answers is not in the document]
kumagi: Fixed. The wording now describes the Windows value as what `syscall.SyscallN` returns as its error (a GetLastError-style code, not a libc errno) and states that on the other platforms this file builds for, errno is not captured so `err` is always 0.

syscall_32bit.go
```diff
@@ -88,5 +90,10 @@ func SyscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 	copy(floats[:], tmp[:16])
 	s := syscall_SyscallN(fn, tmp[:], floats[:], 0)
 	defer thePool.Put(s)
-	return s.a1, s.a2, s.a3
+	if runtime.GOOS == "darwin" {
+		return s.a1, s.a2, s.a3
+	}
+	// The trampolines only save errno on darwin. Returning s.a3 here
+	// would echo the caller's third argument back as a fake error code.
+	return s.a1, s.a2, 0
```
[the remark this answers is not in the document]
kumagi: Removed. This file's constraint never selects a backend that saves errno (darwin is excluded and 386/arm always use their own trampolines rather than the C fallback), so it just returns 0 now with a comment saying why.

syscall_ppc64le.go
```diff
@@ -37,7 +37,9 @@ func syscall_SyscallN(fn uintptr, sysargs []uintptr, floats []uintptr, r8 uintpt
 
 // SyscallN takes fn, a C function pointer and a list of arguments as uintptr.
 // There is an internal maximum number of arguments that SyscallN can take. It panics
-// when the maximum is exceeded. It returns the result and the libc error code if there is one.
+// when the maximum is exceeded. It returns the result and, on darwin and
+// Windows, the libc error code if there is one. On other platforms errno is
+// not captured so err is always 0.
```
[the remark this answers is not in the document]
kumagi: Fixed. Since this file builds only for linux/ppc64le, the comment no longer mentions darwin or Windows at all: it says the trampoline does not capture errno there, so `err` is always 0.

kumagi:
Thanks for catching this — the Darwin-only check was too coarse and discarded the errno that the C fallback really does report.

- Whether `a3` holds errno is now decided where the backend is selected: a build-tagged `capturesErrno` constant is true on darwin (the `GOOS_darwin` guards in `sys_*.s`) and under exactly the build constraint of `syscall_cgo_linux.go`, so linux/mips and the other Linux architectures that go through `internal/cgo` keep returning the libc errno. Every other platform still returns 0 instead of echoing `a3`.
- The docs no longer call the Windows value a libc error code; it is what `syscall.SyscallN` returns, i.e. a GetLastError-style code. The libc errno wording is now used only where an errno is actually captured.
- The unreachable `runtime.GOOS == "darwin"` branch in `syscall_32bit.go` is removed, since that file's constraint excludes every configuration that captures errno, and `syscall_ppc64le.go` documents the linux/ppc64le-only behavior.
- Tests: `TestErrno` no longer skips everything but darwin; it runs wherever errno is captured (opening `libc.so.6` on Linux), which is the regression test for the C fallback. `TestSyscallNErrnoIsNotAnInputArgument` covers the other side: it fails to open a file with a non-zero third argument and requires `err == 0` on platforms that do not capture errno. That test reports 384 with the old unconditional `return s.a1, s.a2, s.a3`, so it does pin the behavior.

Source selection was verified with `go list` for each GOOS/GOARCH the package builds for; exactly one of the two files defining `capturesErrno` is selected everywhere.

hajimehoshi:
Thanks, the backend-specific condition addresses the earlier errno regression. There is one remaining issue with the new C fallback coverage in `TestErrno`.

On Linux/mips, `SyscallN` copies its integer arguments into the float slots (`copy(floats[:], tmp[:])`), but `internal/cgo/syscall_cgo_unix.go` asserts that all eight float slots are zero. The nonzero pathname pointer passed to `open` therefore triggers that assertion before `open` runs, so the newly enabled test aborts instead of checking errno.

Could the fallback regression test use a zero-argument C helper that sets errno, or could the existing argument-handling limitation be addressed? The assertion is pre-existing; enabling this test on the fallback exposes it.

I confirmed the assertion failure with a native C probe using the fallback source. I did not run cross-architecture Go runtime tests.

Posted by Codex (OpenAI), on behalf of @hajimehoshi.

kumagi:
Reproduced — thanks for the catch. `SyscallN` mirrors every integer argument into the float slots and `internal/cgo`'s `syscall15` asserts that all eight of them are zero, so passing the pathname pointer to `open` aborted on that assertion before `open` ever ran.

`TestErrno` now calls a helper that takes no arguments instead: `testdata/liberrnotest/errno_test.c` sets `errno = ENOENT` and returns -1. The test builds it with the same `buildSharedLib` helper the other tests use, calls it with no arguments so every float slot stays zero, and still converts the returned value with libc's `strerror`, so it keeps proving that the errno the backend writes into `a3` reaches the caller. It is not a vacuous test: with the `a3` read dropped it reports Success instead of No such file or directory. It behaves the same on darwin, where the mirrored float slots are consumed by the trampoline rather than asserted about.

I left the argument handling alone. The assertion is pre-existing and is reached by any `SyscallN` call with a non-zero argument on those architectures — the existing `TestSyscallN` trips it the same way — so narrowing it looked like a separate change, and this commit changes no library behavior at all. The test only runs where errno is captured, i.e. on darwin or on the Linux architectures that already require a C toolchain, so the extra compile step does not need a toolchain anywhere new.

Verified without a mips runtime by routing linux/amd64 through the `internal/cgo` fallback: the `open` version aborts on the assertion, the helper version reports ENOENT, and the helper version with the `a3` read removed reports Success. The native suite still passes with and without Cgo, and the `capturesErrno` file selection is unchanged. If you would rather have `SyscallN` stop mirroring its integer arguments into the float slots where the C fallback is used, I am happy to do that here or in a follow-up.

testdata/liberrnotest/errno_test.c
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
+
+#include <errno.h>
+
+/* setErrno fails the way a C library call does: it returns -1 and sets errno.
+
+   It takes no argument on purpose. SyscallN mirrors every integer argument into
+   the float slots and the C fallback used on the Linux architectures that have
+   no assembly trampoline asserts that those slots are zero, so a call with
+   arguments would abort there before the function is reached. */
```
hajimehoshi: Can we use `//` for C comments?

dlfcn_test.go
```diff
@@ -48,3 +48,30 @@ func TestSyscallN(t *testing.T) {
 		t.Fatalf("SyscallN didn't return the same result as purego.Dlsym: %d", err2)
 	}
 }
+
+// TestSyscallNErrnoIsNotAnInputArgument checks that SyscallN doesn't return the
+// caller's third argument as an error code on platforms where the trampoline
+// doesn't save errno. See https://github.com/ebitengine/purego/issues/505.
```
hajimehoshi: I don't think we need this comment

dlfcn_test.go
```diff
@@ -48,3 +48,30 @@ func TestSyscallN(t *testing.T) {
...
+
+	openSym, err := purego.Dlsym(purego.RTLD_DEFAULT, "open")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// The third argument is non-zero so that echoing it back as errno is visible.
+	const third = uintptr(0o600)
+	r1, _, errno := purego.SyscallN(openSym,
+		uintptr(unsafe.Pointer(&[]byte("_file_that_does_not_exist_\x00")[0])),
```
hajimehoshi: Use `syscall.BytePtrFromString`

syscall_errno_saved.go
```diff
@@ -0,0 +1,16 @@
...
+package purego
+
+// capturesErrno reports whether the trampoline behind syscallXABI0 writes
+// errno back into the a3 field of syscallArgs.
+//
+// The assembly trampolines do that only on darwin - see the GOOS_darwin guards
+// in sys_*.s - while the C fallback in internal/cgo, which is used on the Linux
+// architectures that have no assembly trampoline, always does.
+//
+// Keep this build constraint in sync with syscall_cgo_linux.go.
```
hajimehoshi: > syscall_cgo_linux.go

This might be correct but asymmetry with syscall_errno_not_saved.go

syscall_errno_saved.go
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 The Ebitengine Authors
+
+//go:build darwin || (linux && cgo && !(386 || amd64 || arm || arm64 || loong64 || ppc64le || riscv64 || s390x))
```
hajimehoshi: Which Linux makes capturesErrno true? My understanding that Linux with architectures that require Cgo for PureGo, but is that correct?

testdata/liberrnotest/errno_test.c
kumagi: Done - the helper now uses // comments, matching the rest of the repository including the SPDX headers of the other testdata C files. It still compiles with -Wall -Werror under both gcc and clang.

dlfcn_test.go
kumagi: Removed; the test name says what it checks.

dlfcn_test.go
kumagi: Done, the path now comes from syscall.BytePtrFromString, which also drops the hand-written NUL. The symbol is part of the common syscall API, and go test -c still succeeds for every platform this file builds for.

syscall_errno_saved.go
kumagi: Made symmetric: both files now carry exactly the same comment and both point at syscall_cgo_linux.go, which is where the backend is actually selected, instead of one pointing at the C fallback and the other at its counterpart.

syscall_errno_saved.go
kumagi: Yes, your understanding is correct: capturesErrno is true on Linux exactly where purego has no assembly trampoline and therefore needs Cgo, i.e. where syscall_cgo_linux.go is selected. go list with CGO_ENABLED=1 for every linux GOARCH puts syscall_errno_saved.go on mips, mipsle, mips64, mips64le and ppc64, and syscall_errno_not_saved.go on 386, amd64, arm, arm64, loong64, ppc64le, riscv64 and s390x; without Cgo those five fall back to syscall_errno_not_saved.go because the package does not build there at all. The comment now states that, so the constraint does not have to be decoded to answer the question.

kumagi:
Addressed the new round of comments. They are all comment or test polish, so no behavior changes:

- `testdata/liberrnotest/errno_test.c` now uses `//` comments like the rest of the repository.
- `TestSyscallNErrnoIsNotAnInputArgument` drops its doc comment and builds its C string with `syscall.BytePtrFromString`.
- The two files defining `capturesErrno` now carry the same comment and both point at `syscall_cgo_linux.go`, where the backend is actually chosen. That comment also names the Linux architectures the C fallback is selected on - mips, mipsle, mips64, mips64le and ppc64 today - which answers the question about when `capturesErrno` is true on Linux.

Verified: `gofmt -s` and `go vet ./...` are clean, the suite passes natively with and without Cgo and with `-race`, test binaries still cross-compile for darwin, netbsd, windows and the Linux architectures, and `TestErrno` still passes when the package is routed through the `internal/cgo` fallback.

TotallyGamerJet (commented):
This just seems very complicated for what it's trying to fix. Why not just zero the a3 field at the end of each SyscallN in the assembly. That way when we get around to returning errno on other platforms we don't have to track down a build flag update

kumagi:
Fair question - the flag is two single-constant files, but the trade-off is real. Two reasons it currently sits in Go rather than in the assembly:

- The tests need the same knowledge. `TestErrno` has to know whether to expect ENOENT and `TestSyscallNErrnoIsNotAnInputArgument` whether to expect 0, so the predicate exists either way. If the trampolines own the zeroing it becomes a build-tagged constant in `package purego_test` instead of one the library uses too; as it stands there is exactly one copy of the constraint and the tests read it through `CapturesErrno`.
- It fails in the safer direction. An architecture that is missing from both lists gets `capturesErrno == false`, i.e. "no errno", while a trampoline that forgets to zero `a3` hands the caller its own third argument back as an error code, which is precisely the bug this CL fixes (#505).

What the Go version costs is what you point out: teaching a platform to report errno means touching both build constraints in addition to saving errno in the trampoline. What the assembly version costs is an extra store to `a3` on every C call on every non-darwin platform - cheap next to the `a1`/`a2`/`f1`..`f8` write-backs that are already there, but on the path `syscall_bench_test.go` measures - plus the "zero a3, or save errno" rule has to be repeated in each of the eight `sys_*.s` trampolines instead of stated once.

Both work functionally. I lean toward keeping this shape because it stays in Go and fails safe, but if you and @hajimehoshi would rather have the trampolines own it, I'll switch: zero `a3` at the end of each trampoline, drop `capturesErrno` from the library, and keep the same build-tagged constant on the test side.

TotallyGamerJet:
I prefer to make the change in the assembly by clearing a3 field. Of course the test needs to know which platforms return errorno which can be obvious with an `if runtime.GOOS == ...` instead of adding two files whose only purpose for existing to to key off the environment.

syscall.go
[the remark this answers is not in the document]
kumagi: Update after @TotallyGamerJet's suggestion: the build-tagged constant is gone. The trampolines that cannot capture errno now clear `a3` themselves, and the C fallback has always written errno into it, so `SyscallN` returns `a1`, `a2` and `a3` unconditionally and the decision sits where errno is or is not captured. The behavior described in this thread is unchanged - errno on darwin and on the C fallback, 0 elsewhere - and both regression tests still pin it; they just use a `runtime.GOOS`/`GOARCH` switch to know which one to expect.

syscall_errno_saved.go
kumagi: Superseded by the follow-up commit: the asymmetry is moot because both files defining capturesErrno are gone. The trampolines now clear `a3` where they cannot capture errno, so there is no build constraint left to keep in step with syscall_cgo_linux.go.

syscall_errno_saved.go
kumagi: For the record your understanding was right: the C fallback, and therefore errno, is used on the Linux architectures that have no assembly trampoline and need Cgo (mips, mipsle, mips64, mips64le and ppc64 today). The follow-up drops capturesErrno altogether, and the same list now lives as a `runtime.GOOS`/`GOARCH` switch in the tests, which is the only place that still needs it.

kumagi:
Switched to clearing `a3` in the assembly, as requested.

- Every trampoline that cannot capture errno now clears `a3` after the call instead of leaving the caller's third argument there, so the rule lives next to the code that decides whether errno exists. `sys_amd64.s` and `sys_arm64.s` keep the darwin errno save behind the existing `GOOS_darwin` guard and clear the field in the `#else` branch. `internal/cgo`'s `syscall15` has always written errno, so the C fallback is untouched.
- `SyscallN` returns `a1`, `a2` and `a3` unconditionally again, and `capturesErrno`, its two build-tagged files and the `CapturesErrno` export are gone. The tests use a `runtime.GOOS`/`runtime.GOARCH` switch, which is all they need to know about a platform.

Two notes from testing the encodings: `arm` and `ppc64le` cannot store an immediate, and on `ppc64le` `MOVD $0, off(R11)` is encoded as a store from R0, which is not zero after a call, so both use a register that is already dead by that point instead. The other ports store a zero directly (`movq $0`, `movl $0`, `mvghi`, ZR, `x0`, `$r0`).

Verified: removing the new amd64 store makes `TestSyscallNErrnoIsNotAnInputArgument` report 384 again, so the clearing is what the test pins; with linux/amd64 routed through the `internal/cgo` fallback `TestErrno` still reports ENOENT, so the fallback errno is unaffected; the package assembles and builds for every GOOS/GOARCH it supports, and the suite passes natively with and without Cgo and with `-race`.

syscall_test.go
```diff
@@ -25,24 +25,65 @@ func TestOS(t *testing.T) {
...
 
-	r1, _, errno := purego.SyscallN(openSym, uintptr(unsafe.Pointer(&[]byte("_file_that_does_not_exist_\x00")[0])), uintptr(os.O_RDWR))
+	r1, _, errno := purego.SyscallN(setErrno)
 	if int32(r1) != -1 {
-		t.Errorf("open returned %d, wanted -1", r1)
+		t.Errorf("setErrno returned %d, wanted -1", r1)
+	}
+
+	libcName := "/usr/lib/libSystem.B.dylib"
+	if runtime.GOOS == "linux" {
```
TotallyGamerJet: why not use `getSystemLibrary`?

syscall_cgo_linux.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/ebitengine/purego/internal/cgo"
 )
 
+// syscallXABI0 is the C fallback; unlike the assembly trampolines it always
+// writes errno back into the a3 field of syscallArgs.
```
TotallyGamerJet: I really don't think this is necessary. It's just asking to become out of date with a code change. better to let code describe behavior then comments which aren't compiled

sys_ppc64le.s
```diff
@@ -106,6 +106,8 @@ TEXT syscallX(SB), NOSPLIT, $0
 	// Store integer results back (R3, R4)
 	MOVD R3, syscallArgs_a1(R11)
 	MOVD R4, syscallArgs_a2(R11)
+	MOVD $0, R3                  // no errno here: clear a3 so that SyscallN
+	MOVD R3, syscallArgs_a3(R11) // does not echo the caller's third argument
```
TotallyGamerJet: Just do the same comment as the other platforms: `// no errno here: clear the input argument` no need for both here

sys_arm.s
```diff
@@ -135,6 +135,8 @@ skipfpload:
 	// save R0, R1
 	MOVW R0, syscallArgs_a1(R8)
 	MOVW R1, syscallArgs_a2(R8)
+	MOVW $0, R1                 // no errno here: clear a3 so that SyscallN
+	MOVW R1, syscallArgs_a3(R8) // does not echo the caller's third argument
```
TotallyGamerJet: ditto. Use `// no errno here: clear the input argument`

syscall_test.go
kumagi: Good catch - it uses getSystemLibrary now, so the local name and the comment explaining libc.so.6 are both gone.

syscall_cgo_linux.go
kumagi: Removed. The behavior is internal/cgo's syscall15 assigning args->a3 = errno, and the contract the callers rely on is documented on syscallArgs; a comment on the variable could only drift from both.

sys_ppc64le.s
kumagi: Done - one line, same as the others; the extra instruction needs no extra prose.

sys_arm.s
kumagi: Done as well; all eight trampolines now carry the identical one-line comment.

kumagi:
Addressed the four new comments; no behavior changes:

- `TestErrno` now uses `getSystemLibrary` instead of naming `/usr/lib/libSystem.B.dylib` and `libc.so.6` itself, so the platform library name lives in one place again.
- The comment above `syscallXABI0` is gone. The behavior it described is `internal/cgo`'s `syscall15`, and the `a3` contract is documented on `syscallArgs`, so there was nothing left for it to add that could not drift.
- The `arm` and `ppc64le` clearings use the same one-line comment as the other trampolines; all eight now read "no errno here: clear the input argument".

Verified: `gofmt -s` and `go vet ./...` are clean, the suite passes natively with and without Cgo and with `-race`, the package and its tests still build for every GOOS/GOARCH they support, and `TestErrno` still reports ENOENT when linux/amd64 is routed through the `internal/cgo` fallback.

syscall.go
```diff
@@ -18,6 +18,12 @@ const (
 // syscallArgs is the argument block handed to the syscall15 trampoline. On
 // platforms that go through internal/cgo it is passed to C as a
 // struct syscallArgs, so its layout must match the C ABI.
+//
+// The trampoline reports its results through this block: the return values go
+// to a1 and a2 and, where the platform captures errno, the error code goes to
+// a3. A trampoline that cannot capture errno must clear a3, which still holds
+// the caller's third argument, so that SyscallN never returns an input
+// argument as an error.
```
TotallyGamerJet, line 26:
```suggestion
// The trampoline reports its results through this block: the return values go
// to a1 and a2 and, where the platform captures errno, the error code goes to
// a3. A trampoline that cannot capture errno must clear a3.
```

syscall.go
```diff
@@ -49,7 +55,11 @@ func syscall_SyscallN(fn uintptr, sysargs []uintptr, floats []uintptr, r8 uintpt
 
 // SyscallN takes fn, a C function pointer and a list of arguments as uintptr.
 // There is an internal maximum number of arguments that SyscallN can take. It panics
-// when the maximum is exceeded. It returns the result and the libc error code if there is one.
+// when the maximum is exceeded. It returns the result and, if the platform captures
+// errno, the error code err: on darwin, and on the Linux architectures that go
+// through the C fallback, err is the libc errno, while on Windows err is what
+// syscall.SyscallN returns as its error, which is a GetLastError-style code rather
+// than a libc errno. Everywhere else errno is not captured so err is always 0.
```
TotallyGamerJet, line 62:
```suggestion
// when the maximum is exceeded. It returns the result and GetLastError on Windows or the errno on 
// platforms that capture it otherwise 0.
```

syscall_32bit.go
```diff
@@ -18,6 +18,12 @@ const (
 // syscallArgs is the argument block handed to the syscall15 trampoline. On
 // platforms that go through internal/cgo it is passed to C as a
 // struct syscallArgs, so its layout must match the C ABI.
+//
+// The trampoline reports its results through this block: the return values go
```
TotallyGamerJet, line 22: ditto for the above

syscall_32bit.go
```diff
@@ -51,7 +57,10 @@ func syscall_SyscallN(fn uintptr, sysargs []uintptr, floats []uintptr, r8 uintpt
 
 // SyscallN takes fn, a C function pointer and a list of arguments as uintptr.
 // There is an internal maximum number of arguments that SyscallN can take. It panics
-// when the maximum is exceeded. It returns the result and the libc error code if there is one.
+// when the maximum is exceeded. It returns the result and, on Windows, the error
```
TotallyGamerJet, line 60: ditto for the above

syscall_test.go
```diff
@@ -25,24 +25,62 @@ func TestOS(t *testing.T) {
 	}
 }
 
+// errnoIsCaptured reports whether SyscallN returns the libc errno as its third
+// value on this platform. The darwin trampolines save errno into the args
+// block and so does the C fallback in internal/cgo, which the Linux
+// architectures that have no assembly trampoline have to use. The other
+// trampolines cannot, and clear the field instead, so SyscallN returns 0 there.
```
TotallyGamerJet, line 32: Let's not include the current behavior since that's implied by which platforms return true in this function
```suggestion
// errnoIsCaptured reports whether SyscallN returns the libc errno as its third
// value on this platform.
```

syscall_test.go
```diff
@@ -25,24 +25,62 @@ func TestOS(t *testing.T) {
...
+	if !errnoIsCaptured() {
 		t.Skip("platform does not support returning errno from syscall")
 	}
 
-	libc, err := load.OpenLibrary("/usr/lib/libSystem.B.dylib")
+	// setErrno is called without arguments on purpose. SyscallN mirrors every
+	// integer argument into the float slots and the C fallback in internal/cgo,
+	// which is used on the Linux architectures that have no assembly trampoline,
+	// asserts that those slots are zero. Calling open here would therefore abort
+	// on those architectures before the errno check could run.
```
TotallyGamerJet, line 55: This comment can be removed. It's better explained in a PR comment since it's explaining why moving away from `open` is necessary

syscall.go
kumagi: Applied - the paragraph now stops at "A trampoline that cannot capture errno must clear a3."

syscall.go
kumagi: Applied, with a comma before "otherwise 0" so the sentence parses and without the trailing space from the suggestion.

syscall_32bit.go
kumagi: Done - syscall_32bit.go now carries the same three-line contract as syscall.go.

syscall_32bit.go
kumagi: Done - it now has the same two lines as syscall.go. syscall_ppc64le.go keeps its own sentence since that file only ever builds for linux/ppc64le, where neither Windows GetLastError nor a captured errno is in play.

syscall_test.go
kumagi: Done - the comment is just the two suggested lines; the switch itself shows which platforms capture errno.

syscall_test.go
kumagi: Removed, and the same explanation in testdata/liberrnotest/errno_test.c went with it so only the description of what setErrno does is left. The rationale for the helper: SyscallN mirrors its integer arguments into the float slots and internal/cgo's syscall15 asserts those are zero, so a call with arguments aborts there on the Linux architectures without an assembly trampoline.

kumagi:
Addressed the last round of comments; they were all doc polish, so the behavior is unchanged: the trampolines that cannot capture errno still clear `a3`, `SyscallN` still returns `a1`, `a2`, `a3` unconditionally, and `internal/cgo`'s fallback still reports the libc errno.

- `syscallArgs` now states the contract in three lines: results go to `a1`/`a2`, the error code goes to `a3` where the platform captures errno, and a trampoline that cannot capture errno must clear `a3`.
- `SyscallN`'s doc says it returns the result and GetLastError on Windows or the errno on platforms that capture it, otherwise 0. `syscall_32bit.go` uses the identical wording; `syscall_ppc64le.go` keeps its own sentence because that file only builds for linux/ppc64le, where neither GetLastError nor a captured errno is in play.
- `errnoIsCaptured` is documented by its name and its switch alone.
- The comment explaining why `TestErrno` uses a no-argument helper instead of `open` is gone, from both the test and `testdata/liberrnotest/errno_test.c`. For the record, the reason is that `SyscallN` mirrors every integer argument into the float slots while `internal/cgo`'s `syscall15` asserts those slots are zero, so a call with arguments aborts there on the Linux architectures that have no assembly trampoline. That assertion is pre-existing and untouched by this change.

Verified: `gofmt -s` and `go vet ./...` are clean with and without Cgo, the suite passes natively, the package assembles and builds for every GOOS/GOARCH it supports (darwin included, where the `GOOS_darwin` errno path is the one assembled), the test binaries compile for all of them, and calling the same helper through `internal/cgo`'s `syscall15` on linux/amd64 still yields -1 with errno ENOENT, so the fallback coverage is intact.

TotallyGamerJet (approved):
LGTM

hajimehoshi (approved):
LGTM, thanks!

PR #511, opened 2026-09-05
purego: only expand []any for declared variadic signatures

### LINKED ISSUE #506: RegisterFunc expands []any by value instead of by declared variadic signature
### PureGo Version

main @ 63002e0 (after #504)

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [x] FreeBSD
- [x] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64

`RegisterFunc`'s call closure in `func.go:345` decides `[]any` expansion purely by value, without checking whether the declared signature is actually variadic:

```go
for i, v := range args {
    if variadic, ok := reflect.TypeAssert[[]any](args[i]); ok {
        if i != len(args)-1 {
            panic("purego: can only expand last parameter")
        }
        for _, x := range variadic {
            keepAlive = addValue(reflect.ValueOf(x), keepAlive, addInt, addFloat, addStack, &numInts, &numFloats, &numStack)
        }
        continue
    }
    ...
```

Consequences:

1. A **non-variadic** signature such as `func(a int, b []any)` has its slice argument silently splatted into separate C arguments instead of being passed as a single slice pointer.
2. A signature such as `func(a []any, b int)` **panics** with `purego: can only expand last parameter` even though the caller passed a perfectly legitimate slice value.

Repro — C helper (`probe.c`, built with `cc -shared -fPIC -o probe.so probe.c`):

```c
#include <stdint.h>
long probe3(int a, long b, long c) { return (long)a * 1000000L + b * 1000L + c; }
```

```go
// Case 1: slice wrongly expanded. C receives (1, 10, 20) -> 1010020,
// instead of (1, slicePointer).
var probe func(a int, b []any) int64
purego.RegisterLibFunc(&probe, lib, "probe3")
got := probe(1, []any{10, 20})
fmt.Println(got) // prints 1010020 with the bug

// Case 2: legitimate call panics.
var probe2 func(a []any, b int) int64
purego.RegisterLibFunc(&probe2, lib, "probe3")
_ = probe2([]any{10, 20}, 1) // panics: purego: can only expand last parameter
```

Actual output on Linux/amd64: case 1 prints `1010020` (i.e. the slice was splatted); case 2 panics.

Expansion should only happen when the *declared* last parameter is variadic (`...any`), i.e. something like:

```go
if ty.IsVariadic() && i == ty.NumIn()-1 {
    if variadic, ok := reflect.TypeAssert[[]any](args[i]); ok { ... }
}
```

A non-variadic `[]any` parameter should travel through `addValue` as a single slice pointer, and `func(a []any, b int)` must not panic.

### What happens instead?

Slices are splatted or rejected based on their position and dynamic type rather than the declared signature, so `[]any` cannot be used as a regular (non-variadic) parameter at all.

Minimized reproducible case: the `probe3` C helper plus the two Go snippets above (only a C compiler and `libc` needed; no other deps). Both verified failing on `main @ 63002e0`, Linux/amd64, Go 1.26.7 — case 1 returns `1010020`, case 2 panics with `purego: can only expand last parameter`.

Note the doc comment on `RegisterFunc` (`func.go:76`) describes the special case as "the last argument of fptr is a variadic interface", which matches the proposed fix (check the declared type, not the value).

kumagi:
#506

bug

## What this PR does | solves
The call closure splatted any []any-typed value appearing in the arguments, without consulting the declared signature. Non-variadic functions taking a []any parameter received splatted C arguments (or panicked with 'can only expand last parameter' when the slice was not last). Gate the expansion on ty.IsVariadic() and the final position so a []any elsewhere travels as a single slice pointer.

Closes #506

### DIFF
--- a/callback_test.go
+++ b/callback_test.go
@@ -411,3 +411,51 @@ func TestCallbackFloat32StackPacking(t *testing.T) {
 		t.Errorf("callCallback12Float32() = %d, want %d", got, want)
 	}
 }
+
+func TestRegisterFuncExpandFinalAnyArgument(t *testing.T) {
+	// The last argument is expanded into the C arguments held in it,
+	// whether it is declared as ...any or as []any.
+	// https://github.com/ebitengine/purego/issues/506
+	cb := purego.NewCallback(func(a, b, c uintptr) uintptr {
+		return a*100 + b*10 + c
+	})
+	const want = uintptr(123)
+	t.Run("Variadic", func(t *testing.T) {
+		var fn func(a uintptr, args ...any) uintptr
+		purego.RegisterFunc(&fn, cb)
+		if got := fn(1, uintptr(2), uintptr(3)); got != want {
+			t.Errorf("func(uintptr, ...any) = %d, want %d", got, want)
+		}
+	})
+	t.Run("Slice", func(t *testing.T) {
+		var fn func(a uintptr, args []any) uintptr
+		purego.RegisterFunc(&fn, cb)
+		if got := fn(1, []any{uintptr(2), uintptr(3)}); got != want {
+			t.Errorf("func(uintptr, []any) = %d, want %d", got, want)
+		}
+	})
+}
+
+func TestRegisterFuncKeepNonFinalAnyArgument(t *testing.T) {
+	// A []any that isn't the last argument is a regular argument:
+	// it is passed as one value and must not panic.
+	// https://github.com/ebitengine/purego/issues/506
+	var first any
+	var second uintptr
+	cb := purego.NewCallback(func(args *any, b uintptr) uintptr {
+		first = *args
+		second = b
+		return 0
+	})
+	var fn func(args []any, b uintptr) uintptr
+	purego.RegisterFunc(&fn, cb)
+	s := []any{uintptr(2), uintptr(3)}
+	fn(s, 7)
+	runtime.KeepAlive(s)
+	if want := any(uintptr(2)); first != want {
+		t.Errorf("the first C argument = %v, want %v", first, want)
+	}
+	if second != 7 {
+		t.Errorf("the second C argument = %d, want 7", second)
+	}
+}
--- a/callback_test.go
+++ b/callback_test.go
@@ -414,8 +414,7 @@ func TestCallbackFloat32StackPacking(t *testing.T) {
 
 func TestRegisterFuncExpandFinalAnyArgument(t *testing.T) {
 	// The last argument is expanded into the C arguments held in it,
-	// whether it is declared as ...any or as []any.
-	// https://github.com/ebitengine/purego/issues/506
+	// whether it is declared as ...any or as []any (#506).
 	cb := purego.NewCallback(func(a, b, c uintptr) uintptr {
 		return a*100 + b*10 + c
 	})
@@ -438,8 +437,7 @@ func TestRegisterFuncExpandFinalAnyArgument(t *testing.T) {
 
 func TestRegisterFuncKeepNonFinalAnyArgument(t *testing.T) {
 	// A []any that isn't the last argument is a regular argument:
-	// it is passed as one value and must not panic.
-	// https://github.com/ebitengine/purego/issues/506
+	// it is passed as one value and must not panic (#506).
 	var first any
 	var second uintptr
 	cb := purego.NewCallback(func(args *any, b uintptr) uintptr {
--- a/func.go
+++ b/func.go
@@ -342,14 +342,17 @@ func RegisterFunc(fptr any, cfn uintptr) {
 			}
 		}
 		for i, v := range args {
-			if variadic, ok := reflect.TypeAssert[[]any](args[i]); ok {
-				if i != len(args)-1 {
-					panic("purego: can only expand last parameter")
+			// Only splat a []any value when the declared signature is
+			// variadic and this is its final parameter. A []any in any
+			// other position is an ordinary single argument holding a
+			// slice, and must go through addValue as one pointer.
+			if ty.IsVariadic() && i == len(args)-1 {
+				if variadic, ok := reflect.TypeAssert[[]any](args[i]); ok {
+					for _, x := range variadic {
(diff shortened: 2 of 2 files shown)


hajimehoshi:
The new `ty.IsVariadic()` condition breaks a documented use of `RegisterFunc`: its godoc explicitly includes a final `[]interface{}` parameter in the argument-expansion special case, alongside `...any`.

For example, an existing binding declared as `func(uintptr, []any) uintptr` and called with `(1, []any{uintptr(2), uintptr(3)})` currently passes three C arguments. With this patch, it passes the slice pointer as the second argument and leaves the third argument unset.

I verified this on macOS/arm64 using a `NewCallback` callback that returns `a*100 + b*10 + c`: the base revision returns `123`, while the patched version returns an incorrect pointer-derived value. Separate checks confirmed that `...any` expansion still works and the non-final slice case is fixed.

Could this preserve the documented final-slice expansion behavior while fixing non-final slice arguments? If changing that contract is intentional, the compatibility implications and documentation need to be addressed explicitly. Regression tests for these cases would also help.

Authored by Codex (OpenAI), on behalf of @hajimehoshi.

kumagi:
Thanks for the careful check — you are right, the `ty.IsVariadic()` condition dropped the documented `[]interface{}` case.

I changed the gate so that it depends on the position of the argument only: the last argument is expanded when it holds a `[]any` (which covers both a declared `...any` parameter and the documented final `[]interface{}` parameter), and a `[]any` in any other position is now passed through `addValue` as a single value instead of panicking with `purego: can only expand last parameter`. Your binding `func(uintptr, []any) uintptr` called with `(1, []any{uintptr(2), uintptr(3)})` expands again and returns `123`, while `func([]any, uintptr) uintptr` no longer panics and hands the slice to C as one argument.

For clarity: case 1 of #506 (a non-variadic *final* `[]any` being expanded) is intentionally kept, because that is the special case the `RegisterFunc` godoc has always described and existing bindings rely on it. The doc comment now states explicitly that only the last argument is expanded and that a `[]interface{}` elsewhere is an ordinary argument. Regression tests were added in callback_test.go: TestRegisterFuncExpandFinalAnyArgument (both `...any` and a final `[]any`) and TestRegisterFuncKeepNonFinalAnyArgument (a non-final `[]any` arrives as a single value and does not panic). The first test fails on the previous revision and the second one panics on `main`, so both behaviors are pinned.

`go test ./...` passes on linux/amd64, and the package still cross-compiles for darwin, ios, windows and freebsd. Note that the PR title ("only expand []any for declared variadic signatures") no longer describes the behavior; "only expand the last []any argument" would be accurate.

func.go
```diff
@@ -73,10 +73,12 @@ func RegisterLibFunc(fptr any, handle uintptr, name string) {
 //	unsafe.Pointer, *T <=> void*
 //	[]T => void*
 //
-// There is a special case when the last argument of fptr is a variadic interface (or []interface}
+// There is a special case when the last argument of fptr is a variadic interface (or []interface{}):
```
hajimehoshi: `[]any`?

func.go
```diff
@@ -73,10 +73,12 @@ func RegisterLibFunc(fptr any, handle uintptr, name string) {
...
 // it will be expanded into a call to the C function as if it had the arguments in that slice.
 // This means that using arg ...any is like a cast to the function with the arguments inside arg.
 // This is not the same as C variadic.
+// Only the last argument is expanded this way. A []interface{} in any other position is an ordinary
```
hajimehoshi: `[]any`

func.go
kumagi: Fixed: this now says `[]any`, which is what the rest of the package uses.

func.go
kumagi: Fixed here as well: the sentence now uses `[]any`.

kumagi:
Switched both mentions in the `RegisterFunc` doc comment to `[]any` so the wording matches the rest of the package, which uses the `any` alias everywhere. No behavior change; `gofmt`, `go vet ./...` and `go test ./...` still pass on linux/amd64, and the package still cross-compiles for darwin and windows.

callback_test.go
```diff
@@ -411,3 +411,51 @@ func TestCallbackFloat32StackPacking(t *testing.T) {
 		t.Errorf("callCallback12Float32() = %d, want %d", got, want)
 	}
 }
+
+func TestRegisterFuncExpandFinalAnyArgument(t *testing.T) {
+	// The last argument is expanded into the C arguments held in it,
+	// whether it is declared as ...any or as []any.
+	// https://github.com/ebitengine/purego/issues/506
```
hajimehoshi: Just #506 is fine.

```
	// The last argument is expanded into the C arguments held in it,
	// whether it is declared as ...any or as []any (#506).
```

callback_test.go
```diff
@@ -411,3 +411,51 @@ func TestCallbackFloat32StackPacking(t *testing.T) {
...
+		if got := fn(1, []any{uintptr(2), uintptr(3)}); got != want {
+			t.Errorf("func(uintptr, []any) = %d, want %d", got, want)
+		}
+	})
+}
+
+func TestRegisterFuncKeepNonFinalAnyArgument(t *testing.T) {
+	// A []any that isn't the last argument is a regular argument:
+	// it is passed as one value and must not panic.
+	// https://github.com/ebitengine/purego/issues/506
```
hajimehoshi: ditto

callback_test.go
kumagi: Done — the comment now reads `whether it is declared as ...any or as []any (#506).`

callback_test.go
kumagi: Done — the same here: `it is passed as one value and must not panic (#506).`

kumagi:
Both test comments now reference the issue as `(#506)` instead of the full URL. Comment-only change; `gofmt`, `go vet ./...` and `go test ./...` still pass on linux/amd64.

hajimehoshi (approved):
LGTM

@TotallyGamerJet PTAL

TotallyGamerJet (approved):
LGTM

PR #512, opened 2026-09-05
purego: validate callback inputs before touching reflect

### LINKED ISSUE #507: NewCallback/RegisterFunc panic with reflect internals (or nil deref) on non-function input
### PureGo Version

main @ 63002e0 (after #504)

### Operating System

- [x] Windows
- [x] macOS
- [x] Linux
- [x] FreeBSD
- [x] NetBSD
- [x] Android
- [x] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64

`NewCallback` (`syscall_unix.go:28`) dereferences the type before validating it:

```go
func NewCallback(fn any) uintptr {
	ty := reflect.TypeOf(fn)
	for i := range ty.NumIn() {   // panics here for non-func / nil
```

while the proper `Kind() != reflect.Func` check lives one call deeper in `compileCallback`, which is never reached. Likewise `RegisterFunc` (`func.go:131`) calls `.Elem()` before checking:

```go
fn := reflect.ValueOf(fptr).Elem()   // panics here for non-pointers
ty := fn.Type()
if ty.Kind() != reflect.Func {
    panic("purego: fptr must be a function pointer")
}
```

Repro (pure Go, any platform):

```go
_ = purego.NewCallback(123) // want: "purego: ..." message
_ = purego.NewCallback(nil) // want: "purego: ..." message
purego.RegisterFunc(123, 1) // want: "purego: fptr must be a function pointer"
```

Actual panics observed on `main @ 63002e0`:

- `NewCallback(123)` → `reflect: NumIn of non-func type int`
- `NewCallback(nil)` → `runtime error: invalid memory address or nil pointer dereference`
- `RegisterFunc(123, 1)` → `reflect: call of reflect.Value.Elem on int Value`

None of them is the intended `purego: ...` message, and the nil case is a raw nil-dereference.

Misuse should fail with the library's own actionable errors, e.g. `purego: the type must be a function ...` / `purego: fptr must be a (non-nil) function pointer`, instead of leaking `reflect`/runtime internals (or a nil dereference) that users cannot attribute to their mistake.

### What happens instead?

Users passing a non-function (a very common mistake when wiring callbacks) get an internal-looking panic with no hint that purego rejected their argument, and `nil` crashes with a bare nil-pointer dereference.

Minimized reproducible case: the three one-liners above — no C code, no shared library, runnable on every GOOS/GOARCH. Verified on Linux/amd64, Go 1.26.7, `main @ 63002e0`.

Suggested fix:

```go
func NewCallback(fn any) uintptr {
	ty := reflect.TypeOf(fn)
	if ty == nil || ty.Kind() != reflect.Func {
		panic("purego: the type must be a function but was not")
	}
	...
```

```go
rv := reflect.ValueOf(fptr)
if rv.Kind() != reflect.Ptr || rv.IsNil() {
	panic("purego: fptr must be a non-nil function pointer")
}
fn := rv.Elem()
```

kumagi:
#507

bug

## What this PR does | solves
NewCallback (unix and windows) called ty.NumIn() before checking the kind, so non-function input died inside reflect (or with a bare nil dereference) instead of the intended purego error. RegisterFunc called Value.Elem() before checking for a function pointer with the same effect. Validate first so misuse fails with actionable messages.

Closes #507

### DIFF
--- a/func.go
+++ b/func.go
@@ -128,7 +128,11 @@ func RegisterLibFunc(fptr any, handle uintptr, name string) {
 // [Cgo rules]: https://pkg.go.dev/cmd/cgo#hdr-Go_references_to_C
 func RegisterFunc(fptr any, cfn uintptr) {
 	const is32bit = unsafe.Sizeof(uintptr(0)) == 4
-	fn := reflect.ValueOf(fptr).Elem()
+	rv := reflect.ValueOf(fptr)
+	if rv.Kind() != reflect.Ptr || rv.IsNil() {
+		panic("purego: fptr must be a non-nil function pointer")
+	}
+	fn := rv.Elem()
 	ty := fn.Type()
 	if ty.Kind() != reflect.Func {
 		panic("purego: fptr must be a function pointer")
--- a/func_test.go
+++ b/func_test.go
@@ -181,6 +181,60 @@ func TestRegisterLibFunc_Bool(t *testing.T) {
 	}
 }
 
+func TestNewCallback_NotAFunction(t *testing.T) {
+	for _, tc := range []struct {
+		name string
+		fn   any
+	}{
+		{"nil", nil},
+		{"int", 42},
+		{"string", "not a function"},
+		{"pointer", new(int)},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			defer func() {
+				r := recover()
+				if r == nil {
+					t.Fatal("NewCallback did not panic")
+				}
+				const want = "purego: the type must be a function but was not"
+				if got := fmt.Sprint(r); got != want {
+					t.Fatalf("panic mismatch:\n  got:  %q\n  want: %q", got, want)
+				}
+			}()
+			purego.NewCallback(tc.fn)
+		})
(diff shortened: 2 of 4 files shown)


hajimehoshi (approved):
LGTM

hajimehoshi:
@TotallyGamerJet PTAL

TotallyGamerJet:
Add a test case?

kumagi:
Done — I added TestNewCallback_NotAFunction and TestRegisterFunc_InvalidFunctionPointer in func_test.go. They check that NewCallback panics with "purego: the type must be a function but was not" for nil and non-function inputs, and that RegisterFunc panics with "purego: fptr must be a non-nil function pointer" for nil/non-pointer/nil-function-pointer inputs and with "purego: fptr must be a function pointer" for pointers to non-functions. The RegisterFunc cases run on all platforms and the NewCallback cases cover the unix and windows paths since both validate before any platform-specific code.

TotallyGamerJet (approved):
LGTM

PR #513, opened 2026-09-05
objc: anchor ivarRegex so malformed names are rejected

### LINKED ISSUE #508: objc ivarRegex is unanchored so RegisterClass accepts malformed ivar names
### PureGo Version

main @ 63002e0 (after #504)

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [x] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64 (predicate reproduced standalone; full path needs darwin)

`objc/objc_runtime_darwin.go:272` validates ivar names with an unanchored pattern used via `MatchString` (partial match) at `:312`:

```go
// ivarRegex checks to make sure the Ivar is correctly formatted
var ivarRegex = regexp.MustCompile("[a-z_][a-zA-Z0-9_]*")
...
if !ivarRegex.MatchString(ivar.Name) {
    return nil, fmt.Errorf(...)
}
```

`regexp.MatchString` reports a match for any *substring*, so every name containing at least one lowercase letter or underscore passes, contradicting the comment ("must start lowercase, no spaces"):

```go
ivarRegex.MatchString("Bar")     // true ("ar" matches) — must be rejected
ivarRegex.MatchString("foo bar") // true ("foo" matches) — must be rejected
ivarRegex.MatchString("1abc")    // true ("abc" matches) — must be rejected
ivarRegex.MatchString("foo-bar") // true ("foo" matches) — must be rejected
```

I verified this predicate standalone on Linux (same pattern + `MatchString` semantics, which are platform-independent). The consequence on darwin is that `objc.RegisterClass` accepts malformed ivar names such as:

```go
cls, err := objc.RegisterClass("MyClass", superclass, nil, []objc.FieldDef{
    {Name: "Bar", Type: reflect.TypeOf(float32(0)), Attr: ...},
}, nil)
// err == nil with the bug; should be a validation error
```

and the failure surfaces later inside `class_addIvar` / `GetIvar` / `SetIvar`, where it is much harder to diagnose.

Names violating the documented rule (must match `^[a-z_][a-zA-Z0-9_]*$`) are rejected by `RegisterClass` with a clear error.

### What happens instead?

Effectively no validation: any name containing a lowercase letter or underscore is accepted, so typos like `"Bar"` or `"foo bar"` slip through into the ObjC runtime.

Minimized reproducible case:

1. Pure-predicate (runs anywhere, verified on Linux/amd64, Go 1.26.7):

```go
package main

import (
	"fmt"
	"regexp"
)

var ivarRegex = regexp.MustCompile("[a-z_][a-zA-Z0-9_]*") // copy of objc_runtime_darwin.go:272

func main() {
	for _, s := range []string{"Bar", "foo bar", "1abc", "foo-bar"} {
		fmt.Printf("MatchString(%q) = %v\n", s, ivarRegex.MatchString(s))
	}
	// All print true; all should print false.
}
```

2. Full path (darwin only): the `objc.RegisterClass` snippet above with `Name: "Bar"` returns `nil` error with the bug.

Suggested fix (one line):

```go
var ivarRegex = regexp.MustCompile(`^[a-z_][a-zA-Z0-9_]*$`)
```

kumagi:
#508

bug

## What this PR does | solves
MatchString does substring matching, so the unanchored pattern let names like Bar or 'foo bar' through RegisterClass validation. Anchor it for a full-string match.

Closes #508

### DIFF
--- a/objc/objc_runtime_darwin.go
+++ b/objc/objc_runtime_darwin.go
@@ -269,7 +269,7 @@ type FieldDef struct {
 }
 
 // ivarRegex checks to make sure the Ivar is correctly formatted
-var ivarRegex = regexp.MustCompile("[a-z_][a-zA-Z0-9_]*")
+var ivarRegex = regexp.MustCompile(`^[a-z_][a-zA-Z0-9_]*$`)
 
 // RegisterClass takes the name of the class to create, the superclass, a list of protocols this class
 // implements, a list of fields this class has and a list of methods. It returns the created class or an error


hajimehoshi (approved):
LGTM

@TotallyGamerJet PTAL

TotallyGamerJet (approved):
LGTM

PR #514, opened 2026-09-05
objc: start SendSuper lookup at the superclass

### LINKED ISSUE #509: objc SendSuper passes the receiver's own class instead of its superclass
### PureGo Version

main @ 63002e0 (after #504)

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [x] iOS

### Go Version (`go version`)

go1.26.7 (code-verified; runtime repro below needs darwin — this box is Linux, see note)

Both `SendSuper` helpers in `objc/objc_runtime_darwin.go` fill `objc_super.superClass` with the receiver's **own** class instead of its superclass:

```go
// :186-189  (ID.SendSuper)
super := &objc_super{
    receiver:   id,
    superClass: id.Class(),   // == object_getClass(id), i.e. the class itself (:134-135)
}
return objc_msgSendSuper2(super, sel, args...)

// :199-202  (SendSuper[T]) — identical mistake
super := &objc_super{
    receiver:   id,
    superClass: id.Class(),
}
```

Per Apple's runtime contract, `objc_super.super_class` is the class at which method lookup **starts** — for a super-send that must be the superclass. Starting at the receiver's own class makes `objc_msgSendSuper2` behave exactly like a plain `objc_msgSend`.

Repro (darwin):

```go
// 1. Register a subclass of NSObject overriding "description" (or any method)
//    via objc.RegisterClass with a MethodDef.
// 2. Inside the override implementation, call id.SendSuper(objc.RegisterName("description")).
// 3. Call the method on an instance.
```

With the bug, step 2 dispatches back into the same override → unbounded recursion → stack overflow. With correct behavior, the superclass implementation runs exactly once. The generic `objc.SendSuper[T]` variant behaves the same way (same wrong field), including through the `_stret` path.

Note on verification: this environment is Linux-only so I could not execute the darwin path; the finding is established by code inspection — `ID.Class()` is defined as `object_getClass(id)` (`objc_runtime_darwin.go:134-135`), which by definition is never the superclass — plus the resulting equivalence with `Send` when the method is overridden. Reviewing maintainers on macOS should be able to confirm the infinite recursion in seconds with the steps above.

`SendSuper` starts lookup at the superclass, so calling it from an override invokes the parent implementation once and returns.

### What happens instead?

Lookup starts at the receiver's own class: `SendSuper` is an alias for `Send`, and the canonical use (calling it from an override) recurses forever.

Minimized reproducible case: the three darwin steps above (subclass + override + `SendSuper` inside the override). No third-party dependencies; only the `objc` package and `NSObject`.

Suggested fix — the package already imports exactly the right primitives (`object_getClass` and `class_getSuperclass`, `:95` / `:103`) and even exposes `Class.SuperClass()` (`:540-543`):

```go
super := &objc_super{
    receiver:   id,
    superClass: id.Class().SuperClass(),
}
```

applied to both `ID.SendSuper` and `SendSuper[T]`.

kumagi:
#509

bug

## What this PR does | solves
objc_super.superClass is where method lookup begins, so handing it the receiver's own class made SendSuper an alias for Send and recursed forever when called from an override. Use the already-available Class.SuperClass() helper in both SendSuper variants.

Closes #509


hajimehoshi:
This change skips the immediate superclass in both `SendSuper` helpers. The package calls `objc_msgSendSuper2`, whose contract differs from `objc_msgSendSuper`: [Apple's runtime header](https://github.com/apple-oss-distributions/objc4/blob/main/runtime/objc-abi.h) explicitly says that `objc_msgSendSuper2()` takes the current search class, not its superclass. The runtime advances to the superclass itself, so passing `id.Class().SuperClass()` starts lookup at the grandparent.

I verified this on macOS arm64 using the existing examples. At the PR base (`63002e0`), both `ExampleID_SendSuper` and `ExampleSendSuper` pass. With the PR's two changes applied through a Go overlay, running each example separately aborts with an unrecognized selector for `-[ChildObject doSomething]` and `-[ChildObject2 doSomething]`, respectively: the parent implementation is skipped and lookup reaches `NSObject`.

Please keep `id.Class()` in both helpers. The runtime contract cited in #509 does not apply to the `objc_msgSendSuper2` entry point used here.

Posted by Codex (OpenAI), on behalf of @hajimehoshi.

kumagi:
Good catch, and you are right. I checked the two entry points again: objc_msgSendSuper2 takes the current search class in objc_super.superClass and the runtime advances to its superclass by itself, unlike objc_msgSendSuper, which takes the superclass to message. Passing id.Class().SuperClass() therefore skipped the immediate superclass and let the lookup fall through to NSObject, which matches the unrecognized selector you saw for -[ChildObject doSomething] and -[ChildObject2 doSomething].

I reverted both SendSuper helpers to id.Class(), restoring the behavior the examples expect, and reworded the objc_super doc comment so the objc_msgSendSuper vs objc_msgSendSuper2 distinction is written down and this doesn't get "fixed" again. The diff is now documentation only, so #509's premise does not apply to the code path used here and closing this PR (and the issue) works for me.

hajimehoshi (approved):
LGTM

@TotallyGamerJet PTAL

objc/objc_runtime_darwin.go
```diff
@@ -172,8 +172,11 @@ func Send[T any](id ID, sel SEL, args ...any) T {
 }
 
 // objc_super data structure is generated by the Objective-C compiler when it encounters the super keyword
-// as the receiver of a message. It specifies the class definition of the particular superclass that should
-// be messaged.
+// as the receiver of a message.
+//
+// SendSuper calls objc_msgSendSuper2 instead of objc_msgSendSuper. Unlike objc_msgSendSuper, which expects
```
hajimehoshi: Add `[]` to `SendSuper`.

```go
// [SendSuper] calls ...
```

kumagi: Done — the comment now reads `[SendSuper] calls objc_msgSendSuper2 instead of objc_msgSendSuper.`, so it links to the helper.

kumagi:
Thanks for the review. I also made SendSuper a doc link in the objc_super comment as requested, so the reference to the helper is clickable. The change is still documentation only; the helpers keep passing id.Class().

TotallyGamerJet (approved):
LGTM

PR #515, opened 2026-09-05
purego: respect field offsets in amd64 struct packing

### LINKED ISSUE #520: amd64 struct arguments ignore field offsets, corrupting values and shifting later args
### PureGo Version

main @ 763bb99

### Operating System

- [x] Windows
- [x] macOS
- [x] Linux
- [x] FreeBSD
- [x] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64

On amd64, `tryPlaceRegister` in `struct_amd64.go` packs struct fields back-to-back into a 64-bit accumulator without looking at each field's in-memory offset:

1. 64-bit kinds (`Pointer`, `Int64`/`Int`, `Uint64`/`Uint`/`Uintptr`, `Float64`) do `val = ...` instead of `val |= ...`, so a 64-bit field following a smaller field **overwrites** it. The first eightbyte is dropped and all later arguments shift by one slot.
2. Small fields (`Int8/16/32`, `Uint8/16/32`, `Bool`, `Float32`) ignore padding, so e.g. the `int32` at offset 4 in `{int8; int32}` is placed at bit 8 instead of bit 32.
3. The recursive `place()` clobbers the outer eightbyte cursor with the inner tail position, so later siblings of nested structs are misclassified (e.g. `StructInStruct{A, B, C inner}` loses B and C).

Repro — C helper (`pack.c`, built with `cc -shared -fPIC -o pack.so pack.c`):

```c
#include <stdint.h>
struct S1 { int8_t a; int64_t b; };   // B at offset 8, size 16
struct S2 { int8_t a; int32_t b; };   // B at offset 4, size 8
struct S1 ident_S1(struct S1 s) { return s; }
struct S2 ident_S2(struct S2 s) { return s; }
int64_t trailing_S1(struct S1 s, int64_t extra) { (void)s; return extra; }
```

```go
type S1np struct {
    _ structs.HostLayout
    A int8
    B int64
}
type S2np struct {
    _ structs.HostLayout
    A int8
    B int32
}

var identS1 func(S1np) S1np
purego.RegisterLibFunc(&identS1, lib, "ident_S1")
got1 := identS1(S1np{A: 0x11, B: 0x2222222222222222})
// got1 = {A:0x22 B:0x0}, want {A:0x11 B:0x2222222222222222}

var identS2 func(S2np) S2np
purego.RegisterLibFunc(&identS2, lib, "ident_S2")
got2 := identS2(S2np{A: 0x11, B: 0x22222222})
// got2 = {A:0x11 B:0x22}, want {A:0x11 B:0x22222222}

var trailing func(S1np, int64) int64
purego.RegisterLibFunc(&trailing, lib, "trailing_S1")
got3 := trailing(S1np{A: 0x11, B: 0x2222222222222222}, 0x3333333333333333)
// got3 = 0x0, want 0x3333333333333333 (later argument shifted by one slot)
```

The existing tests only pass because they use explicit padding fields (e.g. `_[7]int8`), which happen to flush the accumulator. Naturally-padded structs hit this every time.

Struct arguments are packed exactly as their in-memory image (per System V AMD64 ABI §3.2.3): `{int8; int64}` goes out as `eightbyte0 = {A + pad}`, `eightbyte1 = {B}`, and later arguments keep their slots. All three repro cases return the input unchanged.

### What happens instead?

Field values are corrupted (`A=0x22 B=0x0`, `B=0x22`) and arguments after the struct shift by one slot (`extra` reads as `0x0`). Depending on the callee this is silent data corruption or a crash.

Minimized reproducible case: the `pack.c` helper plus the three Go snippets above (only a C compiler needed, no other deps). All three verified failing on `main @ 763bb99`, Linux/amd64, Go 1.26.7.

Related PR: #515.

kumagi:
Closes #520

bug

## What this PR does | solves
tryPlaceRegister packed fields back-to-back and overwrote pending small fields when a 64-bit field followed (val = instead of |=), dropping the first eightbyte and shifting all later arguments by one slot. Small fields also ignored padding, misplacing e.g. the int32 at offset 4 in {int8; int32}. Track each field's in-memory offset: flush the pending eightbyte on crossing, place wide fields directly, and realign the bit cursor for small fields. The recursive place() also clobbered the outer eightbyte index with the inner tail position, misclassifying later sibling fields of nested structs (e.g. StructInStruct lost B and C); save and restore the cursor around recursion.

### DIFF
--- a/struct_amd64.go
+++ b/struct_amd64.go
@@ -211,8 +211,10 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 		shift = 0
 		class = _NO_CLASS
 	}
-	var place func(v reflect.Value)
-	place = func(v reflect.Value) {
+	var place func(v reflect.Value, base uintptr)
+	// curEight tracks the eightbyte index of the pending accumulator.
+	var curEight uintptr
+	place = func(v reflect.Value, base uintptr) {
 		var numFields int
 		if v.Kind() == reflect.Struct {
 			numFields = v.Type().NumField()
@@ -226,57 +228,110 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 			}
 			flushed = false
 			var f reflect.Value
+			var fieldOff uintptr
 			if v.Kind() == reflect.Struct {
 				f = v.Field(i)
+				fieldOff = base + v.Type().Field(i).Offset
 			} else {
 				f = v.Index(i)
+				fieldOff = base + uintptr(i)*f.Type().Size()
+			}
+			// The System V ABI classifies eightbytes from the in-memory
+			// image: a field starting in a later eightbyte than the pending
+			// accumulator ends it. Flush first so a wide field (e.g. the
+			// int64 in {int8; int64}) never overwrites accumulated smaller
+			// fields. A field wider than 4 bytes always starts a fresh
+			// eightbyte in practice (its alignment > remaining space), so
+			// after flushing we can place it directly without shifting.
+			needFresh := shift != 0 && fieldOff/8 != curEight
+			if needFresh {
+				flushIfNeeded()
+			}
+			curEight = fieldOff / 8
+			// Small fields accumulate at the in-memory offset within the
+			// current eightbyte. Realign the bit cursor when the field
+			// starts later (padding), e.g. the int32 at offset 4 in
+			// {int8; int32}.
+			alignTo := func(off uintptr) {
+				want := byte((off % 8) * 8)
+				if want > shift {
+					shift = want
+				}
 			}
 			switch f.Kind() {
 			case reflect.Struct:
-				place(f)
+				savedEight := curEight
+				place(f, fieldOff)
+				curEight = savedEight
 			case reflect.Bool:
+				alignTo(fieldOff)
 				if f.Bool() {
 					val |= 1 << shift
 				}
 				shift += 8
 				class |= _INTEGER
 			case reflect.Pointer, reflect.UnsafePointer:
-				val = uint64(f.Pointer())
-				shift = 64
+				if !needFresh {
+					val = uint64(f.Pointer())
+					shift = 64
+				} else {
+					flushed = false
+					addInt(uintptr(f.Pointer()))
+					flushed = true
+				}
 				class = _INTEGER
 			case reflect.Int8:
+				alignTo(fieldOff)
 				val |= uint64(f.Int()&0xFF) << shift
 				shift += 8
 				class |= _INTEGER
 			case reflect.Int16:
+				alignTo(fieldOff)
 				val |= uint64(f.Int()&0xFFFF) << shift
 				shift += 16
 				class |= _INTEGER
 			case reflect.Int32:
+				alignTo(fieldOff)
 				val |= uint64(f.Int()&0xFFFF_FFFF) << shift
 				shift += 32
 				class |= _INTEGER
 			case reflect.Int64, reflect.Int:
-				val = uint64(f.Int())
-				shift = 64
+				if !needFresh {
+					val = uint64(f.Int())
+					shift = 64
+				} else {
+					flushed = false
+					addInt(uintptr(f.Int()))
+					flushed = true
+				}
 				class = _INTEGER
 			case reflect.Uint8:
+				alignTo(fieldOff)
 				val |= f.Uint() << shift
 				shift += 8
 				class |= _INTEGER
 			case reflect.Uint16:
+				alignTo(fieldOff)
 				val |= f.Uint() << shift
 				shift += 16
 				class |= _INTEGER
 			case reflect.Uint32:
+				alignTo(fieldOff)
 				val |= f.Uint() << shift
 				shift += 32
 				class |= _INTEGER
 			case reflect.Uint64, reflect.Uint, reflect.Uintptr:
-				val = f.Uint()
-				shift = 64
+				if !needFresh {
+					val = f.Uint()
+					shift = 64
+				} else {
+					flushed = false
+					addInt(uintptr(f.Uint()))
+					flushed = true
+				}
 				class = _INTEGER
 			case reflect.Float32:
+				alignTo(fieldOff)
 				val |= uint64(math.Float32bits(float32(f.Float()))) << shift
 				shift += 32
 				class |= _SSE
@@ -285,11 +340,19 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 					ok = false
 					return
 				}
-				val = uint64(math.Float64bits(f.Float()))
-				shift = 64
+				if !needFresh {
+					val = uint64(math.Float64bits(f.Float()))
+					shift = 64
+				} else {
+					flushed = false
+					addFloat(uintptr(math.Float64bits(f.Float())))
+					flushed = true
+				}
 				class = _SSE
 			case reflect.Array:
-				place(f)
+				savedArrEight := curEight
+				place(f, fieldOff)
+				curEight = savedArrEight
 			default:
 				panic("purego: unsupported kind " + f.Kind().String())
 			}
@@ -304,7 +367,7 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 		}
 	}
 
-	place(v)
+	place(v, 0)
 	flushIfNeeded()
 	return ok
 }
--- a/struct_amd64.go
+++ b/struct_amd64.go
@@ -260,9 +260,12 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 			}
 			switch f.Kind() {
 			case reflect.Struct:
-				savedEight := curEight
+				// Recursion shares the accumulator, so it must also share
+				// curEight: when a nested struct spans eightbytes its tail
+				// is pending in the last one it touched, and the next
+				// sibling field belongs to that same eightbyte. Restoring
+				// the outer index would flush the tail prematurely.
 				place(f, fieldOff)
-				curEight = savedEight
 			case reflect.Bool:
 				alignTo(fieldOff)
 				if f.Bool() {
@@ -350,9 +353,7 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 				}
 				class = _SSE
 			case reflect.Array:
-				savedArrEight := curEight
 				place(f, fieldOff)
-				curEight = savedArrEight
 			default:
 				panic("purego: unsupported kind " + f.Kind().String())
 			}
--- a/struct_amd64.go
+++ b/struct_amd64.go
@@ -246,6 +246,12 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
 			needFresh := shift != 0 && fieldOff/8 != curEight
 			if needFresh {
 				flushIfNeeded()
+				// The intermediate flush above ended the pending eightbyte;
+				// the field placed below starts a fresh accumulator that
+				// must still reach the flush at the end of this iteration or
+				// at the end of place(). Leave the flag clear until a field
+				// is actually emitted.
+				flushed = false
 			}
 			curEight = fieldOff / 8
 			// Small fields accumulate at the in-memory offset within the
--- a/struct_test.go
+++ b/struct_test.go
@@ -944,6 +944,88 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 				}
 				runtime.KeepAlive(ptr)
 			}
+			t.Run("CharLong", func(t *testing.T) {
+				// Small field followed by a wide field crossing the eightbyte
+				// boundary: the wide field must not overwrite the pending
+				// small fields.
+				type CharLong struct {
+					_ structs.HostLayout
+					A int8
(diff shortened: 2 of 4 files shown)


hajimehoshi:
If this is a bug fix, file an issue first

kumagi:
Done — I've filed #520 for this bug and updated the PR description to reference it. This PR now also adds regression tests that reproduce the broken packing described there: identity round trips for {int8; int64}, {int8; int32}, a struct passed between scalar arguments, and a padded nested struct followed by a sibling field. They fail on main and pass with this change. PTAL.

kumagi:
The linux/arm64 CI failure was a pre-existing bug that the new NestedSmallTail regression test exposed: in placeRegistersArm64, when a field's alignment pushed the bit cursor past a register boundary, the pending eightbyte was flushed but the flush was marked as final, so the field accumulated afterwards (the trailing int8) was never placed into a register and was lost. This commit leaves that intermediate flush non-final so the final flush still emits the trailing value. TestRegisterFunc_structArgs now passes on both linux/amd64 and linux/arm64.

struct_amd64.go
```diff
@@ -226,57 +228,110 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
...
+				if want > shift {
+					shift = want
+				}
 			}
 			switch f.Kind() {
 			case reflect.Struct:
-				place(f)
+				savedEight := curEight
+				place(f, fieldOff)
+				curEight = savedEight
```
hajimehoshi: [P2] Preserve the pending eightbyte index after recursion

Restoring `curEight` leaves it inconsistent with the shared accumulator when a nested struct spans eightbytes. For `struct { A struct { X, Y, Z int32 }; B int32 }`, after processing `A`, the accumulator contains `Z` in eightbyte 1, but this assignment restores the index to 0. Processing `B` then prematurely flushes `Z`; `B` is accumulated while `flushed` remains true and is never emitted. The array branch's corresponding restoration has the same problem with `struct { A [3]int32; B int32 }`.

Confirmed with real C functions summing the four fields initialized to 1, 2, 3, and 4: both layouts return 6 on this PR head, versus the expected 10 on base commit `763bb996ee5bc7661be7c7d59fb8f5bf53bc10c5` (macOS amd64 under Rosetta, `CGO_ENABLED=0`). Please keep `curEight` associated with the pending accumulator after recursion and add regression coverage for both layouts.

hajimehoshi (commented):
Found one amd64 argument-packing regression, reproduced with real C calls against both the PR head and its base.

Reviewed by Codex (OpenAI), on behalf of @hajimehoshi.

struct_amd64.go
kumagi: Good catch — fixed. Recursion shares the pending accumulator (val/shift/class) across levels, so restoring curEight to the outer field's eightbyte left it inconsistent with the accumulator whenever a nested struct spans eightbytes: the pending tail (Z, or the trailing array element) was flushed prematurely, and because that flush set flushed, the sibling field accumulated afterwards was never emitted. I removed the save/restore around both the struct and the array recursion, so curEight now stays associated with the pending accumulator (it is the eightbyte of the last placed field when shift != 0, and irrelevant when shift == 0 since a crossing flush requires shift != 0). Added regression coverage for both layouts as C functions summing the four fields initialized to 1, 2, 3, 4 (SumNestedIntsPlusOne / SumArrayIntsPlusOne), run through both RegisterLibFunc and the callback implementation: they return 6 on the previous head and 10 with this fix. Verified on linux/amd64 and linux/arm64 (qemu).

kumagi:
Addressed the review: tryPlaceRegister no longer restores the pending eightbyte index around struct/array recursion. The accumulator is shared across recursion levels, so keeping the outer field's index after recursion made it inconsistent whenever a nested struct spanned eightbytes — the tail was flushed prematurely and the flushed flag then dropped the following sibling. Both flagged layouts (struct { A struct { X, Y, Z int32 }; B int32 } and struct { A [3]int32; B int32 }) now keep curEight associated with the pending accumulator, and I added regression coverage for them using C sum functions over the four fields initialized to 1, 2, 3, 4. The tests return 6 on the previous head and 10 with this fix; verified on linux/amd64 and linux/arm64, including the callback variants. PTAL.

struct_amd64.go
```diff
@@ -226,57 +228,113 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
...
+			// The System V ABI classifies eightbytes from the in-memory
+			// image: a field starting in a later eightbyte than the pending
+			// accumulator ends it. Flush first so a wide field (e.g. the
+			// int64 in {int8; int64}) never overwrites accumulated smaller
+			// fields. A field wider than 4 bytes always starts a fresh
+			// eightbyte in practice (its alignment > remaining space), so
+			// after flushing we can place it directly without shifting.
+			needFresh := shift != 0 && fieldOff/8 != curEight
+			if needFresh {
+				flushIfNeeded()
```
hajimehoshi, line 248: [P2] Reset `flushed` after an intermediate boundary flush

For `struct { A struct { X int32; Y int8 }; B int8 }`, the inner struct has trailing padding, so `B` starts at offset 8 while the first eightbyte is still pending. This call emits that eightbyte and sets `flushed = true`. The `Int8` case then accumulates `B` without clearing the flag, so the final `flushIfNeeded()` skips it.

Confirmed with a real C function summing the fields initialized to 1, 2, and 3: the latest head returns 3 instead of 6 on macOS amd64 under Rosetta (`CGO_ENABLED=0`). Setting `flushed = false` immediately after this intermediate flush makes the reproducer and `TestRegisterFunc_structArgs` pass. Please reset the flag before accumulating the next field and add coverage for this trailing-padding layout.

This layout also fails on the PR base, so this is a remaining gap in the padding fix, not a new regression.

hajimehoshi (commented):
Confirmed that the previous nested-struct/array regression is fixed and the added regression tests pass. One related padding case remains, described inline.

Reviewed by Codex (OpenAI), on behalf of @hajimehoshi.

struct_amd64.go
kumagi: Fixed, thanks. The boundary flush emitted by the crossing check set flushed = true for the register slot it ended, and since the flag was only reset at the top of each field iteration, a small field accumulated into the fresh accumulator right after that flush was suppressed by the end-of-iteration and final flushIfNeeded. The crossing flush is now immediately followed by flushed = false so the new accumulator can still be emitted (the wide-field direct-placement paths keep setting the flag themselves, so they still suppress the final flush of their now-empty accumulator). Added SumNestedPadTail covering struct { A struct { X int32; Y int8 }; B int8 } with the fields initialized to 1, 2, 3: it returns 3 on the previous head and 6 with this fix, on both the RegisterLibFunc and callback paths. As you noted the layout also fails on the base, and on linux/arm64 too: there placeRegistersArm64 positioned fields purely by per-field alignment, so the recursion dropped the sibling at the last inner field's end (bit 40) instead of its in-memory offset (bit 64 of the next slot). That function now tracks the in-memory offset corresponding to the pending register and, after recursing into a struct/array field, realigns the cursor to the composite's end (emitting the pending register when the padding carries past the slot). Verified on linux/amd64 and linux/arm64 (qemu).

kumagi:
Addressed the latest review: the boundary-crossing flush in tryPlaceRegister left the accumulator marked as flushed, so a small field accumulated afterwards (the trailing int8 in struct { A struct { X int32; Y int8 }; B int8 }) was skipped by the final flushIfNeeded and never emitted. The flag is now reset after that intermediate flush. The new SumNestedPadTail regression test also exposed the same layout failing on linux/arm64: placeRegistersArm64 ignored a nested composite's trailing padding and packed the following sibling back-to-back after the last inner field instead of at its in-memory offset. It now tracks the memory offset backing the pending register and realigns the cursor to the composite's end after recursion. Verified on linux/amd64 and linux/arm64 (qemu), direct and callback paths; the test returns 3 before the fix and 6 after. PTAL.

struct_amd64.go
```diff
@@ -226,57 +228,119 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
...
+				}
 			}
 			switch f.Kind() {
 			case reflect.Struct:
-				place(f)
+				// Recursion shares the accumulator, so it must also share
+				// curEight: when a nested struct spans eightbytes its tail
+				// is pending in the last one it touched, and the next
+				// sibling field belongs to that same eightbyte. Restoring
+				// the outer index would flush the tail prematurely.
```
hajimehoshi, line 273: Shorten comments

struct_amd64.go
```diff
@@ -226,57 +228,119 @@ func tryPlaceRegister(v reflect.Value, addFloat func(uintptr), addInt func(uintp
...
 				f = v.Index(i)
+				fieldOff = base + uintptr(i)*f.Type().Size()
+			}
+			// The System V ABI classifies eightbytes from the in-memory
+			// image: a field starting in a later eightbyte than the pending
+			// accumulator ends it. Flush first so a wide field (e.g. the
+			// int64 in {int8; int64}) never overwrites accumulated smaller
+			// fields. A field wider than 4 bytes always starts a fresh
+			// eightbyte in practice (its alignment > remaining space), so
+			// after flushing we can place it directly without shifting.
```
hajimehoshi, line 245: Shorten comments

struct_test.go
```diff
@@ -944,6 +944,154 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+				if ret := fn(expected); ret != expected {
+					t.Fatalf("IdentityNestedSmallTail returned %+v wanted %+v", ret, expected)
+				}
+			})
+			t.Run("NestedIntsPlusOne", func(t *testing.T) {
+				// A nested struct whose tail remains pending in the second
+				// eightbyte followed by a sibling field in that same
+				// eightbyte. Recursion must leave the pending eightbyte
+				// index consistent with the accumulator so the sibling is
+				// merged into it instead of dropping it.
```
hajimehoshi, line 1034: Shorten comments

struct_test.go
```diff
@@ -944,6 +944,154 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+				if ret := sum(NestedIntsPlusOne{A: inner{X: 1, Y: 2, Z: 3}, B: 4}); ret != 10 {
+					t.Fatalf("SumNestedIntsPlusOne returned %d wanted 10", ret)
+				}
+			})
+			t.Run("NestedPadTail", func(t *testing.T) {
+				// The nested struct has trailing padding, so the sibling
+				// field starts in the next eightbyte while the first one is
+				// still pending. The boundary flush must not mark the
+				// accumulator as final: the sibling accumulated afterwards
+				// still needs the final flush.
```
hajimehoshi, line 1059: Shorten comments

TotallyGamerJet:
If we do this we need to update the documentation to say it is handled

> Purego can handle the most common structs that have fields of builtin types like int8, uint16, float32, etc. However, it does not support aligning fields properly. It is therefore the responsibility of the caller to ensure that all padding is added to the Go struct to match the C one. See `BoolStructFn` in struct_test.go for an example.

struct_amd64.go
kumagi: Shortened. The recursion note is now one sentence over two lines: the nested tail stays pending in the accumulator, so curEight must not be restored here. The neighbouring crossing and padding comments in the same function, and the cursor comments in placeRegistersArm64, were trimmed the same way.

struct_amd64.go
kumagi: Shortened. The crossing comment is three lines saying why the flush has to happen before a wide field, the note after the flush is one line, and the alignTo note is two.

struct_test.go
kumagi: Shortened. Each new subtest now carries a single sentence of at most two lines; the layout itself is readable from the Go struct and the matching C prototype, so the comments only name the failure mode they guard against.

struct_test.go
kumagi: Shortened. NestedIntsPlusOne and NestedPadTail are one sentence each about the pending eightbyte, and ArrayIntsPlusOne just refers to NestedIntsPlusOne.

kumagi:
Addressed the latest feedback.

The comments are shortened: the eightbyte-crossing note, the boundary-flush note and the recursion note in `struct_amd64.go`, the cursor notes in `struct_arm64.go`, and the comments on the new subtests in `struct_test.go` are now one or two lines each and state only the invariant they protect.

I also followed up on @TotallyGamerJet's point about the documentation. The `RegisterFunc` Structs section no longer says purego cannot align fields: it now explains that each field is placed at the offset it has in the Go struct's memory image, so the padding the Go compiler inserts is preserved and explicit padding fields are not required. The Go struct still has to mirror the C declaration (same fields, same order, embedding `structs.HostLayout`), and purego still does not verify that. The stale "purego won't do padding for you" note in `struct_test.go` is gone, and a new `BoolFloatNoPadding` subtest calls the existing C `BoolFloat` with a Go struct that relies on the compiler's padding, through both the `RegisterLibFunc` and the callback implementations; the same layout also round-tripped on linux/loong64 and linux/ppc64le under qemu.

The branch remains mergeable with main. Locally: `gofmt -s -l` is empty, `go vet ./...` passes, all the CI cross-compile targets build, and `TestRegisterFunc_structArgs` / `TestRegisterFunc_structReturns` pass on linux/amd64 (Cgo and non-Cgo) and linux/arm64 (qemu).

PTAL.

PR #516, opened 2026-09-05
purego: document that NewCallback consumes one table entry per call

### LINKED ISSUE #521: Passing the same Go callback to C in a loop exhausts the 2000-entry callback table
### PureGo Version

main @ 763bb99

### Operating System

- [x] Windows
- [x] macOS
- [x] Linux
- [x] FreeBSD
- [x] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64

Every call to `NewCallback` — including the implicit one that `RegisterFunc` performs for each `func`-typed argument on **every invocation** (`func.go: addValue`, `case reflect.Func: addInt(NewCallback(...))`) — consumes one entry of the fixed `maxCB = 2000` callback table (`syscall_unix.go: compileCallback`), which is never released. Passing the same Go callback to C in a loop therefore kills the process after 2000 calls.

Repro (pure Go + libc, no C toolchain needed):

```go
libc, _ := purego.Dlopen("libc.so.6", purego.RTLD_NOW|purego.RTLD_GLOBAL)
compare := func(a, b *int) int { return *a - *b }
var qsort func(data []int, nitms uintptr, size uintptr, compar func(a, b *int) int)
purego.RegisterLibFunc(&qsort, libc, "qsort")
for i := 0; i < 2005; i++ {
    data := []int{2, 1}
    qsort(data, uintptr(len(data)), unsafe.Sizeof(int(0)), compare)
}
// panics: purego: the maximum number of callbacks has been reached
```

Bisected further: the loop completes iterations 0..1999 and dies on the 2000th call, i.e. exactly one table slot per call even though the function value never changes.

The doc comment on `NewCallback` mentions the 2000-entry limit but not that the *same* function value consumes a new entry on every call, so users have no way to know that a per-frame `qsort`/enumerate callback (60fps ≈ 33 seconds to death) is fatal, nor what the workaround is.

Either the same function value reuses its trampoline, or — if that is not possible — the documentation tells users to resolve the callback once with `NewCallback` and reuse the returned pointer instead of passing a Go func value on every call.

### What happens instead?

A loop that passes a Go callback to C panics with `purego: the maximum number of callbacks has been reached` after 2000 iterations, taking down the whole process. There is no API to release entries.

Minimized reproducible case: the `qsort` loop above (verified failing on `main @ 763bb99`, Linux/amd64, Go 1.26.7 — last completed iteration is 1999).

Note on automatic deduplication: `reflect.Value.Pointer()` on func values returns the shared code pointer, so two different closures created by `reflect.MakeFunc` compare equal and must NOT be used as a dedup key (verified: two distinct callbacks collapsed to the same trampoline address when keyed this way). That is why PR #516 documents the pitfall and the manual-reuse workaround instead of deduplicating.

Related PR: #516.

kumagi:
Closes #521

bug

## What this PR does | solves
Passing a Go func value to C in a loop (e.g. via RegisterFunc's func arguments, which call NewCallback on every invocation) exhausts the fixed 2000-entry callback table and panics. Function values have no comparable identity available to reflect, so automatic deduplication is not possible; document the pitfall and the workaround of resolving the callback once and reusing the pointer.

### DIFF
--- a/syscall_unix.go
+++ b/syscall_unix.go
@@ -28,7 +28,7 @@ func syscall_syscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 //
 // Every call to NewCallback creates a new callback even for the same function value, so passing a Go callback to C
 // inside a loop (e.g. a qsort comparator) keeps consuming callbacks and eventually panics once they are exhausted.
-// The same happens when a func value is passed to a C function, as RegisterFunc creates a new callback for each
+// The same happens when a func value is passed to a C function, as [RegisterFunc] creates a new callback for each
 // call. Create the callback once with NewCallback and reuse the returned pointer instead.
 func NewCallback(fn any) uintptr {
 	ty := reflect.TypeOf(fn)
--- a/syscall_unix.go
+++ b/syscall_unix.go
@@ -25,6 +25,12 @@ func syscall_syscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 // of uintptr. Only a limited number of callbacks may be created in a single Go process, and any memory allocated
 // for these callbacks is never released. At least 2000 callbacks can always be created. Although this function
 // provides similar functionality to windows.NewCallback it is distinct.
+//
+// Each call to NewCallback consumes one entry of the fixed callback table,
+// even for the same function value: passing a Go callback to C inside a
+// loop (e.g. a qsort comparator) exhausts the table after 2000 calls and
+// panics. Resolve the callback once with NewCallback and reuse the returned
+// pointer instead of letting RegisterFunc create a new one per call.
 func NewCallback(fn any) uintptr {
 	ty := reflect.TypeOf(fn)
 	for i := range ty.NumIn() {
--- a/syscall_unix.go
+++ b/syscall_unix.go
@@ -23,14 +23,13 @@ func syscall_syscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 // This is useful when interoperating with C code requiring callbacks. The argument is expected to be a
 // function with zero or one uintptr-sized result. The function must not have arguments with size larger than the size
 // of uintptr. Only a limited number of callbacks may be created in a single Go process, and any memory allocated
-// for these callbacks is never released. At least 2000 callbacks can always be created. Although this function
+// for these callbacks is never released. At least 1024 callbacks can always be created. Although this function
 // provides similar functionality to windows.NewCallback it is distinct.
 //
-// Each call to NewCallback consumes one entry of the fixed callback table,
-// even for the same function value: passing a Go callback to C inside a
-// loop (e.g. a qsort comparator) exhausts the table after 2000 calls and
-// panics. Resolve the callback once with NewCallback and reuse the returned
-// pointer instead of letting RegisterFunc create a new one per call.
+// Every call to NewCallback creates a new callback even for the same function value, so passing a Go callback to C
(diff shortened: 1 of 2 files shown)


syscall_unix.go
```diff
@@ -25,6 +25,12 @@ func syscall_syscallN(fn uintptr, args ...uintptr) (r1, r2, err uintptr) {
 // of uintptr. Only a limited number of callbacks may be created in a single Go process, and any memory allocated
 // for these callbacks is never released. At least 2000 callbacks can always be created. Although this function
 // provides similar functionality to windows.NewCallback it is distinct.
+//
+// Each call to NewCallback consumes one entry of the fixed callback table,
+// even for the same function value: passing a Go callback to C inside a
+// loop (e.g. a qsort comparator) exhausts the table after 2000 calls and
```
hajimehoshi: 2000 is an internal detail. We should not expose this.

hajimehoshi: Or, we can show a minimum guarantee like https://pkg.go.dev/syscall?GOOS=windows#NewCallback

In this case, 1024 is the safest minimum value, and actually NewCallback invokes Go's NewCallback on Windows, so 1024 seems the correct value.

@TotallyGamerJet Do you agree with this?

kumagi: Good point — addressed. The concrete table size is no longer mentioned: the doc now says "At least 1024 callbacks can always be created", which matches syscall.NewCallback on Windows (purego.NewCallback delegates to the standard library there, and the Unix table is larger than that, so 1024 holds on every supported platform). The pitfall paragraph now just says that every call creates a new callback and that the callbacks are eventually exhausted and panic, without naming the internal limit.

kumagi:
Thanks for the review! I dropped the internal 2000 figure from the docs. NewCallback now states the same kind of minimum guarantee as syscall.NewCallback on Windows — at least 1024 callbacks can always be created — and the added note explains that every call creates a new callback (including the func arguments handled by RegisterFunc) without exposing the table size. Happy to adjust if a different wording or number is preferred.

hajimehoshi (approved):
LGTM

@TotallyGamerJet PTAL

syscall_unix.go
kumagi: Agreed — the docs now state a minimum guarantee instead of the internal table size: "At least 1024 callbacks can always be created", matching syscall.NewCallback on Windows. purego.NewCallback delegates to syscall.NewCallback/NewCallbackCDecl there, and the Unix callback table is larger than that, so 1024 holds on every supported platform without exposing maxCB.

I also moved the pitfall note into the Windows NewCallback documentation, since delegating to the standard library means every call creates a new callback there too.

kumagi:
Thanks again for the review! The docs no longer mention the internal table size: NewCallback now states the same minimum guarantee as syscall.NewCallback on Windows (at least 1024 callbacks), which holds everywhere because purego.NewCallback delegates to the standard library on Windows and the Unix table is larger than that.

Since the behavior is not Unix specific — purego.NewCallback on Windows goes straight to syscall.NewCallback, which also creates a brand new callback on every call — I extended the same note to the Windows NewCallback documentation so both platforms describe the loop pitfall and the workaround of creating the callback once and reusing the returned pointer. The notes now link to RegisterFunc, which is where the per-call cost is easiest to hit accidentally.

Everything is documentation only; go build, go vet and go test still pass, and the docs were checked for both Unix and Windows.

TotallyGamerJet (approved):
LGTM

PR #517, opened 2026-09-05
purego: pass non-HFA arm64 structs in integer registers

### LINKED ISSUE #522: arm64 sends mixed non-HFA structs on x0/v0 instead of x0/x1
### PureGo Version

main @ 763bb99

### Operating System

- [ ] Windows
- [x] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [x] Android
- [x] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64 (finding is in arm64-only code; verified by code inspection + cross-build, see below)

On arm64, `placeRegistersArm64` (`struct_arm64.go`, the non-darwin path) routes struct members straight to FP or integer registers **by kind**:

```go
case reflect.Int64, reflect.Int:
    addInt(uintptr(f.Int()))       // GPR
...
case reflect.Float64:
    addFloat(...)                  // FPR
```

So a non-HFA mixed struct such as `{int64; float64}` is sent as `x0` + `v0`. But AAPCS64 says a non-HFA composite of 16 bytes or less is packed into 1–2 **general-purpose** registers (`x0`, `x1`) — and purego's own receive paths already do exactly that:

- `getCallbackStruct` (`struct_arm64.go:595-611`): "Non-HFA composite ≤ 16 bytes: packed into 1–2 integer registers", reading only `f[numOfFloatRegisters()+*intsN ...]`.
- `getStruct` (`struct_arm64.go:38-55`): non-`isAllSameFloat` returns fall back to `syscall.a1, syscall.a2`.

Three code paths, one of which disagrees — the sender is the odd one out (3-against-1 within the repo, plus the ABI spec).

Repro (arm64 only — this box is amd64/Linux, so no execution trace; the case below is what to run on an M-series Mac, Linux/arm64, or Android/iOS device):

```c
struct Mixed { int64_t i; double d; };
struct Mixed ident_mixed(struct Mixed s) { return s; }
```

```go
type Mixed struct {
    _ structs.HostLayout
    I int64
    D float64
}
var ident func(Mixed) Mixed
purego.RegisterLibFunc(&ident, lib, "ident_mixed")
got := ident(Mixed{I: 1, D: 2.5})
// want {I:1 D:2.5}; with the bug D arrives in v0 while C reads x1
```

Additionally a Go→C→Go round trip through `NewCallback` is self-inconsistent even without C: the sender writes `x0`/`v0` while the receiver reads `x0`/`x1`.

Mixed non-HFA structs travel packed in `x0`/`x1`, matching AAPCS64 and purego's own receive paths, so the round trip above returns `{I:1 D:2.5}`.

### What happens instead?

The float member goes out on `v0` while the callee (real C ABI, or purego's own callback receiver) reads `x1`: the double arrives as garbage/zero and integer register accounting shifts for all following arguments.

Minimized reproducible case: the `ident_mixed` C helper plus the Go snippet above. Requires an arm64 host (or device) to execute; on this amd64 box I verified instead that (1) `GOOS=linux GOARCH=arm64` and `GOOS=darwin GOARCH=arm64` builds pass, (2) the existing test suite passes, and (3) the three in-repo code paths cited above disagree exactly as described.

Related PR: #517.

kumagi:
Closes #522

bug

## What this PR does | solves
placeRegistersArm64 routed float/64-bit members straight to FP or integer registers by kind, so mixed structs such as {int64; float64} went out on x0/v0 while AAPCS64 (and our own getCallbackStruct) expect them packed into x0/x1. Copy the in-memory image eightbyte by eightbyte for non-HFA/HVA aggregates of 16 bytes or less.

### DIFF
--- a/struct_arm64.go
+++ b/struct_arm64.go
@@ -89,6 +89,13 @@ func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFl
 			*numFloats = numOfFloatRegisters()
 		} else if hva && *numInts+numABIFields(v.Type()) > numOfIntegerRegisters() {
 			*numInts = numOfIntegerRegisters()
+		} else if !hfa && !hva && !isDarwin && *numInts+int(roundUpTo8(size)/8) > numOfIntegerRegisters() {
+			// A non-HFA/HVA composite packed into integer registers is
+			// passed entirely on the stack when the remaining integer
+			// registers cannot hold all of its eightbytes (AAPCS64;
+			// mirrors getCallbackStruct). Darwin may instead split it
+			// between registers and the stack.
+			*numInts = numOfIntegerRegisters()
 		}
 
 		placeRegisters(v, addFloat, addInt)
--- a/struct_arm64.go
+++ b/struct_arm64.go
@@ -107,6 +107,35 @@ func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr
 }
 
 func placeRegistersArm64(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
+	// Non-HFA composites of 16 bytes or less are passed packed into
+	// 1-2 general-purpose registers (AAPCS64 C.7; mirrors
+	// getCallbackStruct). Copy the in-memory image eightbyte by
+	// eightbyte instead of routing members by kind, so mixed structs
+	// such as {int64; float64} land in x0/x1 rather than x0/v0.
+	if !isHFA(v.Type()) && !isHVA(v.Type()) && v.Type().Size() <= 16 {
+		if !v.CanAddr() {
+			tmp := reflect.New(v.Type()).Elem()
+			tmp.Set(v)
+			v = tmp
+		}
+		ptr := v.Addr().UnsafePointer()
+		size := v.Type().Size()
+		for off := uintptr(0); off < size; off += 8 {
+			var chunk uintptr
+			if remaining := size - off; remaining >= 8 {
+				chunk = *(*uintptr)(unsafe.Add(ptr, off))
+			} else {
+				// Final partial chunk: read byte-by-byte to avoid
+				// reading beyond the value's allocation.
+				for i := range remaining {
+					b := *(*byte)(unsafe.Add(ptr, off+i))
+					chunk |= uintptr(b) << (i * 8)
+				}
+			}
+			addInt(chunk)
+		}
+		return
+	}
 	var val uint64
 	var shift byte
 	var flushed bool
--- a/struct_arm64.go
+++ b/struct_arm64.go
@@ -91,10 +91,12 @@ func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFl
 			*numInts = numOfIntegerRegisters()
 		} else if !hfa && !hva && !isDarwin && *numInts+int(roundUpTo8(size)/8) > numOfIntegerRegisters() {
 			// A non-HFA/HVA composite packed into integer registers is
-			// passed entirely on the stack when the remaining integer
-			// registers cannot hold all of its eightbytes (AAPCS64;
-			// mirrors getCallbackStruct). Darwin may instead split it
-			// between registers and the stack.
+			// passed entirely on the stack, consuming the remaining
+			// integer registers, when they cannot hold all of its
+			// eightbytes (AAPCS64; mirrors getCallbackStruct). Darwin
+			// enforces the same all-or-nothing rule earlier, in the
+			// stack-argument bundling path (see shouldBundleStackArgs),
+			// so there is nothing to do here.
 			*numInts = numOfIntegerRegisters()
 		}
 
--- a/struct_arm64.go
+++ b/struct_arm64.go
@@ -127,22 +127,7 @@ func placeRegistersArm64(v reflect.Value, addFloat func(uintptr), addInt func(ui
 			tmp.Set(v)
 			v = tmp
 		}
-		ptr := v.Addr().UnsafePointer()
-		size := v.Type().Size()
-		for off := uintptr(0); off < size; off += 8 {
-			var chunk uintptr
-			if remaining := size - off; remaining >= 8 {
-				chunk = *(*uintptr)(unsafe.Add(ptr, off))
-			} else {
-				// Final partial chunk: read byte-by-byte to avoid
-				// reading beyond the value's allocation.
-				for i := range remaining {
-					b := *(*byte)(unsafe.Add(ptr, off+i))
-					chunk |= uintptr(b) << (i * 8)
-				}
-			}
-			addInt(chunk)
-		}
+		copyStruct8ByteChunks(v.Addr().UnsafePointer(), v.Type().Size(), addInt)
 		return
 	}
 	var val uint64
@@ -348,12 +333,14 @@ func isHVA(t reflect.Type) bool {
 	}
 }
 
-// copyStruct8ByteChunks copies struct memory in 8-byte chunks to the provided callback.
-// This is used for Darwin ARM64's byte-level packing of non-HFA/HVA structs.
+// copyStruct8ByteChunks copies struct memory in 8-byte chunks to the provided
+// callback. The final partial chunk is read byte-by-byte so that nothing beyond
+// the value's allocation is touched, and is zero-extended.
+//
+// Both Darwin ARM64's byte-level packing and the AAPCS64 rule for non-HFA/HVA
+// composites of 16 bytes or less pass such an argument as consecutive chunks of
+// its in-memory image, so the two paths share this helper.
 func copyStruct8ByteChunks(ptr unsafe.Pointer, size uintptr, addChunk func(uintptr)) {
-	if !isDarwin {
-		panic("purego: should only be called on darwin")
-	}
 	for offset := uintptr(0); offset < size; offset += 8 {
 		var chunk uintptr
 		remaining := size - offset
--- a/struct_test.go
+++ b/struct_test.go
@@ -853,6 +853,35 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 					t.Fatalf("IdentityFloatAndInt returned %+v wanted %+v", ret, expected)
 				}
 			}
+			if runtime.GOARCH == "arm64" {
+				// Mixed int64 + float64: non-HFA, so it must be packed
+				// into x0/x1 instead of being split across x0/v0 (AAPCS64).
+				type Int64AndDouble struct {
+					_ structs.HostLayout
+					A int64
+					B float64
+				}
+				var fn func(Int64AndDouble) Int64AndDouble
+				register(&fn, lib, "IdentityInt64AndDouble", func(s Int64AndDouble) Int64AndDouble {
+					return s
+				})
+				expected := Int64AndDouble{A: -1234, B: 5.25}
+				if ret := fn(expected); ret != expected {
+					t.Fatalf("IdentityInt64AndDouble returned %+v wanted %+v", ret, expected)
+				}
+				if runtime.GOOS != "darwin" {
+					// Only one integer register is left, so the whole
+					// struct must go on the stack rather than being split
+					// between a register and the stack.
+					var fn func(int64, int64, int64, int64, int64, int64, int64, Int64AndDouble) Int64AndDouble
+					register(&fn, lib, "IdentityInt64AndDoubleAfterRegisters", func(a, b, c, d, e, f, g int64, s Int64AndDouble) Int64AndDouble {
+						return s
+					})
+					if ret := fn(1, 2, 3, 4, 5, 6, 7, expected); ret != expected {
+						t.Fatalf("IdentityInt64AndDoubleAfterRegisters returned %+v wanted %+v", ret, expected)
+					}
+				}
+			}
 			{
 				// Struct > 16 bytes: hidden pointer on amd64, pointer in int register on arm64.
 				type ThreeInt64 struct {
--- a/struct_test.go
+++ b/struct_test.go
@@ -869,10 +869,12 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 				if ret := fn(expected); ret != expected {
 					t.Fatalf("IdentityInt64AndDouble returned %+v wanted %+v", ret, expected)
 				}
-				if runtime.GOOS != "darwin" {
-					// Only one integer register is left, so the whole
-					// struct must go on the stack rather than being split
-					// between a register and the stack.
+				{
+					// Only one integer register is left, so the whole struct
+					// must go on the stack rather than being split between
+					// a register and the stack. Darwin's ABI agrees: clang
+					// spills both eightbytes to the stack and consumes the
+					// remaining integer register.
 					var fn func(int64, int64, int64, int64, int64, int64, int64, Int64AndDouble) Int64AndDouble
 					register(&fn, lib, "IdentityInt64AndDoubleAfterRegisters", func(a, b, c, d, e, f, g int64, s Int64AndDouble) Int64AndDouble {
 						return s
--- a/testdata/structtest/struct_test.c
+++ b/testdata/structtest/struct_test.c
@@ -417,6 +417,19 @@ struct FloatAndInt IdentityFloatAndInt(struct FloatAndInt s) {
     return s;
 }
 
+struct Int64AndDouble {
+    int64_t a;
+    double b;
+};
+
+struct Int64AndDouble IdentityInt64AndDouble(struct Int64AndDouble s) {
+    return s;
+}
+
+struct Int64AndDouble IdentityInt64AndDoubleAfterRegisters(int64_t a, int64_t b, int64_t c, int64_t d, int64_t e, int64_t f, int64_t g, struct Int64AndDouble s) {
+    return s;
+}
+
 struct ThreeInt64 {
     int64_t a, b, c;
 };


hajimehoshi:
Fix the test fail

struct_arm64.go
```diff
@@ -107,6 +107,35 @@ func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr
 }
 
 func placeRegistersArm64(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
+	// Non-HFA composites of 16 bytes or less are passed packed into
+	// 1-2 general-purpose registers (AAPCS64 C.7; mirrors
+	// getCallbackStruct). Copy the in-memory image eightbyte by
+	// eightbyte instead of routing members by kind, so mixed structs
+	// such as {int64; float64} land in x0/x1 rather than x0/v0.
+	if !isHFA(v.Type()) && !isHVA(v.Type()) && v.Type().Size() <= 16 {
```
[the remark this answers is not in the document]
kumagi: Fixed. `addStruct` now checks whether all `(size+7)/8` eightbyte slots of a non-HFA/HVA struct fit in the remaining integer registers, and if not, exhausts them (`*numInts = numOfIntegerRegisters()`) before calling `placeRegisters`, so every chunk spills to the stack contiguously — the same all-or-nothing behavior as `getCallbackStruct`. This matches AAPCS64 and also the Windows arm64 ABI; Darwin keeps its split-across-registers-and-stack convention, so the check is gated on `!isDarwin`. Covered by the new `IdentityInt64AndDoubleAfterRegisters` test (7 leading int64 args, only x7 left, whole struct must land on the stack), which fails without this change and passes with it under qemu-aarch64.

struct_arm64.go
```diff
@@ -107,6 +107,35 @@ func placeRegisters(v reflect.Value, addFloat func(uintptr), addInt func(uintptr
 }
 
 func placeRegistersArm64(v reflect.Value, addFloat func(uintptr), addInt func(uintptr)) {
+	// Non-HFA composites of 16 bytes or less are passed packed into
+	// 1-2 general-purpose registers (AAPCS64 C.7; mirrors
+	// getCallbackStruct). Copy the in-memory image eightbyte by
+	// eightbyte instead of routing members by kind, so mixed structs
+	// such as {int64; float64} land in x0/x1 rather than x0/v0.
```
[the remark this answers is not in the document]
kumagi: Added. `TestRegisterFunc_structArgs` now has an arm64-only round-trip for `struct{ int64; double }` (`IdentityInt64AndDouble`), which runs both as a Go→C identity call and through a Go callback (`GoCallbackFunc`), so `placeRegistersArm64` and `getCallbackStruct` are both exercised. I also added the register-overflow variant `IdentityInt64AndDoubleAfterRegisters` (non-Darwin) to pin down the all-or-nothing stack spill. Verified that both tests fail on the pre-fix code — B comes back as 0 because the float was routed to v0 — and pass with the packing fix.

kumagi:
Addressed both review comments:

1. Small non-HFA/HVA structs that don't fit entirely in the remaining integer registers are now forced to the stack as a whole. `addStruct` exhausts the integer registers before placement in that case, mirroring the all-or-nothing rule in `getCallbackStruct` (AAPCS64). Darwin is unchanged since Apple's ABI does allow splitting a struct between registers and the stack.
2. Added arm64 round-trip tests for `struct{ int64; double }` (issue #522's case): a plain identity call plus a variant with seven leading `int64` arguments that exercises the register-overflow path. Both run as Go→C identity (`RegisterLibFunc`) and Go→C→Go callback (`GoCallbackFunc`).

Verified under qemu-aarch64 against real compiled C: the new tests fail on the pre-fix code (float dropped to v0 / struct split across x7+stack) and pass now. The full test suite passes on linux/amd64 and on linux/arm64 (CGO enabled and disabled). Tests should be green.

struct_test.go
```diff
@@ -853,6 +853,35 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+				}
+				var fn func(Int64AndDouble) Int64AndDouble
+				register(&fn, lib, "IdentityInt64AndDouble", func(s Int64AndDouble) Int64AndDouble {
+					return s
+				})
+				expected := Int64AndDouble{A: -1234, B: 5.25}
+				if ret := fn(expected); ret != expected {
+					t.Fatalf("IdentityInt64AndDouble returned %+v wanted %+v", ret, expected)
+				}
+				if runtime.GOOS != "darwin" {
```
hajimehoshi, line 872: What happens with the below function on Darwin?

kumagi: On Darwin this case behaves the same way as on Linux. I verified with `clang --target=arm64-apple-macosx11.0 -S` that for `f(int64_t a..g, struct { int64_t a; double b; } s)` the callee reads both eightbytes from the incoming stack (`ldp x0, x1, [sp]`), and in a variant with a trailing `int64_t` after the struct that argument also lands on the stack — so Apple's ABI consumes the remaining integer register and passes the whole struct on the stack, no register/stack split. (Mixed members stay in GP registers as well: `struct { float, float, int64_t }` arrives as `stp x0, x1` in the callee.)

Purego's Darwin path already produces exactly that placement: `shouldBundleStackArgs` intercepts a small non-HFA struct when `numInts + slotsNeeded > 8`, `collectStackArgs`/`structFitsInRegisters` see it doesn't fit, `bundleStackArgs` copies the whole struct to the stack, and the callback side (`getCallbackStruct` → `readStructFromStackArm64` → `callbackArgFromStack`) reads it back all-or-nothing. That bundling path is already covered on macOS CI by the `stack_8int_*struct*` cases in `TestABI_ArgumentPassing`.

Given the skip was unnecessary, I removed it: `IdentityInt64AndDoubleAfterRegisters` now runs on all arm64 including Darwin, so macOS CI verifies this directly. The `!isDarwin` gate in `addStruct` stays — Darwin already made the same decision earlier in the bundling layer — and its comment is corrected to say so instead of claiming Darwin splits the struct.

kumagi:
Followed up on the Darwin question about `IdentityInt64AndDoubleAfterRegisters`.

I checked what Apple's arm64 ABI actually does for `f(int64_t x7, struct { int64_t a; double b; } s)` by inspecting `clang --target=arm64-apple-macos -S` output: the callee reads both eightbytes from the incoming stack area (`ldp x0, x1, [sp]`), and in a variant with a trailing `int64_t` after the struct that argument also lands on the stack, i.e. Darwin consumes the last integer register and passes the whole struct on the stack — the same all-or-nothing choice as AAPCS64 for this type. Purego's Darwin path already produces exactly that layout: a small non-HFA struct that no longer fits is diverted to `bundleStackArgs` before any register placement (the path exercised by the existing `stack_8int_*struct*` cases in `TestABI_ArgumentPassing` on macOS), and `getCallbackStruct`/`callbackArgFromStack` read it back the same way.

So the `runtime.GOOS != "darwin"` skip was unnecessary; I removed it, and the test now runs on Darwin too. I also corrected the `addStruct` comment, which wrongly suggested Darwin splits such a struct across registers and stack. Re-verified: the full suite passes on linux/amd64 and on linux/arm64 under qemu, and the new tests still fail on the pre-fix code.

struct_arm64.go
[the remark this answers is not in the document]
kumagi: Fixed: `addStruct` now checks `*numInts + ceil(size/8) > numOfIntegerRegisters()` for non-HFA/HVA composites and sets `*numInts = numOfIntegerRegisters()` in that case, so all chunks spill contiguously instead of splitting the struct between the last register and the stack. That mirrors `getCallbackStruct` and AAPCS64, and `IdentityInt64AndDoubleAfterRegisters` (seven leading `int64` arguments, only x7 left) fails without the check. Darwin is excluded there because `shouldBundleStackArgs`/`bundleStackArgs` already divert the whole struct to the stack earlier, and clang's `arm64-apple-macos` codegen confirms the same all-on-stack placement.

struct_arm64.go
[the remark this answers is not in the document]
kumagi: Added: `TestRegisterFunc_structArgs` now has an arm64 round-trip for `struct{ int64; double }` (`IdentityInt64AndDouble`), exercised both as a Go to C identity call and through a Go callback (`GoCallbackFunc`), so `placeRegistersArm64` and `getCallbackStruct` are both covered. The register-overflow variant `IdentityInt64AndDoubleAfterRegisters` pins the whole-struct stack spill. Both fail on the pre-fix code — `B` comes back as 0 because the float was routed to v0 — and pass with the packing fix.

struct_test.go
kumagi: The same placement as Linux. For `f(int64_t a..g, struct { int64_t a; double b; } s)`, `clang --target=arm64-apple-macos -S` produces a callee that reads both eightbytes from the incoming argument stack (`ldr x8, [sp]` and `ldr d0, [sp, #8]`), and with a trailing `int64_t` after the struct that argument goes to the next stack slot too, so Apple's ABI consumes the remaining integer register and passes the whole struct on the stack rather than splitting it.

Purego's Darwin path already yields exactly that layout: a small non-HFA struct that no longer fits is intercepted by `shouldBundleStackArgs`, `structFitsInRegisters` reports it as not fitting, and `bundleStackArgs` copies the whole struct to the stack, with `getCallbackStruct` reading it back the same way. That path is already exercised on macOS by the `stack_8int_*struct*` cases in `TestABI_ArgumentPassing`.

So the `runtime.GOOS != "darwin"` skip was unnecessary and is removed — `IdentityInt64AndDoubleAfterRegisters` now runs on Darwin too. The `!isDarwin` gate in `addStruct` stays (the decision is simply made earlier there), and its comment no longer claims that Darwin splits such a struct.

kumagi:
All review threads are addressed and the branch applies cleanly on top of main.

- Overflow: `addStruct` now exhausts the integer registers when a small non-HFA/HVA composite needs more eightbytes than remain, so the whole struct goes on the stack — the same all-or-nothing rule `getCallbackStruct` already implements.
- Coverage: `TestRegisterFunc_structArgs` round-trips `struct{ int64; double }` (`IdentityInt64AndDouble`) both as a Go to C identity call and through a Go callback, and `IdentityInt64AndDoubleAfterRegisters` (seven leading `int64` arguments, only x7 left) pins the whole-struct stack spill.
- Darwin: `clang --target=arm64-apple-macos -S` shows Apple's ABI making the same choice for this shape — both eightbytes land in the incoming argument stack and the last integer register is consumed — so the Darwin skip is removed and macOS CI verifies it as well. Purego reaches that layout through `shouldBundleStackArgs`/`bundleStackArgs` before any register placement.

One follow-up cleanup: the eightbyte loop added to `placeRegistersArm64` duplicated `copyStruct8ByteChunks`, so the AAPCS64 packing path and the Darwin byte-packing path now share that helper. Its Darwin-only assertion is gone because every caller already pins the platform it belongs to.

Verification: `gofmt -s`, `go vet`, the documented cross-compilations, and the full test suite on linux/amd64 (Cgo enabled and disabled, and with `-gcflags=all=-N -l`). On linux/arm64 the suite runs under qemu-aarch64 against a real cross-compiled C library; the new tests still fail without the packing fix (`B` comes back as 0) and pass with it.

struct_arm64.go
```diff
@@ -89,6 +89,15 @@ func addStruct(v reflect.Value, numInts, numFloats, numStack *int, addInt, addFl
...
 		} else if hva && *numInts+numABIFields(v.Type()) > numOfIntegerRegisters() {
 			*numInts = numOfIntegerRegisters()
+		} else if !hfa && !hva && !isDarwin && *numInts+int(roundUpTo8(size)/8) > numOfIntegerRegisters() {
+			// A non-HFA/HVA composite packed into integer registers is
+			// passed entirely on the stack, consuming the remaining
+			// integer registers, when they cannot hold all of its
+			// eightbytes (AAPCS64; mirrors getCallbackStruct). Darwin
+			// enforces the same all-or-nothing rule earlier, in the
+			// stack-argument bundling path (see shouldBundleStackArgs),
+			// so there is nothing to do here.
```
hajimehoshi, line 99: Shorten comments

struct_test.go
```diff
@@ -853,6 +853,37 @@ func TestRegisterFunc_structArgs(t *testing.T) {
...
+				expected := Int64AndDouble{A: -1234, B: 5.25}
+				if ret := fn(expected); ret != expected {
+					t.Fatalf("IdentityInt64AndDouble returned %+v wanted %+v", ret, expected)
+				}
+				{
+					// Only one integer register is left, so the whole struct
+					// must go on the stack rather than being split between
+					// a register and the stack. Darwin's ABI agrees: clang
+					// spills both eightbytes to the stack and consumes the
+					// remaining integer register.
```
hajimehoshi, line 877: Shorten comments

struct_arm64.go
```diff
@@ -310,12 +333,14 @@ func isHVA(t reflect.Type) bool {
...
 
-// copyStruct8ByteChunks copies struct memory in 8-byte chunks to the provided callback.
-// This is used for Darwin ARM64's byte-level packing of non-HFA/HVA structs.
+// copyStruct8ByteChunks copies struct memory in 8-byte chunks to the provided
+// callback. The final partial chunk is read byte-by-byte so that nothing beyond
+// the value's allocation is touched, and is zero-extended.
+//
+// Both Darwin ARM64's byte-level packing and the AAPCS64 rule for non-HFA/HVA
+// composites of 16 bytes or less pass such an argument as consecutive chunks of
+// its in-memory image, so the two paths share this helper.
```
hajimehoshi, line 342: > Both Darwin ARM64's byte-level...

This seems an internal detail. Move this into the implementation

struct_test.go
```diff
@@ -853,6 +853,37 @@ func TestRegisterFunc_structArgs(t *testing.T) {
 					t.Fatalf("IdentityFloatAndInt returned %+v wanted %+v", ret, expected)
 				}
 			}
+			if runtime.GOARCH == "arm64" {
```
TotallyGamerJet: Why can't these tests run on other platforms?

struct_arm64.go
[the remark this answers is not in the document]
kumagi: This stays fixed: `addStruct` still exhausts the integer registers when a small non-HFA/HVA composite needs more eightbytes than remain, so the whole struct spills to the stack. The explanatory comment was shortened per the latest review; the behavior is unchanged.

struct_arm64.go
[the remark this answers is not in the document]
kumagi: Coverage is in place: `TestRegisterFunc_structArgs` round-trips `struct{ int64; double }` both as a Go to C identity call and through a Go callback. The identity case has also been moved out of the arm64 gate so every supported platform runs it; only the register-overflow variant remains arm64-only.

struct_test.go
kumagi: Unchanged from the earlier answer: Darwin produces the same all-on-stack placement for this shape (verified against clang codegen and already covered by the Darwin stack-argument bundling path), so the overflow test runs on Darwin too. The only follow-up here was shortening the test comment.

struct_arm64.go
kumagi: Shortened: the comment is now three lines stating the all-or-nothing spill and pointing at `shouldBundleStackArgs` for the Darwin case.

struct_test.go
kumagi: Shortened: the overflow case is now described in two lines.

struct_arm64.go
kumagi: Moved: the doc comment now only describes what the helper does, and the note about the Darwin and AAPCS64 paths sharing it is an implementation comment inside the function body.

struct_test.go
kumagi: They mostly can, and the identity case now does: a `struct{ int64; float64 }` round-trip is meaningful on every platform this test file covers, so I removed the arm64 gate from `IdentityInt64AndDouble` and verified it passes on linux/amd64 and under qemu on linux/arm64, ppc64le and loong64. Each ABI places the two eightbytes in the registers their classes select, and purego already does that everywhere.

`IdentityInt64AndDoubleAfterRegisters` stays arm64-only: it pins the all-or-nothing spill that comes from counting the struct's eightbytes against the remaining integer registers, which is the arm64 convention (Darwin reaches the same layout earlier, via the stack-argument bundling path). The analogous amd64 shape - seven leading `int64` arguments exhausting the six integer registers - currently runs into a separate, pre-existing limitation in the amd64 struct sender (the aggregate gets split across a stack slot and an SSE register instead of going wholly to memory) rather than the bug this PR fixes, so exercising it here would mix in unrelated amd64 work.

kumagi:
Followed up on the latest review round:

- Shortened the register-overflow comment in `addStruct` and the packing comment in `placeRegistersArm64`, and trimmed the test comments accordingly.
- `copyStruct8ByteChunks` no longer advertises the Darwin/AAPCS64 sharing detail in its doc comment; that note now lives inside the function, next to the loop it explains.
- The `IdentityInt64AndDouble` round-trip is no longer arm64-only: the plain case is meaningful on every architecture this test file covers (verified on linux/amd64 and under qemu on linux/arm64, ppc64le and loong64), so it now runs unconditionally. Only the `IdentityInt64AndDoubleAfterRegisters` spill case stays arm64-gated, since it asserts the eightbyte-count-based integer register exhaustion that is specific to AAPCS64/Darwin.

Verification: `gofmt -s`, `go vet`, the documented cross-compilations, and the full test suite on linux/amd64 (CGO enabled and disabled, and with `-gcflags=all=-N -l`). On linux/arm64 the struct tests run under qemu-aarch64 against a real cross-compiled C library, and I re-confirmed that `IdentityInt64AndDouble` still fails (`B` comes back as 0) when the integer-register packing path is removed.

PR #518, opened 2026-09-05
objc: create new blocks as stack blocks so Copy moves them

### LINKED ISSUE #523: NewBlock uses __NSMallocBlock__ isa so Copy never moves the block off Go memory
### PureGo Version

main @ 763bb99

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [x] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64 (finding is in darwin-only code; verified by code inspection + cross-build, see below)

`NewBlock` (`objc/objc_block_darwin.go:239-246`) builds its template with an `__NSMallocBlock__` isa and then "copies" the Go-side layout to the heap:

```go
layout := theBlocksCache.getLayout(reflect.TypeOf(fn)) // Go heap value
block := Block(unsafe.Pointer(&layout)).Copy()          // _Block_copy
```

But for a block whose isa is already `_NSConcreteMallocBlock`, `_Block_copy` is specified to be a no-op retain of the same pointer — no relocation happens. The resulting `Block` therefore keeps pointing at the Go allocation (`&layout`, a function-local that escapes to the Go heap). Since `Block` is a `uintptr`, the Go collector cannot see the reference: after a GC cycle the memory can be reused and `Invoke`/`Copy`/`Release` dereference a dangling pointer (use-after-free / random crash).

This contradicts the code's own stated design goal (`objc_block_darwin.go:16-19`: an Objective-C memory-managed block "that won't try to free() a Go pointer") and the comment on `NewBlock` (`:242`: "we'll copy it to a newly-created Objective-C object") — the copy to an Objective-C object never happens.

Repro (darwin only — this box is Linux, so no execution trace; steps for a Mac):

```go
b := objc.NewBlock(func(_ objc.Block, a int32, b int32) int32 { return a + b })
// force GC pressure, then invoke through the ObjC runtime:
runtime.GC(); debug.FreeOSMemory()
r, _ := objc.InvokeBlock[int32](b, int32(8), int32(2))
// want r == 10; with the bug the block may point at recycled Go memory
```

Correct behavior per the Clang Block ABI (`__block_literal_1` + `Block_copy`): a fresh block starts as `__NSStackBlock__`, and only `Copy` relocates it to the heap. Starting life as `__NSMallocBlock__` skips that step entirely.

`NewBlock` returns a genuine heap block owned by the Objective-C runtime, safe to keep across GC cycles and to pass to ObjC APIs.

### What happens instead?

The `Block` aliases Go heap memory invisible to the collector. It works by luck until a GC cycle recycles the allocation, then crashes nondeterministically — the worst kind of bug to diagnose in a shipped app.

Minimized reproducible case: the `NewBlock` + GC-pressure + `InvokeBlock` snippet above (needs macOS/iOS to execute). On this Linux box I verified instead that `GOOS=darwin GOARCH=arm64 go build` and `go vet ./objc/` pass with the fix, and that the existing `TestBlockCopyAndBlockRelease` expectations are consistent with stack-isa semantics (copies compare distinct from the original after a real heap relocation).

Related PR: #518.

kumagi:
Closes #523

bug

## What this PR does | solves
NewBlock built its template with an __NSMallocBlock__ isa, for which _Block_copy is a no-op retain of the Go-side allocation. The Block then points at Go memory invisible to the collector. Start from __NSStackBlock__ so Copy relocates the block to the heap as the ABI requires.

### DIFF
--- a/objc/objc_block_darwin.go
+++ b/objc/objc_block_darwin.go
@@ -19,7 +19,10 @@ const (
 	// when the reference count drops to zero, so the associated function is also unreferenced.
 
 	// blockBaseClass is the name of the class that block objects will be initialized with.
-	blockBaseClass = "__NSMallocBlock__"
+	// New blocks start on the stack so that Block_copy moves them to the
+	// heap; starting from __NSMallocBlock__ would make Copy a no-op retain
+	// of the Go-side allocation.
+	blockBaseClass = "__NSStackBlock__"
 	// blockFlags is the set of flags that block objects will be initialized with.
 	blockFlags = blockHasCopyDispose | blockHasSignature
 


hajimehoshi:
Let's continue the discussion at #523

kumagi:
Sure, keeping the discussion in #523.

Outcome there: this is not a use-after-free. `Block_copy` chooses its path from the flags word (`BLOCK_NEEDS_FREE` / `BLOCK_IS_GLOBAL`) rather than from the isa, and it gives the relocated block its `__NSMallocBlock__` isa itself, so `NewBlock` was already returning heap memory (verified on macOS arm64 via `malloc_size` / `malloc_zone_from_ptr`).

I left the change as a naming/clarity one only: the constant stays `__NSStackBlock__` because that describes a not-yet-copied template, and I dropped the comment claim that `Copy` would be a no-op retain. I'll retitle the PR accordingly, and I'm happy to close it if you would rather not touch this at all.

hajimehoshi (approved):
LGTM, assuming this is just a code clarification, not a bug fix

hajimehoshi:
@TotallyGamerJet PTAL

TotallyGamerJet (approved):
LGTM

PR #519, opened 2026-09-05
purego: return 64-bit callback results on 386

### LINKED ISSUE #524: 386 callback trampoline drops the upper half of int64/uint64 returns
### PureGo Version

main @ 763bb99

### Operating System

- [x] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

### Go Version (`go version`)

go1.26.7 linux/amd64 (finding is in 386-only assembly; verified via cross-build + disassembly, see below)

The 386 callback trampoline (`sys_unix_386.s: callbackasm1`) has two coupled defects around `callbackArgs.result`:

1. It allocates only **one** result word (`MOVL $0, 32(SP)`) and returns only `EAX` (`MOVL 32(SP), AX`), while Go's `callbackArgs` has `result [4]uintptr` and `setInt64Result`/`setUint64Result` (`syscall_notstackargs.go:43-55`) write the upper half to `result[1]` on 32-bit platforms. i386 cdecl returns 64-bit values in `EDX:EAX`, so every `NewCallback(func(...) int64/uint64)` silently loses its upper 32 bits (`0x1_00000000` becomes `0`).
2. The args copy starts at `36(SP)` — overlapping `result[1..3]` of the Go struct — so the copy and the result alias each other.

Contrast `sys_unix_arm.s:70-71`, which correctly returns both `R0` and `R1` for 64-bit callback results.

Verification on this box (no 32-bit execution environment, so proven at the binary level instead):

```
// GOOS=linux GOARCH=386 go build, then objdump the trampoline:
CALL crosscall2(SB)
ADDL $0x10, SP
MOVL 0x20(SP), AX        // result[0] -> EAX only; no DX load anywhere before RET
...
RET
```

`NewCallback` accepts `int64`/`uint64` returns (`syscall_unix.go: compileCallback` output switch), the Go side dutifully writes both halves — and the assembly drops the top half on the floor.

Repro (386 Linux or Windows/386 — needs a 32-bit host to execute):

```go
cb := purego.NewCallback(func(a, b int32) int64 {
    return int64(a)<<32 | int64(uint32(b)) // e.g. 0x1_00000002
})
// call cb from C as int64_t f(int32_t, int32_t); observe the return
// want 0x100000002; with the bug the upper half reads 0x0
```

64-bit callback results round-trip intact (`EDX:EAX`), and the args area no longer overlaps the result area.

### What happens instead?

Deterministic truncation: the upper 32 bits of every `int64`/`uint64` callback return read as zero (whatever happened to be in `EDX`), and deep argument lists can clobber pending results.

Minimized reproducible case: the `NewCallback` snippet above driven from C (or any 64-bit-returning callback on a 386 host). On this amd64 box I verified instead that (1) `GOOS=linux GOARCH=386` and `GOOS=windows GOARCH=386` builds pass, (2) the rebuilt trampoline disassembles to `MOVL 0x20(SP), AX` + `MOVL 0x24(SP), DX` before the register restores, and (3) the full test suite passes on amd64.

Related PR: #519.

kumagi:
Closes #524

bug

## What this PR does | solves
The 386 trampoline allocated only one result word and returned EAX alone, so int64/uint64 callbacks lost their upper half (cdecl wants EDX:EAX) and the args copy overlapped result[1..]. Widen the callbackArgs area to the full Go struct, shift the args copy by 12 bytes, and return both halves.

### DIFF
--- a/callback_test.go
+++ b/callback_test.go
@@ -154,6 +154,51 @@ func TestNewCallbackFloat32AndFloat64(t *testing.T) {
 	}
 }
 
+func TestNewCallbackInt64Result(t *testing.T) {
+	// A 64-bit result needs two words: EDX:EAX on 386 and R1:R0 on arm.
+	// The trampoline must forward both halves of the value.
+	const mask = int64(0x0123456789abcdef)
+	imp := purego.NewCallback(func(v int64) int64 { return v ^ mask })
+	var fn func(v int64) int64
+	purego.RegisterFunc(&fn, imp)
+	for _, v := range []int64{
+		0,
+		-1,
+		1 << 32,
+		-0x7edcba9876543210,
+	} {
+		if got, want := fn(v), v^mask; got != want {
+			t.Errorf("callback(%#x) = %#x, want %#x", uint64(v), uint64(got), uint64(want))
+		}
+	}
+}
+
+func TestNewCallbackUint64Result(t *testing.T) {
+	// The upper half of a uint64 result must not be dropped by the trampoline.
+	imp := purego.NewCallback(func(lo, hi uint32) uint64 { return uint64(hi)<<32 | uint64(lo) })
+	var fn func(lo, hi uint32) uint64
+	purego.RegisterFunc(&fn, imp)
+	const want = uint64(0x0123456789abcdef)
+	if got := fn(0x89abcdef, 0x01234567); got != want {
+		t.Errorf("callback() = %#x, want %#x", got, want)
+	}
+}
+
+func TestNewCallbackInt64ResultWithStackArgs(t *testing.T) {
+	// On 386 all the arguments are passed on the stack, so this also checks
+	// that the copied arguments do not overlap the result of the callback.
+	const wantSum = 1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 + 10 + 11 + 12 + 13 + 14 + 15 + 16 + 17 + 18 + 19 + 20
+	imp := purego.NewCallback(func(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20 int) int64 {
+		return int64(a1+a2+a3+a4+a5+a6+a7+a8+a9+a10+a11+a12+a13+a14+a15+a16+a17+a18+a19+a20)<<32 | 0x0badf00d
+	})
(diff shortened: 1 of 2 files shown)


hajimehoshi:
Can we have tests?

kumagi:
Tests added in callback_test.go:

- `TestNewCallbackInt64Result` and `TestNewCallbackUint64Result` return values whose upper word is significant, so the EDX half of the result is checked.
- `TestNewCallbackInt64ResultWithStackArgs` passes 20 stack arguments together with a 64-bit result, which covers the shifted argument copy in the trampoline.

I verified them with `GOARCH=386 CGO_ENABLED=0 go test`: all three fail against the previous trampoline (the upper word came back as garbage) and pass with this change, and they also pass on amd64.

While doing that I noticed that the last copied argument was still written to 288(SP) instead of 300(SP) after the argument area was shifted, so argument slot 63 was never written and slot 60 was overwritten. That is fixed as well, along with the frame layout comments.

hajimehoshi:
Fix the test failure

kumagi:
The failure was on Linux ppc64le: `TestNewCallbackInt64ResultWithStackArgs` passed 20 int arguments, but ppc64le only has `maxArgs` (15) slots of which 8 are integer registers, so `RegisterFunc` panicked with `purego: too many stack arguments` and the panic took down the whole test binary.

The test now passes 12 arguments. That is within every supported architecture's limit and still puts all the arguments on the stack on 386, and still spills onto the stack on arm (8), s390x (7) and the other 64-bit architectures (4), so the shifted argument copy remains covered.

Verified locally:

- `GOARCH=386 CGO_ENABLED=0`: all three tests fail against the previous trampoline (the upper word came back as garbage) and pass with this change.
- `GOARCH=ppc64le`, `GOARCH=riscv64` and `GOARCH=s390x` (s390x needs Go 1.27 for `CGO_ENABLED=0`) under qemu: the tests run and pass, and the ppc64le panic is gone.
- `GOARCH=amd64` with `CGO_ENABLED=0`/`CGO_ENABLED=1` and with `-gcflags=all=-N -l`: pass, and `gofmt -s -l` plus `go vet` are clean.

hajimehoshi (approved):
LGTM (but TBH I don't fully understand the code...)

@TotallyGamerJet PTAL

TotallyGamerJet (approved):
LGTM

Issue #528, opened 2026-09-07
Question: how should property-based tests be accepted without making CI flaky?


kumagi:
### Operating System

- [X] Windows
- [X] macOS
- [X] Linux
- [X] FreeBSD
- [X] NetBSD
- [X] Android
- [X] iOS

### What feature would you like to be added?

I would like to ask how the maintainers feel about property-based tests (PBTs) for purego, and if they are welcome, in what form they should be contributed so that CI does not become flaky.

What I mean by PBT here: instead of one hand-written case with one fixed signature, a test generates many random inputs/types from a small model, runs them against an oracle, and asserts an invariant that must hold for every generated case. The oracle in my prototypes is the real C ABI itself: the test emits C declarations equivalent to the generated Go types, compiles them into a shared library, and compares purego's behavior byte-for-byte against what the C compiler did. Generation is seeded so failures reproduce deterministically, and every failure dumps the seed, the Go type, the emitted C source, and a hex diff.

I prototyped three such probes locally:

1. `internal/strings` `CString` / `GoString` round trip. Generates 5000 random strings over a mixed alphabet (ASCII, symbols, multibyte runes, sometimes NUL-terminated to cover both the zero-copy and copy paths in `CString`) plus deterministic edge cases (empty, leading/interior NULs, 64KB strings, 4096 Japanese characters). Invariant: `GoString(CString(s))` equals the prefix of `s` up to the first NUL, i.e. the C-string contract survives both paths untouched.

2. `RegisterFunc` struct ABI round trip. Generates random struct types (scalars, small arrays, nested structs) and call signatures mixing structs with scalars, emits equivalent C declarations, compiles them, and round-trips values through the real C ABI in both argument and return directions, including register-exhaustion and stack layouts. Invariant: every argument byte the Go side passes must arrive intact, and every return byte must come back intact.

3. `NewCallback` argument/return round trip. Generates random callback signatures (scalar/array/struct parameters with scalar/struct/void returns), emits a C forwarder that calls the Go callback through the real C ABI, and checks both directions byte-for-byte. Invariant: arguments passed from Go must reach the Go callback uncorrupted, and the callback's return value must come back uncorrupted. The current prototype skips Windows (stdcall restrictions on callback argument types) and excludes float returns per `compileCallback`; I am not sure how far the Windows boundary should be covered, hence all OS boxes are checked above.

[...]

PR #529, opened 2026-09-07
purego: add a fuzz test for the CString/GoString round trip

kumagi:
#528

feature

## What this PR does | solves
Seeded edge cases run on every plain go test; longer exploration happens only under go test -fuzz.


kumagi:
GoString won't have fuzz-able bugs in the future.

PR #534, opened 2026-09-22
objc: add explicit-class super sends

### LINKED ISSUE #533: objc: SendSuper recursively reenters inherited overrides
### PureGo Version

Reproduced with v0.11.0. Source inspection of main at 656af01456b2cd7a761884de06ccd88b31c30032 shows the same runtime-class-based dispatch in both `ID.SendSuper` and `SendSuper[T]`; main was not executed.

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] NetBSD
- [ ] Android
- [ ] iOS

macOS 26.5.1 (25F80), arm64.

### Go Version (`go version`)

go version go1.27.0 darwin/arm64

Register three classes: `Base : NSObject`, `Override : Base`, and `Child : Override`. Define a method on Base and override it on Override, calling `self.SendSuper(cmd)` from that override. Child inherits the override without adding any methods.

Calling the method on an Override instance works. Calling it on a Child instance recursively reenters Override's method.

Save the following as `main.go` in a module requiring `github.com/ebitengine/purego v0.11.0`, then run `go run .` on macOS. It uses Foundation only and opens no window. The recursion is capped at five calls to avoid a stack overflow. The third case checks explicit defining-class dispatch through `objc_msgSendSuper2`.

```go
package main

import (
	"fmt"
	"structs"

	"github.com/ebitengine/purego/objc"
)

type objcSuper struct {
	_        structs.HostLayout
	receiver objc.ID
	class    objc.Class
}

func main() {
	lib, err := purego.Dlopen("/System/Library/Frameworks/Foundation.framework/Foundation", purego.RTLD_NOW|purego.RTLD_GLOBAL)
	if err != nil {
		panic(err)
	}
	var superSend func(*objcSuper, objc.SEL) objc.ID
	purego.RegisterLibFunc(&superSend, lib, "objc_msgSendSuper2")
	sel := objc.RegisterName("probe")
	var baseCalls, overrideCalls int
	var fixed bool
	base, err := objc.RegisterClass("Issue3740Base", objc.GetClass("NSObject"), nil, nil, []objc.MethodDef{
		{
			Cmd: sel,
			Fn:  func(self objc.ID, cmd objc.SEL) { baseCalls++ },
		},
	})
	if err != nil {
		panic(err)
	}
	var override objc.Class
	override, err = objc.RegisterClass("Issue3740Override", base, nil, nil, []objc.MethodDef{
		{
			Cmd: sel,
			Fn: func(self objc.ID, cmd objc.SEL) {
				overrideCalls++
				if overrideCalls >= 5 {
					return
				}
				if fixed {
					superSend(&objcSuper{
						receiver: self,
						class:    override,
					}, cmd)
				} else {
					self.SendSuper(cmd)
				}
			},
		},
	})
	if err != nil {
		panic(err)
	}
	child, err := objc.RegisterClass("Issue3740DynamicSubclass", override, nil, nil, nil)
	if err != nil {
		panic(err)
	}
	for _, tc := range []struct {
		name         string
		class        objc.Class
		fixed        bool
		wantBase     int
		wantOverride int
	}{
		{
			name:         "direct instance, existing SendSuper",
			class:        override,
			wantBase:     1,
			wantOverride: 1,
		},
		{
			name:         "subclass instance, existing SendSuper (guarded at 5)",
			class:        child,
			wantOverride: 5,
		},
		{
			name:         "subclass instance, lexical-class super dispatch",
			class:        child,
			fixed:        true,
			wantBase:     1,
			wantOverride: 1,
		},
	} {
		baseCalls, overrideCalls, fixed = 0, 0, tc.fixed
		obj := objc.ID(tc.class).Send(objc.RegisterName("new"))
		obj.Send(sel)
		obj.Send(objc.RegisterName("release"))
		fmt.Printf("%s: base=%d override=%d\n", tc.name, baseCalls, overrideCalls)
		if baseCalls != tc.wantBase || overrideCalls != tc.wantOverride {
			panic("unexpected dispatch counts")
		}
	}
}
```

A super call made from Override's method should reach Base's implementation once, including when the receiver is a Child instance. There should be an API that allows the caller to specify Override as the class defining the method, matching Objective-C super semantics.

### What happens instead?

Observed output:

```text
direct instance, existing SendSuper: base=1 override=1
subclass instance, existing SendSuper (guarded at 5): base=0 override=5
subclass instance, lexical-class super dispatch: base=1 override=1
```

For a Child instance, `ID.SendSuper` supplies `id.Class() == Child` to `objc_msgSendSuper2`. That function starts lookup in Child's superclass, Override, and finds the same override again. Without the guard, the method keeps recursively reentering itself.

This differs from #509: passing the current class to `objc_msgSendSuper2` is correct, and the direct-instance control above succeeds. The problem is using the receiver's runtime class instead of the class defining the executing method. Substituting `id.Class().SuperClass()` would break ordinary direct-instance calls by advancing an extra level.

An explicit defining-class parameter or a separate class-anchored super-send API would allow correct dispatch for inherited methods and dynamically subclassed receivers. Both the method and generic helpers currently use `id.Class()`; the executable reproduction above exercises the method helper and the raw explicit-class fix.

Discovered while investigating https://github.com/hajimehoshi/ebiten/issues/3740. The connection to that application's crash remains unconfirmed: neither the reporter's dynamic receiver class nor a reproduction of the resizing trigger has been established locally. The class hierarchy above independently reproduces the purego API limitation.

Filed by Codex (OpenAI), on behalf of @hajimehoshi.

hajimehoshi:

Closes #533

bug

## What this PR does | solves

An override that calls SendSuper can recursively reenter itself when the
receiver inherits the override, because dispatch uses the receiver's
runtime class instead of the class defining the method.

Add SendSuper1 and SendSuper2, each as an ID method and a generic helper.
SendSuper1 invokes objc_msgSendSuper and starts lookup in the supplied
superclass. SendSuper2 invokes objc_msgSendSuper2 and starts lookup in
the supplied defining class's superclass. Deprecate SendSuper while
preserving its behavior for compatibility; callers must migrate to an
explicit-class helper to handle inherited overrides correctly.

Add regression tests for direct and inherited calls, receiver and
argument forwarding, return values, and amd64 struct-return dispatch.
Update examples, including a three-level hierarchy showing the class
arguments required by the two helpers.

Validation:

- go test ./... on macOS arm64
- GOARCH=amd64 CGO_ENABLED=0 go test ./objc on macOS
- Executed the updated hierarchy example separately after its final edit

Authored by Codex (OpenAI), on behalf of @hajimehoshi.


hajimehoshi:
@TotallyGamerJet PTAL

hajimehoshi:
I've added an explanation about the background https://github.com/ebitengine/purego/issues/533#issuecomment-5779263666

Apparently `SendSuper2` might be redundant as `SendSuper1` can already represent the same thing. I'm happy to remove `SendSuper2` if you think it's better.

objc/objc_runtime_darwin_test.go
```diff
@@ -114,17 +125,23 @@ func ExampleID_SendSuper() {
...
 			{
-				Cmd: objc.RegisterName("doSomething"),
-				Fn: func(self objc.ID, _cmd objc.SEL) {
-					fmt.Println("In Child")
-					self.SendSuper(_cmd)
+				Cmd: sel,
+				Fn: func(self objc.ID, cmd objc.SEL) {
+					// Start lookup in Base, Child1's superclass.
+					fmt.Print("SendSuper1: ")
+					self.SendSuper1(base, cmd)
```
TotallyGamerJet: This is an example for Super2 so I don't think this should be here. Plus there is `ExampleSendSuper1`

hajimehoshi: Done.

examples/objc/main_darwin.go
```diff
@@ -18,7 +18,7 @@ var (
 )
 
 func BarInit(id objc.ID, cmd objc.SEL) objc.ID {
-	return id.SendSuper(cmd)
+	return id.SendSuper1(objc.GetClass("BarObject").SuperClass(), cmd)
```
TotallyGamerJet, line 21: Why use `SendSuper1` instead of `SendSuper2` to which would allow not calling `SuperClass`

objc/objc_runtime_darwin_test.go
```diff
@@ -95,17 +97,26 @@ func ExampleIMP() {
 	// Output: IMP: 105 567 9 2 3 -5 4 8 9
 }
 
-func ExampleID_SendSuper() {
-	super, err := objc.RegisterClass(
-		"SuperObject",
+// With Base <- Child1 <- Child2, a super call in Child1's method starts at Base,
+// even when the receiver is a Child2 instance. SendSuper1 takes Base,
+// the superclass of the class defining the method.
```
TotallyGamerJet, line 102: This comment is really confusing for someone coming across this example in pkg.

```suggestion
// Showcases multiple inheritance with Base <- Child1 <- Child2.
// SendSuper1 takes Base, the superclass of the class defining the method.
```

objc/objc_runtime_darwin_test.go
```diff
@@ -133,9 +145,99 @@ func ExampleID_SendSuper() {
...
+	// Output:
+	// Child1 instance
+	// SendSuper1: Base
+	// Child2 instance
+	// SendSuper1: Base
+}
+
+// With Base <- Child1 <- Child2, a super call in Child1's method starts at Base,
+// even when the receiver is a Child2 instance. SendSuper2 takes Child1,
+// the class defining the method.
```
TotallyGamerJet, line 173:
```suggestion
// Showcases multiple inheritance with Base <- Child1 <- Child2.
// SendSuper2 takes Child1, the class defining the method.
```

objc/objc_runtime_darwin_test.go
```diff
@@ -207,7 +309,7 @@ func ExampleSendSuper() {
 		panic(err)
 	}
 
-	res := objc.SendSuper[int](objc.ID(child).Send(objc.RegisterName("new")), objc.RegisterName("doSomething"))
+	res := objc.SendSuper1[int](objc.ID(child).Send(objc.RegisterName("new")), child.SuperClass(), objc.RegisterName("doSomething"))
```
TotallyGamerJet, line 312: If we are only going to have one example I'd prefer SendSuper2 since it's the more common one to use

objc/objc_runtime_darwin.go
```diff
@@ -171,36 +183,75 @@ func Send[T any](id ID, sel SEL, args ...any) T {
...
+//
+// Deprecated: Use [SendSuper1] with an explicit superclass or [SendSuper2]
+// with the class defining the executing method.
 func SendSuper[T any](id ID, sel SEL, args ...any) T {
+	return SendSuper2[T](id, id.Class(), sel, args...)
+}
+
+// SendSuper1 sends a message starting lookup in superClass and returns T.
+// For a super call, superClass must be the superclass of the class defining the executing method.
+func SendSuper1[T any](id ID, superClass Class, sel SEL, args ...any) T {
```
TotallyGamerJet, line 232: Do we really need SendSuper1? SendSuper2 seems to be what clang always produces. If we do we should at least make a comment in it's doc that if all you are doing is calling `ID.SuperClass` then use `SendSuper2` since it's one less C function call

objc/super_darwin_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"sync/atomic"
+	"testing"
+
+	"github.com/ebitengine/purego"
+	"github.com/ebitengine/purego/objc"
+)
+
+var superTestClassID atomic.Uint64
+
+func TestSendSuperDispatch(t *testing.T) {
```
TotallyGamerJet, line 21: What does this catch that the examples don't exercise? Can this be rewritten to be less difficult to read?

examples/objc/main_darwin.go
hajimehoshi: I preferred `SendSuper1` since `objc_msgSendSuper` is documented while `objc_msgSendSuper2` is not. If you prefer `SnedSuper2`, I'm fine to switch it.

objc/objc_runtime_darwin_test.go
hajimehoshi: > If we are only going to have one example I'd prefer SendSuper2 since it's the more common one to use

Do you mean `objc_msgSendSuper2` is more common than `objc_msgSendSuper`? Really? I thought it was opposite since `objc_msgSendSuper2` is not documented.

objc/objc_runtime_darwin.go
hajimehoshi: The same reason: `objc_msgSendSuper` is documented while `objc_msgSendSuper2` is not.

hajimehoshi: > SendSuper2 seems to be what clang always produces. 

This is an interesting point. IIUC, actually we don't need both and either one is enough, so we need to decide which 1 or 2 is preferable.

TotallyGamerJet: We already rely on undocumented symbols with NewBlock. Apple can't remove it since every compiled binary would need to recompiled. I prefer `SendSuper2` since it's faster since it means fewer C calls which is important for performance

hajimehoshi: Sure, then let me remove SendSuper1.

objc/objc_runtime_darwin_test.go
hajimehoshi: We decided to drop SendSuper1

examples/objc/main_darwin.go
hajimehoshi: We decided to drop SendSuper1

objc/super_darwin_test.go
hajimehoshi: Done

objc/objc_runtime_darwin_test.go
hajimehoshi: Done

objc/objc_runtime_darwin_test.go
hajimehoshi: Removed

objc/objc_runtime_darwin.go
```diff
@@ -171,36 +171,44 @@ func Send[T any](id ID, sel SEL, args ...any) T {
...
 	}
 	return objc_msgSendSuper2(super, sel, args...)
 }
 
-// SendSuper is a convenience method for sending message to object's super that can return any type.
-// This function takes a SEL instead of a string since RegisterName grabs the global Objective-C lock.
-// It is best to cache the result of RegisterName.
+// SendSuper sends a message starting lookup in the receiver's runtime superclass.
+//
+// Deprecated: Use [SendSuper2] with the class defining the executing method.
```
TotallyGamerJet, line 201: Is there a reason to deprecate instead of just removing it since purego is beta software so it's okay to have breaking changes. Plus we aren't backporting. If you think it's important to keep perhaps add `go:fix inline` so that people automatically get moved to the new API?

hajimehoshi: > Is there a reason to deprecate instead of just removing it since purego is beta software so it's okay to have breaking changes. 

Even though purego is beta, a transition period should be provided. Without a priod, if one dependency uses PureGo 0.11 and another uses 0.12, the library would fail to compile. Thus, the following schedule would be safer:

* 0.12: Deprecate `SendSuper` and `ID.SendSuper`
* 0.13: Remove `SendSuper` and `ID.SendSuper`

hajimehoshi: > Plus we aren't backporting.

I think we have not reached an agreement yet, but ok, now we agreed not backporting :)

hajimehoshi: > go:fix inline

I'm not sure we should do this. Giving an receiver's class as the first argument is often not correct.

TotallyGamerJet (approved):
LGTM

